PR #7, opened 2019-12-21
remove Cargo.lock

epwalsh:
I think it makes sense to *not* track changes to the `Cargo.lock` file for this project. More info here: https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html. Removing from git will definitely lead to fewer merge conflicts down the road

### DIFF
--- a/bindings/python/Cargo.lock
+++ b/bindings/python/Cargo.lock
@@ -448,15 +448,15 @@ dependencies = [
 
 [[package]]
 name = "tokenizers"
-version = "0.0.8"
+version = "0.0.9"
 dependencies = [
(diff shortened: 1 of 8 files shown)


bsima:
Note that in order to build with Nix you would need a Cargo.lock file: https://nixos.org/nixpkgs/manual/#compiling-rust-applications-with-cargo

n1t0:
Yes, the Cargo.lock file is necessary when building an application (binary), but since we don't provide any CLI at the moment, `tokenizers` is a library only.

We'll add one again when we'll be working on the CLI, in which case it would make sense to allow installing from other package managers maybe. But at the moment, I think the only use-case for `tokenizers` is in the `[dependencies]` of a binary app.

PR #27, opened 2020-01-02
TokenizerBuilder

epwalsh:
What do you think @n1t0?

### DIFF
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -120,6 +120,113 @@ impl std::cmp::PartialEq for AddedToken {
 }
 impl std::cmp::Eq for AddedToken {}
 
+struct Config {
+    // Tokenizer parts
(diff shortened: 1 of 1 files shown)


sai-prasanna:
There is a `derive-builder` macro, which makes creating builders trivial. Can we explore that?

https://github.com/colin-kiegel/rust-derive-builder

Usage in the wild

https://github.com/reneklacan/symspell/blob/master/src/symspell.rs#L22

epwalsh:
@sai-prasanna I like that! Could potentially remove a whole lot of boilerplate. I'll take a look

epwalsh:
@sai-prasanna would you be interested in making a PR to see if that works for replacing the other manually implemented builders (such as in the `src/models/*` modules)?

sai-prasanna:
@epwalsh Sure, will try to do it next weekend if I have time. Bit busy the following week.

n1t0:
This is out of date

PR #47, opened 2020-01-08
Added SentencePiece and YouTokenToMe model extractors.

mfuntowicz:


thomwolf:
Note for the future: we should rename this script to `extractor` (or move the YTTM part in another file).

PR #55, opened 2020-01-13
Fix a few errors in the README.md

benjamin-ny:
Most notably, the convention for representing Python code (using ">>>" for code, without for output) was used the wrong way round.


n1t0:
Thank you for this!

PR #64, opened 2020-01-14
Update Python training examples to reflect the higher-order API

mfuntowicz:
This PR aims to update the two high-level Python training examples:
- ByteLevelBPETokenizer
- BertWordPieceTokenizer


n1t0:
Nice, thanks @mfuntowicz

PR #68, opened 2020-01-14
Github action for node release

Pierrci:
Should be good now!


bindings/node/build.js
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env node
+
+/**
+ * Inspired from https://github.com/IronCoreLabs/recrypt-node-binding
```
julien-c: *from => by

.github/workflows/node-release.yml
```diff
@@ -0,0 +1,104 @@
...
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  AWS_DEFAULT_REGION: us-east-1
+
+on:
+  push:
+    tags:
+      - node-v*
+
+jobs:
+  build_all:
```
n1t0: Is this step doing something that the next one doesn't? Otherwise, since the `npm_publish` depends on `rust_publish` we can probably remove it. If one of the builds (in `rust_publish`) for any of the os/node were to fail, it would just fail the whole job anyway, right?
I think it would maybe make more sense to run the `build_all` job for each commit, to make sure we don't break anything in these bindings. What do you think?

bindings/node/build.js
```diff
@@ -0,0 +1,121 @@
...
+
+  // Cleanup the previous build, if it exists
+  shell.rm("-rf", distPath);
+
+  shell.exec("npm ci --ignore-scripts");
+  shell.mkdir(distPath);
+
+  shell.exec("npx tsc");
+  shell.cp("-r", ["lib/bindings"], distPath);
+  shell.mv([`${distPath}/bindings/native.prod.js`], [`${distPath}/bindings/native.js`]);
```
n1t0: :+1:

.github/workflows/node-release.yml
Pierrci: `rust_build` only builds rust, not typescript so that's why. I agree it would be better to run that at each commit, but maybe only the ts build then (and only when files inside `bindings/node` are updated or something like that)? There is already a rust build running at each commit.

Pierrci: And js tests could be integrated at this step too (when they arrive 😆)

n1t0: Sure but the `Rust` job doesn't check that the bindings still compile after something has been modified in the main library. At each commit, I think we would like to make sure everything is still compiling: the main library, but also all the bindings.

Pierrci: yes you're right, I pushed a new workflow `node.yml` with the `build_all` job and removed it from the release workflow

Pierrci: And we can maybe add a `paths-ignore` with `./bindings` in `rust.yml` since I don't see the need to rebuild the rust source if only code in bindings has been updated?

n1t0: Nice thanks! You are right, at the moment it's totally useless. I don't know what would be better between having a separate workflow for the main library and each bindings, or actually having one main `Build` workflow, that takes care of building everything on each OS, mutualizing the setup parts. What do you think? (This is out of the scope of this PR though).

n1t0:
Great job @Pierrci!

PR #71, opened 2020-01-15
Use the same vocabs in python's example.py

mfuntowicz:
Use the same vocab/merges files in order to get accurate numbers.

### DIFF
--- a/bindings/python/examples/example.py
+++ b/bindings/python/examples/example.py
@@ -18,7 +18,7 @@ parser.add_argument("--type", default="gpt2", type=str, help="The type of tokeni
 parser.add_argument("--file", default=None, type=str, help="The file to encode")
 parser.add_argument("--vocab", default=None, type=str, required=True, help="The vocab file")
 parser.add_argument("--merges", default=None, type=str, help="The merges.txt file")
-parser.add_argument("--debug", default=False, type=bool, help="Verbose output")
+parser.add_argument("--debug", action='store_true', type=bool, help="Verbose output")
 args = parser.parse_args()
 
 if args.type == "gpt2" and args.merges is None:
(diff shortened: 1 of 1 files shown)


mfuntowicz:
@n1t0 @thomwolf When using the same vocabs for Bert WordPiece (trained beforehand) I observe the same encoding from both Python and Rust side. But, when decoding, there are 70% (80k out of 128k lines) that have differences, mainly regarding comma:


```
Original:  As in the question of astronomy then, so in the question of
Rust:      [CLS] as in the question of astronomy then , so in the question of [SEP]
Python:    [CLS] as in the question of astronomy then, so in the question of [SEP]

Original:  history now, the whole difference of opinion is based on the
Rust:      [CLS] history now , the whole difference of opinion is based on the [SEP]
Python:    [CLS] history now, the whole difference of opinion is based on the [SEP]

Original:  recognition or nonrecognition of something absolute, serving as the
Rust:      [CLS] recognition or nonrecognition of something absolute , serving as the [SEP]
Python:    [CLS] recognition or nonrecognition of something absolute, serving as the [SEP]
```

n1t0:
Indeed, this is known behavior. If you want to have the same results, you can turn of the `clean_up_tokenization_spaces=False` On the `transformers` tokenizer. This is handled on transformers side.

Thank you for modifying this script! Is it really `from_pretrained` that should be used in this case?

mfuntowicz:
Understood for `clean_up_tokenization_spaces` 👍.

Regarding `from_pretrained` I would say *yes*, as we're not training the tokenizer as part of the script, so it should be trained beforehand and `from_pretrained` might allow us to provide `bert-base-cased` through the CLI `python example.py --type bert --vocab bert-base-cased`

I can modify the script so that I should work with our pretrained vocabs/merges on GPT2 too.

What do you think ?

n1t0:
My question is mainly because rusts tokenizer expects a file, and transformers one expect a name I think. If both work using this method, then it works for me too!

mfuntowicz:
Yap you're right, so in the current situation they will both be able to load locally stored models (rust and python), but not transformers' based one.

n1t0:
Ok then perfect!

PR #75, opened 2020-01-15
Added doc for setting tokenizers level of parallelism.

mfuntowicz:


n1t0:
Thanks! Maybe we should provide some other way to do this indeed. But it the meantime works great!

PR #76, opened 2020-01-15
node: add tests + linting

Pierrci:
Actual tests are in a020128 and eb0174b, last two commits are about adding linters to the project and running them for the first time.


bindings/node/lib/bindings/tokenizer.test.ts
```diff
@@ -0,0 +1,64 @@
...
+// const TokenizerMock = mocked(Tokenizer);
+
+let tokenizer: Tokenizer;
+beforeEach(() => {
+  // Clear all instances and calls to constructor and all methods:
+  // TokenizerMock.mockClear();
+
+  const model = BPE.empty();
+  tokenizer = new Tokenizer(model);
+  tokenizer.addTokens(["my", "name", "is", "john", "pair"]);
```
Pierrci: @n1t0 on my machine the tests in `tokenizer.test.ts` fail here with this error: `internal error in Neon module: attempt to subtract with overflow`. It seems to be due to this line in the rust source: https://github.com/huggingface/tokenizers/blob/4b5291b769b96d3ee9ae95f70062d1a0e52c3286/tokenizers/src/tokenizer/mod.rs#L590
It sounds rather logical since the vocab size is initially `0`, same for the added tokens, which gives a `-1` result incompatible with an unsigned int.

I was wondering how it's possible the tests pass in the CI, it turns out it's related to the rust toolchain used (nightly-2019-11-01). If I override on my laptop with this version the tests pass too, but not sure that's the way we want to solve that. Also related to https://github.com/huggingface/tokenizers/pull/93

n1t0: Indeed, I don't know how this happened haha but it is now fixed thanks to #93. I don't think you really have to use the nightly version of the toolchain for `node` though. The last stable should work fine. It is for `pyo3` if I remember correctly so only needed for python bindings.

PR #77, opened 2020-01-15
Improve truncation

n1t0:
Applying truncation on an `Encoding` will now produce an output `Encoding` and populate its `overflowing` property with as many other `Encoding` as needed to cover all the possible combinations.

@thomwolf @mfuntowicz This should cover everything as we discussed it.

### DIFF
--- a/bindings/node/native/src/encoding.rs
+++ b/bindings/node/native/src/encoding.rs
@@ -125,8 +125,8 @@ declare_types! {
             Ok(js_offsets.upcast())
         }
 
-        method getOverflowings(mut cx) {
-            // getOverflowings(): Encoding[]
+        method getOverflowing(mut cx) {
+            // getOverflowing(): Encoding[]
 
             let this = cx.this();
             let guard = cx.lock();
--- a/bindings/node/native/src/encoding.rs
+++ b/bindings/node/native/src/encoding.rs
@@ -125,27 +125,28 @@ declare_types! {
             Ok(js_offsets.upcast())
         }
 
-        method getOverflowing(mut cx) {
-            // getOverflowing(): Encoding | undefined;
+        method getOverflowings(mut cx) {
+            // getOverflowings(): Encoding[]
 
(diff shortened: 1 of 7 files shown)


bindings/node/native/src/encoding.rs
```diff
@@ -125,8 +125,8 @@ declare_types! {
             Ok(js_offsets.upcast())
         }
 
-        method getOverflowings(mut cx) {
-            // getOverflowings(): Encoding[]
+        method getOverflowing(mut cx) {
```
Pierrci: That's where the tests I added would have broken 😁

bindings/node/native/src/encoding.rs
```diff
@@ -125,27 +125,28 @@ declare_types! {
             Ok(js_offsets.upcast())
         }
 
-        method getOverflowing(mut cx) {
-            // getOverflowing(): Encoding | undefined;
+        method getOverflowings(mut cx) {
+            // getOverflowings(): Encoding[]
```
Pierrci: Actually it's going to break, perfect case :)

mfuntowicz:
I've some changes on the Python bindings as some method names are not correct, but we can merge and I'll push those changes on master/another branch.

Currently with truncating a string with max_length=X and pad_to_max_length=True the latest overflowing Encoding tokens are not padded. It's ok I'll do the necessary on the Python side.

That's very cool @n1t0 !

mfuntowicz:
I think there is an issue over the overflowing tokens, they seems to consider the special_chars [CLS], [SEP] in the length of the truncation, even if in the middle of the string

image

Is it something we can fix easily ?

n1t0:
@mfuntowicz What is the status on this? Did you try to integrate it? Can I merge?

mfuntowicz:
Yes 👌, sorry for the late reply on this.

n1t0:
Thanks for the fix @Pierrci!

PR #78, opened 2020-01-16
Cache cargo registry and build target directory in CI

epwalsh:
Big speed up to the Rust workflow!

Before:

image

After:

image


n1t0:
Thank you for this @epwalsh. We will probably do the same in other workflows! You can merge it if it’s ready!

PR #80, opened 2020-01-17
Remove trailing paren in Python binding README

iechevarria:
Removes a trailing paren so the code can run


n1t0:
Thank you @iechevarria!

Issue #89, opened 2020-01-19
Export Trainer::train API into Python bindings


vilunov:
HI!
Currently [Trainer::train](https://docs.rs/tokenizers/0.6.0/tokenizers/tokenizer/trait.Trainer.html#tymethod.train) method of the built trainer is missing from the Python object returned by [BpeTrainer.new](https://github.com/huggingface/tokenizers/blob/ef21c9a7b0ac374a3bfe7f982abe10157912797a/bindings/python/src/trainers.rs#L7-L10). This is important to have if you want to construct models using token counters from Python.

tRosenflanz:
Any news on this? Training from token counts would be very nice.
For instance, in my use case original data is processed in Spark so obtaining word counts is easy while dumping all of the strings into files would be very slow

n1t0:
I'm really not sure giving the ability to feed word counts is a good idea because the `PreTokenizer` is in charge of doing the pre-tokenization and it would introduce potential discrepancies between training and tokenization. Also, this trainer API is subject to change in the future, and may not be compatible with word counts anymore.

What about being able to stream raw strings instead?

tRosenflanz:
That would be useful too albeit more cumbersome. 

Would you mind expanding on the discrepancies you foresee? Currently, I use a custom loop using tfds subwordtextencoder methods that takes in word counts -> iterates over words, tokenizes them -> for each token increments token counter by the source word count -> builds encoder vocab using token counter.

PR #90, opened 2020-01-19
Expose train with word counts API

vilunov:
Issue #89


vilunov:
Lack of tests worries. I've tested this manually and I think it works now, but I've run into segmentation faults in the process. Perhaps it makes sense to create test suites for both bindings (e.g. with pytest for Python bindings)?

n1t0:
Thank you for this, but training with word counts is unlikely to stay, so this shouldn't be exposed as public API

PR #92, opened 2020-01-20
Fix BPETokenizer

n1t0:
This will, unfortunately, break any existing vocabulary since we change the PreTokenizer. But it is necessary to reproduce exactly the original BPE splitting only on whitespaces.


n1t0:
@Pierrci Can you make sure that I didn't break or forgot anything for the Node bindings? I'll let you do the next release too.

PR #99, opened 2020-01-23
Expose get_vocab_size in tokenizer python API.

### LINKED ISSUE #85: [Python bindings] Add method in BaseTokenizer to get vocab size.
### Feature request

Please add a method in `BaseTokenizer` to get vocab size. It is useful for specifying arguments to `nn.Embedding` layers in our external model pipeline. It can simply be:

```python
def get_vocab_size(self, with_added_tokens: Optional[bool] = True) -> int:
    self._tokenizer.get_vocab_size(with_added_tokens)
```

Should be as similar as the underlying Rust implementation. Here: https://github.com/huggingface/tokenizers/blob/master/bindings/python/tokenizers/implementations/base_tokenizer.py

kdexd:
Resolves #85

### DIFF
--- a/bindings/python/tokenizers/implementations/base_tokenizer.py
+++ b/bindings/python/tokenizers/implementations/base_tokenizer.py
@@ -13,6 +13,18 @@ class BaseTokenizer:
             self._tokenizer.get_vocab_size(),
             ', '.join(k + '=' + str(v) for k, v in self._parameters.items()))
 
+    def get_vocab_size(self, with_added_tokens: bool = True):
+        """ Return the size of vocabulary, with or without added tokens.
+
+        Args:
+            with_added_tokens: (`optional`) bool:
+                Whether to count in added special tokens or not
+
+        Returns:
+            Size of vocabulary
+        """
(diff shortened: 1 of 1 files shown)


bindings/python/tokenizers/implementations/base_tokenizer.py
```diff
@@ -13,6 +13,18 @@ def __repr__(self):
...
+        """ Return the size of vocabulary, with or without added tokens.
+
+        Args:
+            with_added_tokens: (`optional`) bool:
+                Whether to count in added special tokens or not
+
+        Returns:
+            Size of vocabulary
+        """
+        return self._tokenizer.get_vocab_size(with_added_tokens)
```
n1t0: Thank you for this PR @kdexd! If I remember correctly `with_added_tokens` should be a named argument and so I don't think it works if provided as a positional argument.

kdexd: Thanks for pointing out @n1t0 ! I hadn't added this option in my local patched version, glad it didn't slip by.

n1t0:
Thanks @kdexd!

PR #102, opened 2020-01-23
python: release action on tag push

Pierrci:
As discussed @n1t0: the python release action will be triggered when a `python-v*` tag is pushed in the repo instead of when a release is created.
It means that the release in itself will have to be created from the tag.


n1t0:
Great thanks! It should keep working with the usual release process since it usually triggers a push anyway.

PR #107, opened 2020-01-28
Improve flexibility in some Python binding 

mfuntowicz:
- [x] Rename BPETokenizer -> CharBPETokenizer
- [x] Possible to turn on/off lowercasing
- [x] Possible to turn on/off Unicode normalization
- [x] Factory method to create Unicode normalization

### DIFF
--- a/bindings/python/tokenizers/normalizers/__init__.pyi
+++ b/bindings/python/tokenizers/normalizers/__init__.pyi
@@ -7,7 +7,7 @@ class Normalizer:
     Normalizer will return an instance of this class when instantiated.
     """
 
-class BertNormalizer:
+class BertNormalizer(Normalizer):
     """ BertNormalizer
 
     Takes care of normalizing raw text before giving it to a Bert model.
@@ -40,7 +40,7 @@ class BertNormalizer:
         """
         pass
 
-class NFD:
+class NFD(Normalizer):
     """ NFD Unicode Normalizer """
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


bindings/python/tokenizers/normalizers/__init__.pyi
```diff
@@ -7,7 +7,7 @@ class Normalizer:
     Normalizer will return an instance of this class when instantiated.
     """
 
-class BertNormalizer:
+class BertNormalizer(Normalizer):
```
n1t0: Is there a specific reason you added these? Or is it just for accuracy?
I think we should add these for all the other parts too if we keep it here (PreTokenizer, Decoder, ...).

mfuntowicz: It's more syntactic sugar as `unicode_normalizer_from_str() -> Normalizer` returns a "Normalizer" and I wanted to reflect the relationship between the Normalizers we have.

We can also stay the way it is currently (i.e. no inheritance) and I change the prototype of the function to `unicode_normalizer_from_str() -> Union[NFC, NFD, NFKC, NFKD]`

What do you prefer @n1t0 ?

n1t0: I think the way you did it is the most accurate. Everywhere we expect `Normalizer` so let’s keep it this way. 
I would just like to keep everything consistent and add it for other classes too. You can do it here if you want, otherwise I’ll do it separately!

mfuntowicz:
As a side note, we need to release a version with the fixes included in this PR before the tokenizers branch gets merged in transformers, so that I can update the tokenizers dependency requirement.

PR #109, opened 2020-01-28
node: add tokenizer truncation / padding bindings

Pierrci:
I'm adding bindings I need for question answering in Node. For now:
- `tokenizer.setTruncation`
- `tokenizer.setPadding`

### DIFF
--- a/bindings/node/lib/bindings/encoding.d.ts
+++ b/bindings/node/lib/bindings/encoding.d.ts
@@ -1,0 +1,89 @@
+/**
+ * An Encoding as returned by the Tokenizer
+ */
+export interface Encoding {
+  /**
+   * Returns the attention mask
+   */
+  getAttentionMask(): number[];
+
+  /**
+   * Returns the tokenized ids
+   */
+  getIds(): number[];
+
+  /**
+   * Returns the offsets
+   */
+  getOffsets(): [number, number][];
+
+  /**
+   * Returns the overflowing encoding, after truncation
+   */
+  getOverflowing(): Encoding | undefined;
+
+  /**
+   * Returns the special tokens mask
+   */
+  getSpecialTokensMask(): number;
+
+  /**
+   * Returns the tokenized string
+   */
+  getTokens(): string[];
+
+  /**
+   * Returns the type ids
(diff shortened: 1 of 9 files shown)


n1t0:
Looks really good! How do you want to handle the case where you need to disable either the truncation or the padding after it's been set? In Python, we do `no_padding` or `no_truncation` for example.

Pierrci:
@n1t0 so I didn't really think about disabling padding/truncation since it was not my use case haha, but it was rather straightforward so I added `disableTruncation` and `disablePadding` methods. And that will be all for this PR.

Pierrci:
@n1t0 and I have noooooo f****** idea of why suddenly clippy is making the CI fail, it's raising errors on code in the main rs source files that were totally fine until now!! 🤬

Pierrci:
So the "problem" seems to come from the new Rust version 1.41.0 released today, but I think it actually behaves in 1.41 as it should have been from the start given this line in the rust CI:
https://github.com/huggingface/tokenizers/blob/6d532fedb1d3626328828304a5c39807733d2fa1/.github/workflows/rust.yml#L56

From my understanding, the `-D warnings` tells clippy to reject all warnings, which wasn't the case until now in the facts. So the possibilities are (for me):
- Fix all the clippy warnings
- Remove the `-D warnings` flag
- Pin to rust 1.40.0

wdyt @n1t0?

Pierrci:
> So the "problem" seems to come from the new Rust version 1.41.0 released today, but I think it actually behaves in 1.41 as it should have been from the start given this line in the rust CI:
> 
> https://github.com/huggingface/tokenizers/blob/6d532fedb1d3626328828304a5c39807733d2fa1/.github/workflows/rust.yml#L56
> 
> From my understanding, the `-D warnings` tells clippy to reject all warnings, which wasn't the case until now in the facts. So the possibilities are (for me):
> 
>     * Fix all the clippy warnings
> 
>     * Remove the `-D warnings` flag
> 
>     * Pin to rust 1.40.0
> 
> 
> wdyt @n1t0?

cc @mfuntowicz since I see you got the same problem x)

Issue #79, opened 2020-01-17
Add Template post-processor


mfuntowicz:
Roberta tokenizer requires inputs (& pairs) to be wrapped around `<s> ... </s>` (respectively cls and eos tokens).

n1t0:
Instead of adding a PostProcessor specific to Roberta, what about adding a TemplateProcessor that takes a list of tokens, and a template string, some like `[CLS] [A] [SEP] [SEP] [B] [EOS]` where A and B are the two sequences. We would be able to cover most cases with only one easily customizable PostProcessor this way.

PR #111, opened 2020-01-29
Roberta PostProcessor

mfuntowicz:
Roberta has a special handling when working with pairs, it will generate two separator tokens between pairs.

This PR implements the behaviour on Rust side and expose it  (tries to) for Python and Node bindings.


n1t0:
Looks good to me! I'm looking forward to having this: https://github.com/huggingface/tokenizers/issues/79 so that we can avoid having to duplicate this code.

Can you just fix the various checks before merging? Thank you @mfuntowicz!

Pierrci:
I pushed a little fix that prevented JS bindings to compile, and I couldn't help run a little lint, but otherwise good for me!

PR #112, opened 2020-01-29
rust: fix `OnlyFirst` and `OnlySecond` truncation strategies

### LINKED ISSUE #108: `OnlyFirst` and `OnlySecond` truncation strategies
The current behavior for `OnlyFirst` and `OnlySecond` truncation strategies is not the one I would expect, and diverges from the current behavior in [transformers](https://github.com/huggingface/transformers/blob/master/src/transformers/tokenization_utils.py#L1107-L1113):

https://github.com/huggingface/tokenizers/blob/88391dd18501ee672bee155a15f16632f2e0ae57/tokenizers/src/utils.rs#L87-L99

It currently takes only the first encoding (`OnlyFirst`) or the second one (`OnlySecond`), and then truncates it to make its length less than the desired `max_length`.

But this doesn't guarantee that the combined encodings have a length inferior to `max_length`, which is the behavior I was expecting: those strategies should take into account the combined encodings length when truncating only the first or second one.

What do you think @n1t0 @mfuntowicz?

Pierrci:
Fixes #108: behavior for `OnlyFirst` and `OnlySecond` truncation strategies aligned with current `transformers` behavior.

They now take into account the length of the combined encoding, potentially raising a new `SequenceShorterThanNumberTokensToRemove` error.

### DIFF
--- a/tokenizers/src/utils.rs
+++ b/tokenizers/src/utils.rs
@@ -25,7 +25,7 @@ pub enum PaddingStrategy {
 #[derive(Debug)]
 pub enum Error {
     SecondSequenceNotProvided,
-    SequenceShorterThanNumberTokensToRemove,
+    SequenceTooShort,
 }
 
 impl std::fmt::Display for Error {
@@ -34,9 +34,9 @@ impl std::fmt::Display for Error {
             Error::SecondSequenceNotProvided => {
                 write!(fmt, "Truncation error: Second sequence not provided")
             }
-            Error::SequenceShorterThanNumberTokensToRemove => write!(
+            Error::SequenceTooShort => write!(
                 fmt,
-                "Truncation error: Sequence to truncate shorter or equal than number of tokens to remove"
+                "Truncation error: Sequence to truncate too short to respect the provided max_length"
             ),
         }
     }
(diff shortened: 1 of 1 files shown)


Pierrci:
My first ever Rust PR @n1t0, so don't hesitate to point any problem! 😁🤞

tokenizers/src/utils.rs
```diff
@@ -93,8 +98,11 @@ pub fn truncate_encodings(
                 Err(Box::new(Error::SecondSequenceNotProvided))
             }?;
 
-            if target.get_ids().len() > params.max_length {
-                target.truncate(params.max_length, params.stride);
+            let target_len = target.get_ids().len();
+            if target_len > to_remove {
+                target.truncate(target_len - to_remove, params.stride);
+            } else {
+                return Err(Box::new(Error::SequenceShorterThanNumberTokensToRemove));
```
n1t0: What do you think about `SequenceTooShort`? I would also change the message in the `Display` implementation to say something like `Sequence to truncate too short to respect the provided max_length`.

Pierrci: Oh yeah sure, I really thought this name was too long but couldn't find anything better ^^' 👍

julien-c: was the name of the option longer than the number of tokens to remove? 😅

PR #117, opened 2020-01-31
python: fix inverted normalized/original string range

Pierrci:
If my understanding is correct, there was an inversion between the original and normalized string range (failing test related to https://github.com/huggingface/tokenizers/pull/109#issuecomment-580417438).


n1t0:
Oops! Thanks for the fix!

PR #124, opened 2020-02-03
rust: fix padding on overflowings

Pierrci:
Shadowing of `pad_length` was making it useless on overflowings


n1t0:
Oh yes! Nice catch!

PR #126, opened 2020-02-04
node: expose tokenizer configuration / truncation / padding

Pierrci:
Add a few new APIs to Node bindings:
- `configuration` on tokenizers implementations
- `truncation` and `padding` configurations, which are now returned by `setPadding` and `setTruncation` methods

### DIFF
--- a/bindings/node/native/src/tokenizer.rs
+++ b/bindings/node/native/src/tokenizer.rs
@@ -342,7 +342,16 @@ declare_types! {
                 }));
             }
 
-            Ok(cx.undefined().upcast())
+            let params_object = JsObject::new(&mut cx);
+            let obj_length = cx.number(max_length as f64);
+            let obj_stride = cx.number(stride as f64);
+            let obj_strat = cx.string(strategy);
+
+            params_object.set(&mut cx, "maxLength", obj_length).unwrap();
+            params_object.set(&mut cx, "stride", obj_stride).unwrap();
+            params_object.set(&mut cx, "strategy", obj_strat).unwrap();
+
+            Ok(params_object.upcast())
         }
 
         method disableTruncation(mut cx) {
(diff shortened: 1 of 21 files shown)


bindings/node/native/src/tokenizer.rs
```diff
@@ -402,6 +411,20 @@ declare_types! {
...
 
+            let params_object = JsObject::new(&mut cx);
+            if let Some(max_length) = max_length {
+                let obj_length = cx.number(max_length as f64);
+                params_object.set(&mut cx, "maxLength", obj_length).unwrap();
+            }
+            let obj_pad_id = cx.number(pad_id);
+            let obj_pad_type_id = cx.number(pad_type_id);
+            let obj_pad_token = cx.string(&pad_token);
+            let obj_direction = cx.string(&direction);
```
Pierrci: @n1t0 I'm really puzzled here:
- I had to create the `params_object` before calling `tokenizer.with_padding`, given I was "losing" `direction` when calling it otherwise
- But... in [`setTruncation`](https://github.com/huggingface/tokenizers/pull/126/files#diff-b9e53414da9657bdf6a3b1db08f83eacR348) I was able to create the `params_object` _after_ the call to `tokenizer.with_truncation`, even though I was expecting the same error with `strategy` than with `direction`

Would you have an explanation of why I got an error in one case, and not in the other? 🤯

tokenizers/src/utils.rs
```diff
@@ -50,6 +50,16 @@ pub enum TruncationStrategy {
     OnlySecond,
 }
 
+impl std::convert::AsRef<str> for TruncationStrategy {
```
Pierrci: Also I added that to get a string representation of the enum (same for `PaddingDirection`), my first ever trait implementation, obviously feel free to let me know if there's a better way!

bindings/node/native/src/tokenizer.rs
n1t0: Ok, so what happens here is that when you create a `PaddingParams` or a `TruncationParams`, you are giving it ownership of all the properties. So when you give it the direction or the strategy during creation, you are not supposed to be able to use them anymore. In these cases, if you want to use some variable again later, you are supposed to clone it. 
Now, there is actually an exception: when a variable is `Copy`, if you give it somewhere and keeps using it afterwards, then it will simply get copied. No need to clone it manually.
So the difference between `TruncationStrategy` and `PaddingDirection` comes from the fact that `TruncationStrategy` derives `Copy`, but not `PaddingDirection`.
https://github.com/huggingface/tokenizers/blob/d4fc17d5bc487127df299a76234b0d4f886ff200/tokenizers/src/utils.rs#L46-L51
https://github.com/huggingface/tokenizers/blob/d4fc17d5bc487127df299a76234b0d4f886ff200/tokenizers/src/tokenizer/encoding.rs#L4-L8

Pierrci: Haaaaaaaaaaaaaaaaaa that's it, it's the `Copy`, thank you!!!!!
I got that it was an ownership problem, but couldn't figure why there wasn't any with `TruncationStrategy`!

Any reason why in the `TruncationStrategy` case it derives `Copy` and `PartialEq` and not in `PaddingDirection`? Should I derive `Copy` for the latter one too to make code a bit cleaner?

Pierrci: OK so from https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy: 

> if your type can implement Copy, it should

I pushed a922560 with this modification, let me know if it's ok to merge for you!

Pierrci:
Also I just noticed that the rust build for the python part is only running for a python release: in the 41fee6d only the JS build failed, but there were errors on the python side too which were not catched by the CI (see 3adf199) @n1t0 @mfuntowicz.

PR #128, opened 2020-02-05
Maintenance : simplifications & update

huitseeker:
Mostly cleans up a few pattern matches (opinions on these refactorings held loosely, happy to acoomodate review).


n1t0:
Thank you @huitseeker! This is indeed way better!

PR #129, opened 2020-02-06
Initialize Normalizer using .new, not constructor

ljos:
Normalizers are initialised using .new and not __init__

### DIFF
--- a/bindings/python/tokenizers/normalizers/__init__.py
+++ b/bindings/python/tokenizers/normalizers/__init__.py
@@ -25,4 +25,4 @@ def unicode_normalizer_from_str(normalizer: str) -> Normalizer:
                 .format(normalizer, NORMALIZERS.keys())
         )
 
-    return NORMALIZERS[normalizer]()
+    return NORMALIZERS[normalizer].new()


mfuntowicz:
Hi @ljos, thanks for looking at this. 

The fix you propose seems perfectly legit, but I wonder if we should go the other way:

Use the tokenizers python binding files structure (*.py) to expose the constructor `__init__()` on all the classes we have which will internally call `new()`.

This would look more pythonic I think and less error prone ? 

What do you think @ljos @n1t0 ?

n1t0:
I’d be happy to review a PR on this, as I feel it would be better indeed. I am totally ok with the current fix too though.

Thank you very much @ljos for noticing this!

mfuntowicz:
Let's fix this way, I can do the __init__ stuff early next week then.

ljos:
I am happy to amend the PR to make the calling convention more pythonic. It
is a bit more involved. It might require editing __call__ or __new__ on the
class or some other small trickery. There could perhaps be something I can
do in the interop layer. I will work on it tomorrow when I get back to the
office.

I try to do the smallest edits as I can the first time I contribute to a
new project as I don’t know my involvement yet, and I want to minimize the
maintenance cost of my contribution.

PR #131, opened 2020-02-07
Python: Implement __new__ to make constructor available

ljos:
We implement `__new__` to make constructors for these classes available in Python
   - Normalizers
   - Trainers
   - Decoders
   - PostProcessors
   - PreTokenizers

This also makes the objects the correct class on inspection from python.

Closes #129

### DIFF
--- a/bindings/python/src/decoders.rs
+++ b/bindings/python/src/decoders.rs
@@ -26,25 +26,25 @@ impl Decoder {
     }
 }
 
-#[pyclass]
+#[pyclass(extends=Decoder)]
 pub struct ByteLevel {}
 #[pymethods]
 impl ByteLevel {
-    #[staticmethod]
-    fn new() -> PyResult<Decoder> {
-        Ok(Decoder {
+    #[new]
+    fn new(obj: &PyRawObject) -> PyResult<()> {
+        Ok(obj.init(Decoder {
             decoder: Container::Owned(Box::new(tk::decoders::byte_level::ByteLevel::new(false))),
-        })
+        }))
     }
 }
 
-#[pyclass]
+#[pyclass(extends=Decoder)]
 pub struct WordPiece {}
 #[pymethods]
 impl WordPiece {
-    #[staticmethod]
+    #[new]
     #[args(kwargs="**")]
-    fn new(kwargs: Option<&PyDict>) -> PyResult<Decoder> {
+    fn new(obj: &PyRawObject, kwargs: Option<&PyDict>) -> PyResult<()> {
         let mut prefix = String::from("##");
 
         if let Some(kwargs) = kwargs {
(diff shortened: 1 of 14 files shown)


ljos:
There are a few things I am a little uncertain of. I don't have much rust experience and I have never done Rust-Python interop before. It seems to me that what I do with `obj.init` makes the correct objects, but I don't know if there is a cleaner way to achieve the same.

There might be some other opportunities to make the code more pythonic, but I think this PR is getting large enough.

n1t0:
Thank you @ljos for this PR. Looks really good! For some reason, I wasn't able to make it work last time I tried.

I think the first call to `obj.init` with the subclass is not needed for this to work. If I understand correctly, the first instance gets replaced by the second one anyway.

ljos:
Glad to help!

If `obj.init` is not called with the subclass first, how are the methods and attributes of the subclass added to the object? I just assumed that was what that call did, so I never questioned it. I will have a look to see if I can remove it.

ljos:
The obj.init on the subclass did indeed not do anything. At least not in the configuration that we have now. The methods to get added to the object, but I saw that attributes do not--even if I did obj.init twice. 

There is probably some way to fix this, but as it doesn't affect this project, I suggest that rather upgrade pyo3. They have changed the api to make inheritance (imo) more intuitive. However, I think that should probably be a new PR.

n1t0:
Indeed, I suspect that it is because we should call the `new` method on the parent struct (from the subclass `new`), giving it the `obj`. I remember now that this is what I didn't get it to work. Since `new` can only have arguments that are convertible to and from Python, we cannot really give it the actual `Decoder` or `PreTokenizer` etc...

The new API provided by PyO3 is indeed way better, but I'd like to wait for it to become a bit more stable first. The current release is the first alpha.

In the meantime, I think this still improves the API, making it more Pythonic. We can't have subclass specific properties, but it was already true before this PR since we were returning an instance of the parent class. So I think, we can probably merge it anyway, and improve it as soon as PyO3 releases a more stable version. What do you think @ljos? Did I miss anything?

ljos:
I agree. Since we don't have any properties on the subclasses (as of now) I don't think we should worry about it. The interface to python has been improved and we have a fix planned.  The only risk I see is that we might forget.

n1t0:
Thank you again @ljos. This is awesome, we really appreciate your work!

To avoid forgetting about it, we can probably start a new PR implementing this with the alpha version and we'll merge it later. If you are interested in doing it, feel free to do so!

ljos:
Cool! Thanks for accepting the PR.

I’m quite interested in learning more about rust/python interop: I will try
to write a PR for the new pyo3 API. I don’t think there should be that many
changes to get something working.

Issue #115, opened 2020-01-30
[bug] BertWordPieceTokenizer, different tokenization than transformers around char `–`


knuser:
`BertWordPieceTokenizer` from *tokenizers* produces different encoding than `BertTokenizer` from *transformers* around char `–`. Tokenization around out-of-ASCII chars could be affected same way.

```
content = "first epoch after the  Cretaceous–Paleogene extinction event 66 million years ago"
tokenizer = BertTokenizer.from_pretrained('bert-base-cased')
fast_tokenizer = BertWordPieceTokenizer('./bert-base-cased/vocab.txt', lowercase=False, clean_text=False, strip_accents=False, add_special_tokens=False)

print(fast_tokenizer)
print(tokenizer.tokenize(content))
print(fast_tokenizer.encode(content).tokens)
```
Output:
```
Tokenizer(vocabulary_size=28996, model=BertWordPiece, add_special_tokens=False, unk_token=[UNK], sep_token=[SEP], cls_token=[CLS], clean_text=False, handle_chinese_chars=True, strip_accents=False, lowercase=False, wordpieces_prefix=##)
['first', 'e', '##po', '##ch', 'after', 'the', 'Cretaceous', '–', 'Pa', '##leo', '##gene', 'extinction', 'event', '66', 'million', 'years', 'ago']
['first', 'e', '##po', '##ch', 'after', 'the', 'Cretaceous', '##–', '##P', '##ale', '##ogen', '##e', 'extinction', 'event', '66', 'million', 'years', 'ago']
```

knuser:
I should mention here, that I did not check what output will produce original bert tokenizer from https://github.com/google-research/bert
So I'm not sure which version corresponds to original bert code, but I'm counting on exact same output.

n1t0:
This has been fixed by #134, thanks to @Mansterteddy. Feel free to reopen if you notice anything!

Issue #83, opened 2020-01-17
Different results compared with python BertWordPieceTokenizer


Mansterteddy:
Hi Team,

Thanks for your great work! Recently I tested BertWordPieceTokenizer in this repo on my Multi-Lingual Dataset, compared with python BertWordPieceTokenizer, the results are different on some cases, I found two types:

* Punctuation Type:
    * Sentence: 台北‧08月20日
    * Python Results: '台', '北', '‧', '08', '月', '20', '日'
    * Rust Results: '台', '北', '‧', '##0', '##8', '月', '20', '日'

* [UNK] Type:
    * Sentence: application remained denied“
    * Python Results: 'application', 'remained', 'denied', '[UNK]'
    * Rust Results: 'application', 'remained', '[UNK]'

Are these differences expected?

Best Wishes!

Mansterteddy:
Add two cases:
* Punctuation Type:
    * Sentence: Del Mundo ... ¿Quién seleccionó las siete maravillas del mundo antiguo y cuándo lo hizo?
    * Python Results: 'Del', 'Mundo', '.', '.', '.', '¿', 'Qui', '##én', 'sel', '##ec', '##cionó', 'las', 'siete', 'mara', '##villa', '##s', 'del', 'mundo', 'antiguo', 'y', 'cu', '##ánd', '##o', 'lo', 'hizo', '?'
    * Rust Results: 'Del', 'Mundo', '.', '.', '.', '¿', '##Q', '##ui', '##én', 'sel', '##ec', '##cionó', 'las', 'siete', 'mara', '##villa', '##s', 'del', 'mundo', 'antiguo', 'y', 'cu', '##ánd', '##o', 'lo', 'hizo', '?'

* [UNK] Type:
    * Sentence: Explore iPhone, the world\u2019s most powerful personal device.
    * Python Results: 'Ex', '##plo', '##re', 'iPhone', ',', 'the', 'world', '[UNK]', 's', 'most', 'powerful', 'personal', 'device', '.'
    * Rust Results: 'Ex', '##plo', '##re', 'iPhone', ',', 'the', '[UNK]', 'most', 'powerful', 'personal', 'device', '.'

epwalsh:
Hey @Mansterteddy, could you share your code to reproduce these results?

Mansterteddy:
Firstly, I use multi-lingual vocab from BERT as vocabulary.

For Rust Results, you can refer following code:
```
[...]

PR #134, opened 2020-02-11
Add is_bert_punc function.

Mansterteddy:
Add is_bert_punc function to make the tokenized results align with [HuggingFace's Python Implementation](https://github.com/huggingface/transformers/blob/master/src/transformers/tokenization_bert.py).

This update can solve following issue:
* https://github.com/huggingface/tokenizers/issues/83
* https://github.com/huggingface/tokenizers/issues/115


n1t0:
Awesome! Thank you @Mansterteddy.

Issue #135, opened 2020-02-11
LNK1112 Error on Windows


Mansterteddy:
Hi,

I can build python binding on Ubuntu, however, when I want to build python binding on Windows, I always met this error: **_fatal error LNK1112: module machine type 'x64' conflicts with target machine type 'x86'_**, I have changed the toolchain from nightly to nightly-2019-11-01, but it doesn't work.
 Do you have any idea?

Best Wishes!

n1t0:
What is your process to compile on windows? Did you run `cargo build`?

Mansterteddy:
Thanks for your reply! On Windows, I run `cargo build` under tokenizers folder, it works well. But when I run `python setup.py install` under python folder, it outputs above error.

mfuntowicz:
Hi @Mansterteddy , 

Do you observe the same error by running `pip install .` from the python folder ?

Mansterteddy:
Hi @mfuntowicz ,

Thanks for your reply! Yes, I observe the same error:
**_fatal error LNK1112: module machine type 'x64' conflicts with target machine type 'x86'_**
**_error: cargo failed with code: 101_**

`pip install . `outputs another line:
**_ERROR: Could not build wheels for tokenizers which use PEP 517 and cannot be installed directly_**

Do you have any idea?

Best Wishes!

n1t0:
I just got the same problem while trying to build for Python 32-bit on windows. I was able to fix it by making sure that both Python and the selected Rust toolchain matched:
- If you build for a python 32-bit version you should use rust `nightly-i686-pc-windows-msvc`.
- For python 64-bit you should use rust `nightly-x86_64-pc-windows-msvc`

Let me know how it goes!

Mansterteddy:
It works, thanks a lot!

PR #136, opened 2020-02-11
Python - update pyo3 and start using new API

ljos:
This should work with the new API (if it stays the way it is) of pyo3 when it is ready.

### DIFF
--- a/bindings/python/CHANGELOG.md
+++ b/bindings/python/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unrelease]
+### [Changed]
+- [#136] Updated Pyo3 version
+
+### [Added]
+- [#136] Models can now be instantiated through object constructors.
+
+### [Removed]
+- [#136] Static methods `Model.from_files` and `Model.empty` are removed in favor of using
+constructors.
+
+### [How to migrate]
+- Change `Model.from_files` and `Model.empty` to use constructor. The model constructor should take
+the same arguments as the old methods.
+
 ## [0.7.0-rc3]
 
 ### Changed
--- a/bindings/python/Cargo.lock
+++ b/bindings/python/Cargo.lock
@@ -74,6 +74,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "cloudabi"
+version = "0.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "bitflags 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "console"
 version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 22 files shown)


n1t0:
Hey @ljos! PyO3 is now out of alpha. They recently released the `0.9.1`. Would you be willing to update this with the last version of `master`? Sorry, a lot of things have changed!

ljos:
I'll have a look

ljos:
I think I have gotten the gist of it. There are some things that are a little more difficult than others, though I think I will be able to fix them. I will look at it tomorrow or/and in the weekend.

ljos:
I haven't actually run  this as I get some linking error on my work computer. I am not sure what is up there. It does seem to compile as the GitHub action gives me the green checkmark, but I haven't been able to run it and see if it actually works. I don't see that there are any tests in the python part of the repo. There are the examples, but I don't see that those are run on the actions.

n1t0:
Thank you for the update @ljos, I'll have a look, probably today. Unfortunately, there aren't any tests on the Python side for now, we plan on adding this really soon though.

If you are trying to build on windows, and have a link error there, this might help you: https://github.com/huggingface/tokenizers/issues/135

ljos:
I am on macOS and I think it might be an update issue from homebrew or something...

ljos:
pip install -e works, but cargo build doesn't. Probably something strange about my environment. I found an error. The creation of a new empty model doesn't work the way I thought it should so that leaves a type error right now. I have a solution that works, but I am trying to find a solution that keeps the type information. What I am trying to do is to make BPE.empty() returns a BPE model and not a Model.

ljos:
I the commit I just pushed I fix this issue, but I am not completely satisfied as it removes type information. I will see what I can do tomorrow.

n1t0:
Oh I see. `cargo build` also fails sometimes for me, I think it might be expected. I found the easiest dev setup was to use `maturin develop [--release]` (you can install it with pip). I also have some problems with `pip` and `python setup.py [install | build]` sometimes.

I quickly looked at the problem for `from_files` and `empty` and I think we can just use `#[new]` with them too. It should then work just like for any other constructor.

I also expect to merge the python tests (https://github.com/huggingface/tokenizers/pull/213) soon, so we'll be able to test all of this!

ljos:
Changing the models to use the new initialiser does look nicer to my eyes. Review my latest commit, and if you a agree I'll change the type data etc in python to match the new API. Maybe you want to still keep the old API?

n1t0:
At first, I wanted to keep the old API because I thought it would be easier to add more constructors, for example, to initialize directly from memory with the vocab/merges dicts, ...
But in the end, I also find this API way nicer and more Pythonic, and we could probably use exactly the same one to provide dicts instead of strings. What do you think?

ljos:
Yes, I think if we wanted to we could have `vocab` and `merges` be either a string/file or a dict. We could also have arguments for providing dict and another set for providing files. If we want different construction methods I think we should either use subclassing or a builder pattern (like it is done in the main library).

n1t0:
:+1: I'm really looking forward to merging this. I love the way PyO3 is evolving.

So, I think we miss the following before we can merge:
 - [ ] Update typings
 - [ ] Update tests to check that it keeps the initial type: example for BPE: `assert isinstance(BPE(), BPE)` along with the `isinstance(BPE(), Model)`
 - [ ] Update the CHANGELOG.md

n1t0:
Thank you @ljos!

PR #146, opened 2020-02-13
Expose num_added_tokens on Python side 

mfuntowicz:
without the need to pass an Encoding to added_tokens.

This allows to compute the max sentence length for single/pair inputs without actually the need to have an Encoding structure.
As the number of added tokens is fixed and static during compilation it allows more flexible usage of the method.

### DIFF
--- a/bindings/python/src/tokenizer.rs
+++ b/bindings/python/src/tokenizer.rs
@@ -38,6 +38,13 @@ impl Tokenizer {
         }
     }
 
+    fn num_added_tokens(&self, is_pair: bool) -> PyResult<usize> {
+        Ok(self.tokenizer
+               .get_post_processor()
+               .map_or(0, |p| p.as_ref().added_tokens(is_pair))
+        )
+    }
+
     #[args(kwargs = "**")]
     fn get_vocab_size(&self, kwargs: Option<&PyDict>) -> PyResult<usize> {
         let mut with_added_tokens = true;
--- a/bindings/python/src/tokenizer.rs
+++ b/bindings/python/src/tokenizer.rs
@@ -38,7 +38,7 @@ impl Tokenizer {
         }
     }
 
-    fn num_added_tokens(&self, is_pair: bool) -> PyResult<usize> {
+    fn num_special_tokens_to_add(&self, is_pair: bool) -> PyResult<usize> {
         Ok(self.tokenizer
                .get_post_processor()
                .map_or(0, |p| p.as_ref().added_tokens(is_pair))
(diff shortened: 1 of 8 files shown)


bindings/python/src/tokenizer.rs
```diff
@@ -38,6 +38,13 @@ impl Tokenizer {
         }
     }
 
+    fn num_added_tokens(&self, is_pair: bool) -> PyResult<usize> {
```
n1t0: I'm not sure it really makes sense directly exposed on the `Tokenizer`. I'd expect this to be the number of added tokens that gets returned by `get_vocab_size(with_added_tokens=True)`.

This can simply be retrieved by doing `tokenizer.post_processor.num_added_tokens(is_pair)`

bindings/python/tokenizers/implementations/base_tokenizer.py
```diff
@@ -13,6 +13,14 @@ def __repr__(self):
             self._tokenizer.get_vocab_size(),
             ', '.join(k + '=' + str(v) for k, v in self._parameters.items()))
 
+    def num_added_tokens(self, is_pair: bool) -> int:
```
n1t0: Same comment here. I find this really confusing.

bindings/python/src/tokenizer.rs
mfuntowicz: I exposed this method on Tokenizer for the case where there is no post_processor. 
In this case, it requires to check if post_processor is not null before accessing the property and returning 0 if null.

Having this wrapped on the Rust side doesn't require such checking to be done in every bindings we would have.

bindings/python/tokenizers/implementations/base_tokenizer.py
mfuntowicz: As the `_tokenizer` property is "kinda" private on `BaseTokenizer`, it's cleaner to forward the method on `BaseTokenizer` and not expose the `_tokenizer` property. 

Its exactly the same behavior as for `get_vocab_size`

bindings/python/src/tokenizer.rs
n1t0: I understand. My main concern is just about having `add_tokens` that does something, and `num_added_tokens` just right beside it, doing something completely different. If we find the right name, we can add it here!

mfuntowicz: What about just keeping added_tokens everywhere ?

n1t0: As discussed, let's use `num_special_tokens_to_add`

PR #162, opened 2020-02-21
Make parameter name on Model.save() optional.

mfuntowicz:

### DIFF
--- a/bindings/python/src/models.rs
+++ b/bindings/python/src/models.rs
@@ -30,14 +30,11 @@ impl Model {
             for (key, value) in kwargs {
                 let key: &str = key.extract()?;
                 match key {
-                    "name" => {
-                        name = value.extract()?
-                    }
+                    "name" => name = value.extract()?,
                     _ => println!("Ignored unknown kwarg option {}", key),
                 }
             }
         }
-
 
         let saved: PyResult<Vec<_>> = ToPyResult(
             self.model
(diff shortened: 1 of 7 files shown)


bindings/python/src/models.rs
```diff
@@ -23,7 +23,19 @@ impl Model {
         ))
     }
 
-    fn save(&self, folder: &str, name: &str) -> PyResult<Vec<String>> {
+    #[args(kwargs = "**")]
+    fn save(&self, folder: &str, kwargs: Option<&PyDict>) -> PyResult<Vec<String>> {
```
n1t0: This is probably breaking as it changes to a named parameter. I'm not sure it can be specified as positional then. What do you think of having this instead?
```rust
#[args(name=None)]
fn save(&self, folder: &str, name: Option<&str>) -> PyResult<Vec<String>> {
```

bindings/python/tokenizers/implementations/base_tokenizer.py
```diff
@@ -209,14 +209,14 @@ def id_to_token(self, id: int) -> Optional[str]:
         """
         return self._tokenizer.id_to_token(id)
 
-    def save(self, directory: str, name: str):
+    def save(self, directory: str, name: str = None):
```
n1t0: I think it should be `Optional[str]`

n1t0:
Thanks @mfuntowicz!

PR #165, opened 2020-02-24
Big improvement over BPE speed (Training & Tokenization)

n1t0:
This PR includes various changes:
- Update the way we apply merges to each word during both Training and Tokenization (Improves the complexity from `O(n2)` to `O(n)`).
-  Add multithreading in some places. 

All these modifications are especially important for some languages where pre-tokenizing on spaces doesn't make sense, thus ending-up with very long words.

Here is a comparison, training and tokenizing a 100MB file in a few different languages:

### `master`:
| | English | Chinese | Japanese |
|-|-|-|-|
|Train|90.1s|437.7s|348.3s|
|Tokenization|6.3s|11.3s|12.7s|

### `faster-bpe`:
| | English | Chinese | Japanese |
|-|-|-|-|
|Train|20.5s (**4.4x faster**)|53.5s (**8.2x faster**)|59.6s (**5.8x faster**)|
|Tokenization|5.8s (**1.1x faster**)|5.0s (**2.3x faster**)|5.3s (**2.4x faster**)|

### DIFF
--- a/tokenizers/src/models/bpe/model.rs
+++ b/tokenizers/src/models/bpe/model.rs
@@ -3,7 +3,6 @@ use super::{
     DEFAULT_CACHE_CAPACITY,
 };
 use crate::tokenizer::{Model, Offsets, Result, Token};
-use rand::{thread_rng, Rng};
 use serde_json::Value;
 use std::{
     collections::HashMap,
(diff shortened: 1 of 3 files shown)


epwalsh:
This is great!!

Do you think it's worth having a test for the trainer? I'm thinking the test would just feed in a very small hash map of word counts to `BpeTrainer::train` so we can the resulting vocab and merges exactly. I could make a separate PR to master with that to master if you think it's worth it.

n1t0:
Definitely! That's a great idea, feel free to make a PR!

PR #169, opened 2020-02-24
test BPE trainer

epwalsh:
see https://github.com/huggingface/tokenizers/pull/165#issuecomment-590515297


n1t0:
That's great @epwalsh! I'm wondering why you changed the test with the `ensure test is deterministic` commit. I would expect it to work every time, doesn't it?

epwalsh:
@n1t0 sorry, that was a bad way to describe the commit.

Before that commit I think the test was in fact deterministic, dependent on `HashMap::iter` being consistent with the order of the key-value pairs in the iterator produced, since this would determine which indices are assigned to which token in the vocab.

However I don't think the test should rely on the ordering of a `HashMap` iterator since the order is arbitrary and could potentially change with future releases of Rust.

What do you think?

n1t0:
I think I liked the first version better because it was actually testing the IDs too. We try to make it deterministic during the generation of the IDs assigned to each token, so it would be awesome to actually know if we were to break this. (cf [here](https://github.com/huggingface/tokenizers/blob/master/tokenizers/src/models/bpe/trainer.rs#L256))

I think testing the equality on the HashMap gives the result you first expected: https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#969-983

What do you think?

epwalsh:
@n1t0 ahhh I see! I overlooked [this sorting](https://github.com/huggingface/tokenizers/blob/master/tokenizers/src/models/bpe/trainer.rs#L256) before. I'll revert

n1t0:
Perfect! Thank you @epwalsh

PR #172, opened 2020-02-26
Update node bindings

### LINKED ISSUE #170: new release for node-binding and general question w.r.t. node releases
The last release (0.4.1) of node binding is from 14 days ago, so it does not contain the latest fixes. Could you please release a new version for node as well?

What is the versioning scheme used for tokenizers? The rust lib is at 0.7, the python binding at 0.5.2.

Ideally, a new release of both node and python bindings should be released when the rust library is updated, right?

Pierrci:
This PR incorporates into the node bindings the changes from the last 2 weeks made in the core. Also adds an all-new `Encoding` JS class which shadows the internal one and exposes getters (#BREAKING 🎉).

Can you check the changes in the rust files @n1t0 when you have time? (you can check the rest too obviously, I'm just more confident on the TS part! 😁)

Closes #170


Pierrci:
The shadowing of the `Encoding` object made me realize we didn't check properly for explicit `undefined` in a bunch of functions with optional params, so the last commit fixes that!

PR #183, opened 2020-03-04
Added Rust code to train from word count (not made available in Python)

chrhad:
Training from tab-separated token count, i.e., token \t count is not made available yet. Would appreciate this feature can be integrated to the master branch

### DIFF
--- a/tokenizers/src/train.rs
+++ b/tokenizers/src/train.rs
@@ -1,0 +1,100 @@
+use std::fs::File;
+use std::io::{self, BufRead};
+use std::path::Path;
+use std::collections::HashMap;
+use clap::{value_t, Arg, App};
+use tokenizers::tokenizer::Model;
+use tokenizers::models::bpe::BpeTrainer;
+
+fn main() {
+    let matches = App::new("BPE Trainer")
+        .version("0.1.0")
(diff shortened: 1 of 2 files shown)


tokenizers/src/train.rs
```diff
@@ -0,0 +1,100 @@
...
+    let wc_fname = matches.value_of("word_count").unwrap().to_string(); // files containing word count
+    let out_dirname = matches.value_of("output_dir").unwrap().to_string(); // directory name to store output
+    let vocab_size = value_t!(matches, "vocab_size", usize).unwrap_or(0);
+    println!("| Vocabulary size is {}", vocab_size);
+    let min_freq = value_t!(matches, "min_freq", u32).unwrap_or(0);
+    println!("| Minimum frequency is {}", min_freq);
+    
+    // Instantiate trainer builder
+    let trainer_builder = BpeTrainer::builder();
+    let mut tokens: Vec<String> = Vec::new();
```
usuyama, line 60: FYI - this needs to be `std::vec::Vec<tokenizers::AddedToken>`

n1t0:
Thank you but we won't go forward with this:
 - The word count interface for training is very unlikely to stay.
 - Tokenizers should be customized according to the use case being solved, and we are unlikely to provide generic scripts to do this.

PR #188, opened 2020-03-05
Fix byte level BPE offsets

### LINKED ISSUE #156: ByteLevelBPETokenizer | Some unicode characters cause the token position to not match the char position
Tested using: Ubuntu 18.04, Python 3.7 and tokenizers==0.5.0

I've found a small problem with one of the tokenizers, to recreate it run:
```
from tokenizers import ByteLevelBPETokenizer
text = "i ⭢ j"
with open("/tmp/tmp.txt", "w") as f:
    f.write(text)
tokenizer = ByteLevelBPETokenizer()
tokenizer.train("/tmp/tmp.txt")

doc = tokenizer.encode(text)
print(doc.tokens)
print(doc.offsets)
print(len(text))


### Outputs
"""
['i', 'Ġ', 'â', 'Ń', '¢', 'Ġ', 'j']
[(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6), (6, 7)]
5
"""
```

As you can see from the prints the offsets are strange and the tokenized text is longer than the original. I'm not sure if this is desired, but it creates many problems when mapping the tokens back to the original string. Also, it does not happen with other tokenizers e.g. the BertWordPieceTokenizer.

n1t0:
- Fix a bug where adding a prefix space actually messes up with the offsets
- Also fixes wrong offsets when a Unicode character gets split up in multiple byte-level characters (Fix #156)
- Add the ability to trim offsets from the ByteLevel to avoid including whitespaces.

@Pierrci I added the rust part of the Node bindings, but I prefer to let you handle the Typescript part. Is it ok for you?

### DIFF
--- a/bindings/node/lib/bindings/normalizers.d.ts
+++ b/bindings/node/lib/bindings/normalizers.d.ts
@@ -30,15 +30,6 @@ export interface BertNormalizerOptions {
 }
 
 /**
- * Instantiate a ByteLevel Normalizer
- *
- * @param [addPrefixSpace=true] Whether to add a space to the first word if there isn't already one.
- * This allows to treat `hello` exactly like `say hello`.
- * @since 0.6.0
- */
-export function byteLevelNormalizer(addPrefixSpace?: boolean): Normalizer;
-
-/**
  * Instantiate a Bert Normalizer with the given options
  *
  * @param [options] Normalizer options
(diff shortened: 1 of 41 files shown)


Pierrci:
Do you plan to merge this one soon @n1t0? (do I wait for it before releasing a new Node version?)

n1t0:
You should probably wait a bit yes. Other things are coming :-)

n1t0:
Awesome! Thanks for the fix @Pierrci!

PR #190, opened 2020-03-06
Improve models builders

n1t0:
- `from_files` now returns a builder, and cannot fail
- All possible failures are handled during the `build` phase of each builder
- This also moves any I/O to the `build` phase. This behavior seems more intuitive: we first construct our builders with all their respective configurations and then finally build at the end, doing the real work only then.
- This simplifies both python and node bindings where we don't need to have any default values.
- Add the `continuing_subword_prefix` option to both bindings for WordPiece.


epwalsh:
I like this! I think it's definitely a good convention.

Pierrci:
👍

PR #196, opened 2020-03-12
Throw a more meaningful error when provided python input is None.

mfuntowicz:
Currently is just throw `TypeError` without any other indication. This PR adds a more meaningful message and match the `ValueError`we have on transformers side.


n1t0:
Nice, thanks @mfuntowicz

PR #197, opened 2020-03-16
Remove NormalizedString from Encoding

n1t0:
Reasons for the removal:
 - The `NormalizedString` embedded on the `Encoding` accounts for 70% of the memory usage. 
 - It prevents from creating an `Encoding` without having this piece of information.
 - In order to remove it, we make the offsets relative to the original string, this will be easier to grasp for users.

The normalized string can still be retrieved if relevant, for debugging purposes or whatever.

This PR also includes a first introduction of integration tests.

Missing before merge:
 - [x] `NormalizedString` should be indexed by offsets on the original string.
 - [x] Node typings.


julien-c:
> In order to remove it, we make the offsets relative to the original string, this will be easier to grasp for users.

Yay 🎉 finally 😅

Issue #175, opened 2020-02-27
Can not add large custom vocab to RobertaTokenizerFast


DarshanPatel11:
I have around **65K** tokens in my custom vocab file and I want to include in the DistilRoberta model to finetune it on my corpus.  I added it on RobertaTokenizer and it was added successfully, but the Tokenization process is very slow. Even with using multiprocessing it is showing that It will take around **200 hrs**.  So I tried to use RobertaTokenizerFast but couldn't add my custom vocab. It failed with an error as follows:
```thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: CompiledTooBig(10485760)', /__w/tokenizers/tokenizers/tokenizers/src/tokenizer/mod.rs:672:22```
So is there any workaround to add my custom vocab and fast tokenization.
Also, I have a question that Roberta/GPT2 has a "Ġ" character in the vocab, so should I add it as a prefix to my vocab, or not?

ameet-1997:
This issue is still not fixed. It throws the exact same error even with the newest version of tokenizers.

Adding up to `25000` tokens works. Anything more than that doesn't work.

ameet-1997:
@n1t0 , just in case this is useful.

n1t0:
The associated PR increased the capacity a lot, but it is not unlimited. Unfortunately, I believe there is no easy way to increase it even more. I would advise against adding so many tokens anyway as it would make the tokenizer extremely slow.

What kind of use-case makes you add so many tokens?

SokolovYaroslav:
Hi everyone!

I'm trying to fit a custom BPE tokenizer on ~87 million words to make a vocabulary with a size 16384. And I got the same exception: `CompiledTooBig(10485760)`. Also, I've added 1041 special tokens as an argument for the `BpeTrainer`.
I'm training the BPE with the `train_from_iterator` method.

I don't think that my case is somehow large, so I'd rather think that exception isn't expected.

kaankork:
Hi everyone, I'm experiencing the same issue with around 100K new words to add. 
It seems that one possible workaround would be to take the words in a batch of 25000 each time (since @ameet-1997 mentioned this is the limit) and keep iterating until all of them are added.

pjuangph:
How is this issue closed? there's no resolution

catsbergers:
I also have this issue.  Batching to under 25k does not resolve it.

Narsil:
@catsbergers Do you mind providing a script to reproduce ?

[...]

PR #202, opened 2020-03-20
Improve Management of added tokens

n1t0:
- Fix a bug where offsets were wrong if an added token was part of the sequence to encode (cf https://github.com/huggingface/transformers/issues/3058#issuecomment-592781499)
- We can now have thousands of added tokens (even if not advised, but that's not the question). (Fix https://github.com/huggingface/tokenizers/issues/175)
- Added tokens can have a customized behavior. This will allow covering cases like: the `<mask>` special token for Roberta does not expect the previous whitespace to be kept in the tokenization. It should be stripped out.

Still missing:
 - [x] `Trainer`s should be able to pass `AddedToken`s instead of `String` to include options
 - [x] Python bindings
 - [x] Node bindings
 - [x] More tests

### DIFF
--- a/bindings/node/lib/bindings/tokenizer.d.ts
+++ b/bindings/node/lib/bindings/tokenizer.d.ts
@@ -85,21 +85,20 @@ export class Tokenizer {
    * Add the given tokens to the vocabulary
    *
    * @param tokens A list of tokens to add to the vocabulary.
-   * Each token can either be a string, or a tuple with a string representing the token,
-   * and a boolean option representing whether to match on single words only.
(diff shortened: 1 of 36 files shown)


Pierrci:
@n1t0 There is a bunch of tests in JS that fail after updating the bindings, but I don't see what is wrong with my code, can you have a look? (though I have the feeling it's rather related to some internal changes in the rust source, but maybe it's me ^^')

n1t0:
Indeed, there was a bug in the method `add_tokens` as I forgot to update something. This is now fixed so the remaining failing tests are expected and actually part of the current fix!

PR #212, opened 2020-03-30
Fix typos in JsAddedToken init in Node binding

thirdwing:
There seem typos in parameter setting.

I think we want to set left/right strip, but we set `single_word` three times.


Pierrci:
Indeed, thank you for catching this!

PR #213, opened 2020-03-30
Add Python tests

n1t0:
Add the Python tests to make sure we don't break the bindings.

One problem of this setup is that we can't test whatever version is in the virtualenv, but only the one living in the repository. In order to fix this we will have to move all the `bindings/python/tokenizers` content under `bindings/python/src/tokenizers` and the Rust's sources somewhere else. Anyway, this requires too many modifications to be included in this PR.

Missing:
 - [x] Run tests in the CI
 - [x] Test the raw bindings (normalizers, decoders, ...)
 - [x] Test the implementations

### DIFF
--- a/bindings/python/tests/bindings/test_tokenizer.py
+++ b/bindings/python/tests/bindings/test_tokenizer.py
@@ -9,7 +9,7 @@ class TestAddedToken:
         assert str(added_token) == "<mask>"
         assert (
             repr(added_token)
-            == 'AddedToken("<mask>", rstrip=false, lstrip=false, single_word=false)'
+            == 'AddedToken("<mask>", rstrip=False, lstrip=False, single_word=False)'
         )
         assert added_token.rstrip == False
         assert added_token.lstrip == False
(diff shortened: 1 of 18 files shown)


bindings/python/tests/bindings/test_tokenizer.py
```diff
@@ -0,0 +1,221 @@
...
+        added_token = AddedToken("<mask>", lstrip=True)
+        assert added_token.rstrip == False
+        assert added_token.lstrip == True
+        assert added_token.single_word == False
+
+    def test_can_set_single_world(self):
+        added_token = AddedToken("<mask>", single_word=True)
+        assert added_token.rstrip == False
+        assert added_token.lstrip == False
+        assert added_token.single_word == True
```
julien-c: Maybe you can use https://docs.python.org/3/library/unittest.html, it's pretty nice/canonical

julien-c: (you can use pytest as a test runner but define all the tests using unittest – we found it's the best combo)

n1t0: Would you have any information/reading material about this? What are the advantages of combining both instead of just using pytest?

PR #220, opened 2020-04-06
Fix overflowings offsets

Pierrci:
This fix ensures that the overflowing offsets are converted back to the original string before returning from `tokenizer.encode`, in addition to the "main" encoding offsets.

@n1t0 there might be a better way to code that in rust, so feel free to improve it :)

### DIFF
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -505,22 +505,30 @@ impl Tokenizer {
         let mut output = self.post_process(encoding, pair_encoding, add_special_tokens)?;
 
         // 5. Convert offsets back to original string
-        let mut current_offset = (0, 0);
-        let mut n_source = &normalized;
-        output
-            .get_offsets_mut()
-            .iter_mut()
-            .for_each(|(start, end)| {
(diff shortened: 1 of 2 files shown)


n1t0:
Nice catch! Maybe you can just extract the duplicated code in a closure, but otherwise looks good!

Pierrci:
I just forced-push with a closure :)

tokenizers/src/tokenizer/mod.rs
```diff
@@ -507,20 +507,28 @@ impl Tokenizer {
...
-                *start = s;
-                *end = e;
-            });
+
+        let mut convert_offsets_back = |encoding: &mut Encoding| {
+            encoding
+                .get_offsets_mut()
+                .iter_mut()
+                .for_each(|(start, end)| {
+                    if (*start, *end) < current_offset {
```
n1t0, line 516: I think the initial `n_source` should be inside the closure (and `current_offset` too I guess), otherwise, it will use the `pair_normalized` after the first one that needed it.

Pierrci: Oh yes you're right, I just forced-push again with the fix!

PR #222, opened 2020-04-07
Require Send for all parts of the tokenizer

llogiq:
This allows us to use a tokenizer from an async context.

The change itself was very trivial: Just add a `Send` bound to all parts of the tokenizer and then add it everywhere the compiler wants it.

### DIFF
--- a/bindings/node/native/src/decoders.rs
+++ b/bindings/node/native/src/decoders.rs
@@ -5,7 +5,7 @@ use neon::prelude::*;
 
 /// Decoder
 pub struct Decoder {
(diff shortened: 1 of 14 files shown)


n1t0:
Sure! I'm wondering if maybe we should add `: Send + Sync` to `PreTokenizer`, `Normalizer`, .. directly, where we declare these traits.

It would allow us to remove all the `Send + Sync` everywhere else I guess. What do you think?

llogiq:
Yeah, that'd certainly simplify the code. Should I do that right now or in a followup PR?

n1t0:
You can do that here!

n1t0:
Thank you @llogiq! Looks very good!

PR #226, opened 2020-04-09
Word indexes are None for special tokens

n1t0:
Before these changes, the indexes were wrong when there were special tokens.


n1t0:
cc @Pierrci @mfuntowicz

PR #229, opened 2020-04-11
Python - improve compatibility with sentencepiece in the conversion script

sdadas:
This fix improves compatibility between the original sentencepiece model and the model converted using `sentencepiece_extractor.py` script. It introduces two simple changes:
1. Merges are sorted according to the piece IDs.
2. Removes the condition `piece_l != piece_r` as it should be valid to merge two instances of the same piece.

I tested the conversion process on a dataset of 5000 sentences by comparing the result produced by the original sentencepiece model and the converted tokenizers model.
For the initial version of the script, 4021 (80%) sentences were mismatched.
For the modified version, the results were identical for all sentences.


n1t0:
Awesome thank you @sdadas. This looks indeed way better and should provide a good compatibility with BPE models.

PR #238, opened 2020-04-16
Fix offsets

Pierrci:
Arf ok I didn't see you had merged the PR already @n1t0, I was looking at a bug introduced by your changes, I changed the corresponding test so that it should fail now 😬


Pierrci:
The problem is at https://github.com/huggingface/tokenizers/blob/fix-offsets/tokenizers/src/pre_tokenizers/byte_level.rs#L255 (if `tl > offsets.1` boom since it's `usize` values for now). I was not sure of the best way to fix that other than casting to `isize`, so I prefer letting you take care of that @n1t0 :)

n1t0:
Indeed nice catch! Since offsets should never be < 0 it makes more sense to keep it as a `usize` but to avoid trimming in this case. I think it should happen only if we are trying to trim spaces that are both leading and trailing.

PR #239, opened 2020-04-16
Update CharLevelBPE to match GPT BPE

thomwolf:
This updates CharLevelBPE to match OpenAI GPT BPE, in particular by having a punctuation splitter instead of a whitespace splitter. We still keep the lowercasing deactivated by default.


n1t0:
Thanks @thomwolf!

PR #241, opened 2020-04-18
Python - Fix bug in bert wordpiece example script

jaymody:
`tokenizer.train()` does not have a return. Changed it to be consistent with the other example scripts.


n1t0:
Thanks @jaymody

PR #249, opened 2020-04-25
Allow pre-tokenized inputs to encode/encode_batch

n1t0:
Completely rewrote the `encode` and `encode_batch` interfaces to allow pre-tokenized inputs.
Now `encode` accepts:
 - Single sequences as well as pair sequences.
 - Each sequence can be either a raw string or a list of strings (pre-tokenized)

Python:
```python
encode("A single sequence")
encode(("A sequence", "And its pair"))
encode([ "A", "pre", "tokenized", "sequence" ])
encode(([ "A", "pre", "tokenized", "sequence" ], "And its pair"))

# and

encode_batch([
   "A single sequence",
   ("A tuple with a sequence", "And its pair"),
   [ "A", "pre", "tokenized", "sequence" ],
   ([ "A", "pre", "tokenized", "sequence" ], "And its pair")
])
```

Rust:
```rust
// Sequences:
tokenizer.encode("Single sequence", false);
tokenizer.encode(("Sequence A", "Sequence B"), false);
                                                                                  
// Pre-tokenized sequences:
tokenizer.encode(&["Single", "sequence"][..], false);
tokenizer.encode((
    &["Sequence", "A"][..],
    &["Sequence", "B"][..]
), false);
                                                                                  
// or even both types together:
tokenizer.encode(("A complete sequence", &["And", "a", "tokenized"][..]), false);
```

In Python, the changes for `encode` are breaking:
```python
# Before
tokenizer.encode(sequence: str, pair: Optional[str], add_special_tokens: bool = True)

# After
tokenizer.encode(input: EncodeInput, add_special_tokens: bool = True)
tokenizer.encode_batch(inputs: List[EncodeInput])
```
I think it makes more sense because any input in the list given to `encode_batch` can also be given to `encode` while it wasn't the case before.

Missing:
 - [ ] More tests
 - [ ] Node bindings (These are going to be tricky I guess...)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 *~
 
+.vim
 .env
 target
 
--- a/bindings/node/lib/bindings/raw-encoding.test.ts
+++ b/bindings/node/lib/bindings/raw-encoding.test.ts
@@ -3,7 +3,7 @@ import { promisify } from "util";
 import { Model, WordPiece, WordPieceOptions } from "./models";
 import { whitespacePreTokenizer } from "./pre-tokenizers";
 import { RawEncoding } from "./raw-encoding";
-import { Tokenizer } from "./tokenizer";
+import { EncodeOptions, InputSequence, Tokenizer } from "./tokenizer";
 
 const MOCKS_DIR = __dirname + "/__mocks__";
 
@@ -11,9 +11,9 @@ describe("RawEncoding", () => {
   const originalString = "my name is john";
   let encoding: RawEncoding;
   let encode: (
-    sequence: string,
-    pair: string | null,
-    addSpecialTokens: boolean
+    sequence: InputSequence,
+    pair?: InputSequence | null,
+    options?: EncodeOptions | null
   ) => Promise<RawEncoding>;
 
   beforeAll(async () => {
@@ -30,7 +30,7 @@ describe("RawEncoding", () => {
   });
 
   beforeEach(async () => {
-    encoding = await encode(originalString, null, false);
+    encoding = await encode(originalString, null);
   });
 
   it("has a list of defined methods", async () => {
--- a/bindings/node/lib/bindings/tokenizer.d.ts
+++ b/bindings/node/lib/bindings/tokenizer.d.ts
@@ -62,15 +62,28 @@ export interface PaddingOptions {
   padToken?: string;
 }
 
-/**
- * A list of tokens
- */
-export type TokenizedSequence = string[];
+export type TextInputSequence = string;
+export type PreTokenizedInputSequence = string[];
+export type InputSequence = TextInputSequence | PreTokenizedInputSequence;
 
-/**
- * A list of tokens, each associated with its offsets
- */
-export type TokenizedSequenceWithOffsets = [string, [number, number]][];
+export type TextEncodeInput = TextInputSequence | [TextInputSequence, TextInputSequence];
+export type PreTokenizedEncodeInput =
+  | PreTokenizedInputSequence
+  | [PreTokenizedInputSequence, PreTokenizedInputSequence];
+export type EncodeInput = TextEncodeInput | PreTokenizedEncodeInput;
+
+export interface EncodeOptions {
+  /**
+   * Whether the given sequence is pre-tokenized
+   * @default false
+   */
+  isPretokenized?: boolean;
+  /**
+   * Whether we should add special tokens
+   * @default true
+   */
+  addSpecialTokens?: boolean;
+}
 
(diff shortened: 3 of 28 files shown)


julien-c:
Are you using Tuple vs. List to differentiate between cases in the `encode` case? That seems pretty brittle, no?

Shouldn't pretokenized inputs only be allowed in `encode_batch`, and then use the level of nesting to figure things out?

```python
encode_batch([
   "A single sequence",
   ["A tuple with a sequence", "And its pair"],
   [[ "A", "pre", "tokenized", "sequence" ]],
   [[ "A", "pre", "tokenized", "sequence" ], "And its pair"]
])
```

n1t0:
I don't think so. It is actually strongly typed and differentiable. From my point of view, there is no ambiguity here. What makes you say it seems brittle?

I find the relation really beautiful actually:
```python
# Represents a sequence, pre-tokenized or not
InputSequence = Union[str, List[str]]
# Represents either a single sequence, or a pair of sequences
EncodeInput   = Union[InputSequence, Tuple[InputSequence, InputSequence]]

def encode(input: EncodeInput)
def encode_batch(inputs: List[EncodeInput])
```

julien-c:
What about the languages that don't have a tuple type?

n1t0:
Well I guess, in this case we'll have to either:
 - keep a method with two distinct arguments (example in typescript):
```typescript
type InputSequence = string | string[];
type EncodeInput = InputSequence | [InputSequence, InputSequence];
encode(sequence: InputSequence, pair?: InputSequence);
encodeBatch(inputs: EncodeInput);
```
 - have a custom type / class for example

thomwolf:
A note on that is that in `Transformers`, we've added a `is_pretokenized` flag to `encode` and `encode_batch` to explicitely say whether the input is tokenized or not to avoid Tuple/List mistakes from the user. The difference is very thin in python and I've noticed people tend to use indifferently both so I wanted to be a little bit safer for a widely used library such as `Transformers`.

(note that we assume both sequences in a pair input are in the same state, which is safe to assume).

julien-c:
Yes I agree with @thomwolf

n1t0:
I also agree! I actually didn't realize the difference between tuples and lists is so thin in python and that both are being used in similar cases. The user couldn't really make any mistake without knowing it as this is type-checked and an error would have been raised, but this is going to happen. Will update!

stefan-it:
@n1t0 Would this also fix the problem with `encode_batch_plus` that I've described in https://github.com/huggingface/transformers/issues/4015 🤔

n1t0:
After updating everything, here is the final API for Python and Node:

## Python
```python
encode("A single sequence")
encode("A sequence", "And its pair")
encode([ "A", "pre", "tokenized", "sequence" ], is_pretokenized=True)
encode([ "A", "pre", "tokenized", "sequence" ], ["And",  "its",  "pair"], is_pretokenized=True)

# and

encode_batch([
   "A single sequence",
   ("A tuple with a sequence", "And its pair"),
])
encode_batch([
   [ "A", "pre", "tokenized", "sequence" ],
   ([ "A", "pre", "tokenized", "sequence" ], ["And",  "its", "pair"])
], is_pretokenized=True, add_special_tokens=False)
```

## Node
```typescript
encode("A single sequence", null)
encode("A sequence", "And its pair")
encode([ "A", "pre", "tokenized", "sequence" ], null, { isPretokenized: true })
encode([ "A", "pre", "tokenized", "sequence" ], [ "And", "its", "pair" ], { isPretokenized: true })

# and

encodeBatch([
    "A single sequence",
    [ "A tuple with a sequence", "And its pair" ]
])
encodeBatch([
    [ "A", "pre", "tokenized", "sequence" ],
    [
        [ "A", "pre", "tokenized", sequence" ],
        [ "And", "its", "pair" ]
    ]
], { isPretokenized: true, addSpecialTokens: false })
```

n1t0:
There is a bug with the offsets generated for byte-level BPE when `trim_offsets=True`. This is an edge case and it shouldn't prevent us from shipping this. The offsets don't really make sense in the case of pre-tokenized input anyway. 

We'll need to fix it though, and I added an ignored test case especially for this.

n1t0:
@stefan-it This should indeed fix the issue you reported. It will require some changes in `transformers` though.

@thomwolf You can start to have a look if you want to, the API shouldn't change any more. We should handle the integration inside of `transformers` before merging this though, to make sure we avoid too much back-and-forth.

@Pierrci I'll need your review on the node typings/bindings :smile:

Pierrci:
Good for me, great parsing refactoring btw! 👍

PR #272, opened 2020-05-19
Add Serialization

### LINKED ISSUE #15: Tokenizer saving/loading
We need to provide a way to save and load tokenizers to/from files.
Things that need to be saved:
 - [ ] Each part (Normalizer, PreTokenizer, ..) and their options
 - [ ] Added tokens / special tokens
 - [ ] The model's vocabulary

We can approach this in multiple ways, but in the end, we would like to have a single self-contained file that represents a tokenizer. We will probably need to have some scripts to convert existing models to this new format.

n1t0:
Closes #15 

- [x] Python bindings for loading/saving
- [x] Node bindings for loading/saving
- [x] Update documentations and READMEs
- [x] Update CHANGELOGs


n1t0:
Ok we are almost ready to merge this! The last missing piece is the Node typings. I'm not sure what's the best way to handle this @Pierrci

AndriyMulyar:
This is a great addition. Any info on an ETA for the merge into master?

bindings/node/lib/bindings/tokenizer.js
```diff
@@ -1,6 +1,11 @@
 const native = require("./native");
 
+class Tokenizer extends native.tokenizer_Tokenizer {
```
Pierrci, line 3: @n1t0 I went for something very simple, I was not completely sure at first that it would work but it does 😁.
Let me know if it's good for you, I might add more JS tests early next week, but feel free to merge without them!

n1t0: That seems great! Thank you @Pierrci!

PR #276, opened 2020-05-22
Speedup train process by reading file in one pass and iterating in parallel over lines.

Narsil:
Looking at [YouTokenToMe](https://github.com/VKCOM/YouTokenToMe) it looks like 
tokenizers is still a bit behind in tokenizer training time (BPE) (3-4s vs ~20s on a 
100Mo dataset and it gets worse with size, 10 vs 120s on 500Mo). I'm starting to 
look into it to see what could be done. The first culprit seems to be file loading.
On a 100Mo file, loading everything into memory in one go takes ~30ms on my computer
whereas it takes ~13s to do the "Reading files" step.

There seem to be 3 culprits:
 - First, the read_line method uses a small buffer by
default (8kb) which make tons of syscalls.
 - Second, the parallel iteration occurs
on various files, and not within a single file (most datasets include large files
IMHO).
 - Third, within the line parsing function, lots of string copies are done.

This PR aims at reading the files in one chunk (or at least larger chunks) and 
iterating over lines in parallel instead. The result on my computer on a 100Mo
dataset is going from ~13s to ~7s on the `Reading files` step while maintining 
backward compatibility. Reducing copies is harder but could probably drop the 
time to ~1-2s (~1s is what youtokentome achieves to end up with the full
word count hashmap).

- It will fail on very large files (using read_to_string, *but* it's much faster than previous
iterations. (No bench yet.)
- It should be easy enough to support large files by chunking them at
space boundaries.
- Currently still pretty slow because lots of string copies happen at
train time (Normalizer + PreTokenizer). I also added one to include extra '\n' that
par_lines would remove. No default iterator seems to keep them so we will probably have
to do some custom logic if '\n' need to be kept. It seems Normalizer and PreTokenizer

(Lower bound is indeed 1s : https://gist.github.com/Narsil/833ef43fe4e6169cf54de57733a753b5)

### DIFF
--- a/tokenizers/Makefile
+++ b/tokenizers/Makefile
@@ -6,7 +6,7 @@ dir_guard=@mkdir -p $(@D)
 
 SHARED_RESOURCES = $(DATA_DIR)/gpt2-vocab.json $(DATA_DIR)/gpt2-merges.txt
 BENCHMARK_RESOURCES = $(SHARED_RESOURCES) $(DATA_DIR)/big.txt
-TESTS_RESOURCES = $(SHARED_RESOURCES) $(DATA_DIR)/bert-base-uncased-vocab.txt
+TESTS_RESOURCES = $(SHARED_RESOURCES) $(DATA_DIR)/bert-base-uncased-vocab.txt $(DATA_DIR)/small.txt
 
 .PHONY : build
 build :
@@ -56,3 +56,6 @@ $(DATA_DIR)/bert-% :
 $(DATA_DIR)/big.txt :
 	$(dir_guard)
 	wget https://norvig.com/big.txt -O $@
+
+$(DATA_DIR)/small.txt : $(DATA_DIR)/big.txt
+	head -100 $(DATA_DIR)/big.txt > $@
--- a/tokenizers/Makefile
+++ b/tokenizers/Makefile
@@ -5,8 +5,8 @@ TESTS_DIR = tests
 dir_guard=@mkdir -p $(@D)
 
 SHARED_RESOURCES = $(DATA_DIR)/gpt2-vocab.json $(DATA_DIR)/gpt2-merges.txt
-BENCHMARK_RESOURCES = $(SHARED_RESOURCES) $(DATA_DIR)/big.txt
-TESTS_RESOURCES = $(SHARED_RESOURCES) $(DATA_DIR)/bert-base-uncased-vocab.txt $(DATA_DIR)/small.txt
+BENCHMARK_RESOURCES = $(SHARED_RESOURCES) $(DATA_DIR)/big.txt $(DATA_DIR)/small.txt
+TESTS_RESOURCES = $(SHARED_RESOURCES) $(DATA_DIR)/bert-base-uncased-vocab.txt
 
 .PHONY : build
 build :
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -17,7 +17,6 @@ use rayon::prelude::*;
 use std::{
     collections::{HashMap, HashSet},
     fs::File,
-    io::{BufRead, BufReader},
     path::{Path, PathBuf},
 };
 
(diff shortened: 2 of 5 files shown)


n1t0:
Hi @Narsil, thank you very much for this. I've been wanting to process lines in parallel for ages but never took the time to do it!

We don't mean to be as fast as YouTokenToMe on the training, and we probably won't ever be. But that is fine! The main reason is that we would like to allow any kind of tokenization, and this kind of flexibility will necessarily mean trading for some speed. YouTokenToMe has been optimized for a specific case and as such, it will obviously perform better on this case.

Now, that being said, I'm all in for everything we can optimize without overcomplicating the API!

I think what takes most of the time is indeed the `Normalizer` and `PreTokenizer` steps. Reading the entire file and counting words is actually really fast (even with `read_line` without parallelization). I'm not sure we can really limit the allocations in the `Normalizer` and `PreTokenizer` since they generally modify the underlying strings. What do you think?

A few things about this PR:
  - I played with it, and it seems to start training around 25% of the `Reading files` steps. I don't know exactly what happens but I keep seeing the progress bar and the progress being made on reading files, while it starts doing everything else at the same time.
 - I'm a bit concerned about reading the files entirely. I feel like it is important to stream them to avoid taking too much memory, especially for users that couldn't use machines with huge amounts of RAM. What do you think?

Narsil:
> We don't mean to be as fast as YouTokenToMe on the training, and we probably won't ever be. But that is fine! The main reason is that we would like to allow any kind of tokenization, and this kind of flexibility will necessarily mean trading for some speed. YouTokenToMe has been optimized for a specific case and as such, it will obviously perform better on this case.

Yes, that makes sense which is why I use this and not youtokentome, however the sharp difference in speed led me to want to investigate a little. 

> I played with it, and it seems to start training around 25% of the Reading files steps. I don't know exactly what happens but I keep seeing the progress bar and the progress being made on reading files, while it starts doing everything else at the same time.

I use par_lines() and use a channel to send to another thread that keeps updating the progressbar. Maybe the channel buffer gets filled faster than it can be depleted ? So the thread updating that progressbar keeps updating it after it's actually finished ? I used that strategy to avoid sharing the progressbar within par_lines. Oddly enough `progressbar.inc` seems to not mutate itself, so probably better to share directly the progressbar.

> I'm a bit concerned about reading the files entirely. I feel like it is important to stream them to avoid taking too much memory, especially for users that couldn't use machines with huge amounts of RAM. What do you think?

For sure ! I was imagining reading 100Mo or so at a time, only reason it's not done here is because it's actually a bit tricky to get back the next newline character (so we don't cut inside a word). It gets the code a bit more complex.
Other thing I wanted to add was a test for this function, maybe on a smaller data file like `data/big.txt` like the first 100lines or so ? (Maybe adding an opt-out of the progress bar here too).

Another question. Is there any reason why there is no pre_tokenizer by default ? Without it, it changes the speed of this function by quite a bit because the `words` variable basically store the entire dataset in memory instead of doing word counting.

Narsil:
> For sure ! I was imagining reading 100Mo or so at a time, only reason it's not done here is because it's actually a bit tricky to get back the next newline character (so we don't cut inside a word). It gets the code a bit more complex.

Took me a while to figure out that `par_bridge` was what I was looking for.. Google led me into old rabbit holes :(

Narsil:
Improved the solution a bit by using a custom Line Iterator that does not remove trailing '\n'  and '\r' ( I copied lines iterator from BufRead trait and removed the pop lines).

I moved the test into benchmark, because it seems Windows does not download the datafiles for tests (it does not read the Makefile). I imagined a unit test would be good, but I admit depending on a file for a unit test is not perfect.

What do you think ? I could remove those benchmarks.

n1t0:
> Took me a while to figure out that par_bridge was what I was looking for.. Google led me into old rabbit holes :(

Wow! Very good finding, I didn't know about this even though I already looked for something similar multiple times... The resulting implementation is really good!

> I moved the test into benchmark, because it seems Windows does not download the datafiles for tests (it does not read the Makefile). I imagined a unit test would be good, but I admit depending on a file for a unit test is not perfect.

Yes that's perfect, I think it makes more sense as a benchmark. I didn't manage to make it work with windows on the CI, and didn't have a windows handy to debug it at the time. Ideally it should run on windows too indeed.

> I use par_lines() and use a channel to send to another thread that keeps updating the progressbar. Maybe the channel buffer gets filled faster than it can be depleted ?

I found this really surprising and so I ran a few tests. `indicatif` seems to be slowing everything down, and not just a little. See below, running with the last version on this branch:

**100MB while showing progress**
| Train            |   English |   Japanese |   Chinese |   Russian |
|--|--|--|--|--|
| YouTokenToMe     |   3.08772 |    7.71737 |   7.79187 |   2.89366 |
| tokenizers |  16.3958  |   30.0505  |  17.3656  |  17.523   |


**100MB no progress**
| Train            |   English |   Japanese |   Chinese |   Russian |
|--|--|--|--|--|
| YouTokenToMe     |   3.51883 |    8.28085 |   8.13546 |   2.94396 |
| tokenizers |  10.8329  |   15.8105  |   9.88912 |  11.7113  |

tokenizers/src/tokenizer/mod.rs
```diff
@@ -558,64 +558,88 @@ impl Tokenizer {
...
-                                prev_prog = curr_prog;
+                            let b = newline.len();
+                            if let Some(pbar) = progress {
+                                pbar.inc(b as u64);
+                            }
+                            Ok(words)
+                        },
+                    )
+                    .try_reduce(HashMap::new, |mut acc, ws| {
+                        {
```
n1t0, line 614: nit: I think we can remove these brackets.

tokenizers/src/tokenizer/mod.rs
```diff
@@ -876,3 +900,35 @@ impl Tokenizer {
         }
     }
 }
+
+#[derive(Debug)]
+pub struct Lines<B> {
```
n1t0, line 905: We can probably move this somewhere under `utils`. It might be interesting to be able to use it somewhere else.

tokenizers/src/tokenizer/mod.rs
```diff
@@ -876,3 +900,35 @@ impl Tokenizer {
...
+                //     }
+                // }
+                Some(Ok(buf))
+            }
+            Err(e) => Some(Err(e)),
+        }
+    }
+}
+
+fn lines_with_ending<B>(buf: B) -> Lines<B>
```
n1t0, line 929: To make this feel more "natural" (like the `.lines()`), we can maybe have a `trait` that requires a method `lines_with_ending` that does this, and we can implement it for `B: BufRead` so that we can then do `file.lines_with_ending().par_bridge()` on the `BufReader` above. What do you think?

tokenizers/Makefile
```diff
@@ -6,7 +6,7 @@ dir_guard=@mkdir -p $(@D)
 
 SHARED_RESOURCES = $(DATA_DIR)/gpt2-vocab.json $(DATA_DIR)/gpt2-merges.txt
 BENCHMARK_RESOURCES = $(SHARED_RESOURCES) $(DATA_DIR)/big.txt
-TESTS_RESOURCES = $(SHARED_RESOURCES) $(DATA_DIR)/bert-base-uncased-vocab.txt
+TESTS_RESOURCES = $(SHARED_RESOURCES) $(DATA_DIR)/bert-base-uncased-vocab.txt $(DATA_DIR)/small.txt
```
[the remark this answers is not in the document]
Narsil: Done

tokenizers/src/tokenizer/mod.rs
Narsil: For sure !

tokenizers/src/tokenizer/mod.rs
Narsil: Much better idea, I'm new to rust and I think you can't implement trait for external structs, But it didn't knwo you could implement a trait for other traits. That's neat.

Narsil:
> > Took me a while to figure out that par_bridge was what I was looking for.. Google led me into old rabbit holes :(
> 
> Wow! Very good finding, I didn't know about this even though I already looked for something similar multiple times... The resulting implementation is really good!

Careful though, apparently it can be significantly slower than native rayon primitives but very useful in some cases. IO and TCP read are cited as good use cases

`This creates a "bridge" from a sequential iterator to a parallel one, by distributing its items across the Rayon thread pool. This has the advantage of being able to parallelize just about anything, but the resulting ParallelIterator can be less efficient than if you started with par_iter instead. However, it can still be useful for iterators that are difficult to parallelize by other means, like channels or file or network I/O.`

The resulting iterator is not guaranteed to keep the order of the original iterator.`

> 
> > I moved the test into benchmark, because it seems Windows does not download the datafiles for tests (it does not read the Makefile). I imagined a unit test would be good, but I admit depending on a file for a unit test is not perfect.
> 
> Yes that's perfect, I think it makes more sense as a benchmark. I didn't manage to make it work with windows on the CI, and didn't have a windows handy to debug it at the time. Ideally it should run on windows too indeed.

Something like this could come in handy for Windows, no ? https://sagiegurari.github.io/cargo-make/

> 
> > I use par_lines() and use a channel to send to another thread that keeps updating the progressbar. Maybe the channel buffer gets filled faster than it can be depleted ?
> 
> I found this really surprising and so I ran a few tests. `indicatif` seems to be slowing everything down, and not just a little. See below, running with the last version on this branch:
> 
> **100MB while showing progress**
> Train 	English 	Japanese 	Chinese 	Russian
> YouTokenToMe 	3.08772 	7.71737 	7.79187 	2.89366
> tokenizers 	16.3958 	30.0505 	17.3656 	17.523
> 
> **100MB no progress**
> Train 	English 	Japanese 	Chinese 	Russian
> YouTokenToMe 	3.51883 	8.28085 	8.13546 	2.94396
> tokenizers 	10.8329 	15.8105 	9.88912 	11.7113


Hmm progressbar by default update every `inc` call, which is indeed pretty wasteful. I had a significant speedup by setting draw_delta(n / 100).
Can you confirm ?

tokenizers/src/tokenizer/mod.rs
n1t0: You can actually implement any trait on any type, as long as one of them is defined in your own crate. It’s not possible if both are from external crates. This is known as the orphan rule, and it prevents having multiple implementations, with rust not knowing which one to use.

So you can either:
 - implement your own trait on any type
 - implement an external trait on any of your types
 - redefine your own type to just wrap another type, so that you can implement a trait on it.

n1t0:
> Hmm progressbar by default update every inc call, which is indeed pretty wasteful. I had a significant speedup by setting draw_delta(n / 100).
Can you confirm ?

Indeed, in this case I am getting performances close to what we can see without showing any progress. That's really nice!

n1t0:
Alright, everything looks good! Thank you again @Narsil, and great job!

Issue #252, opened 2020-04-27
Training a BPE model in Rust


Ayuei:
I couldn't find any examples on how to do this. So I crunched up some code by looking at the test cases and type hints. I'm not sure if this is the correct way to do this however, would like some help.

(Build model, pretokenize text, count text/process, train the model with counts hashmap, save).
```rust 
// Build trainer with fields
let trainer = BpeTrainerBuilder::new()`
        .show_progress(true)
        .vocab_size(vocab_size)
        .min_frequency(0)
        .special_tokens(vec![AddedToken::from("<s>".into()),
        AddedToken::from("<pad>".into()),
        AddedToken::from("</s>".into()),
        AddedToken::from("<unk>".into()),
        AddedToken::from("<mask>".into())])
        .build();

let mut words = HashMap::new();
let b = ByteLevel::default();   // Pretokenizer

for (i, entry) in glob_files.enumerate(){
        pbar.inc(1);
        if let Ok(path) = entry {
            let file = File::open(path);
            let reader = BufReader::new(file.unwrap());

            for line in reader.lines() {
                let mut text = NormalizedString::from(&line.unwrap()[..]);
               
                // Clean text
                text.strip();
                text.nfc();

                // Pretokenize
                let byte_text = b.pre_tokenize(&mut text)
                    .unwrap()
                    .into_iter()
                    .map(|f| f.0)
                    .collect();

                // Get counts
                trainer.process_tokens(&mut words, byte_text);
            }
      }
}

let (model, _special_tokens) = trainer.train(words).unwrap();
model.save(folder, Some("my_tokenizer")).unwrap();
```

n1t0:
Hey @Ayuei. You can actually build a `Tokenizer` and train it like this:

```rust
use tokenizers::models::bpe::{BpeTrainerBuilder, BPE};
use tokenizers::normalizers::{strip::Strip, unicode::NFC, utils::Sequence};
use tokenizers::pre_tokenizers::byte_level::ByteLevel;
use tokenizers::tokenizer::{AddedToken, Tokenizer, Trainer};

let trainer: Box<dyn Trainer> = Box::new(
    BpeTrainerBuilder::new()
        .show_progress(true)
        .vocab_size(vocab_size)
        .min_frequency(0)
        .special_tokens(vec![
            AddedToken::from("<s>".into()),
            AddedToken::from("<pad>".into()),
            AddedToken::from("</s>".into()),
            AddedToken::from("<unk>".into()),
[...]

PR #277, opened 2020-05-24
127 rust example not working

### LINKED ISSUE #127: Rust example not working
where to get vocab and merges files?

```
$ cargo run
   Compiling tokenizer-test v0.1.0 (/home/user/tmp/tokenizer-test)
warning: variable does not need to be mutable
  --> src/main.rs:11:6
   |
11 |     let mut tokenizer = Tokenizer::new(Box::new(bpe));
   |         ----^^^^^^^^^
   |         |
   |         help: remove this `mut`
   |
   = note: `#[warn(unused_mut)]` on by default

    Finished dev [unoptimized + debuginfo] target(s) in 1.94s
     Running `target/debug/tokenizer-test`
Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

KappaDistributive:
Closes #127 

What I've done:

- provided sample `vocab.json`
- provided sample `merges.txt`
- provided expected output of code example
- added manditory second argument to `encode` method

Caveat: This will likely break again in the future. I'd be great if all the example in this repo would be tested automatically.

### DIFF
--- a/tokenizers/README.md
+++ b/tokenizers/README.md
@@ -35,57 +35,67 @@ The various steps of the pipeline are:
 
 ## Quick example
 
-Train and save a Byte Pair Encoder.
+Train and serialize a Tokenizer.
+
 
 ```Rust
-use std::collections::HashMap;
+use tokenizers::models::bpe::{BpeTrainerBuilder, BPE};
+use tokenizers::Result;
+use tokenizers::normalizers::{strip::Strip, unicode::NFC, utils::Sequence};
+use tokenizers::pre_tokenizers::byte_level::ByteLevel;
+use tokenizers::tokenizer::{AddedToken, Tokenizer, Trainer};
+
 use std::path::Path;
-use tokenizers::{Model, Result};
-use tokenizers::models::bpe::BpeTrainer;
 
-fn main() -> Result<()> {
-    let word_counts: HashMap<String, u32> = [
-        (String::from("[UNK]"), 1),
-        (String::from(" "), 1),
-        (String::from("Hey"), 1),
-        (String::from("there"), 1),
-    ].iter().cloned().collect();
+fn main() -> Result<()>{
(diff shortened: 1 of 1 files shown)


n1t0:
Thank you for this @KappaDistributive.

I'd like to avoid providing the content of the files as an example because they are not supposed to be manually edited. I think this leaves to much room for errors, and one shouldn't attempt to do it generally. Would you be willing to provide an example of how to train a tokenizer and obtain these files instead?

KappaDistributive:
Yeah, I agree and will adapt my PR accordingly. Thanks for your feedback!

KappaDistributive:
Is this (roughly) what you had in mind, @n1t0 ?

n1t0:
Thanks for the update! I was more thinking something like this: https://github.com/huggingface/tokenizers/issues/252#issuecomment-620168525

Having examples to show:
 1. How to train a `Tokenizer` directly
 2. Load a `Tokenizer`

In the example I sent you, it actually shows how to save the model (BPE) directly, providing the `merges.txt` and `vocab.json` files as in the example that you are updating. But there is also a new way that we could demonstrate too, have a look [at this](https://github.com/huggingface/tokenizers/blob/master/tokenizers/src/tokenizer/mod.rs#L343-L366)  (especially `save` and `from_file`, `to_string` is more a nice to have). What do you think?

KappaDistributive:
Sounds good.

There's one tangentially related question I'd like to clarify first, though: Why does the train method of Tokenizer expect a Vec of Strings rather than a Vec of Paths for its training data?

n1t0:
For no good reason indeed! I found `str` easier to use than `Path` at first, but it is limiting in many cases.
We could definitely update it to use something like for `from_file` with a `P: AsRef<Path>`.

n1t0:
Thank you @KappaDistributive! Really appreciate this!

PR #303, opened 2020-06-11
Node bindings - Huge refactoring

n1t0:
In this PR, we leverage all the extraction methods that we introduced with #249. This leads to code that is a lot clearer and easier to read. Updating the node bindings after this PR should be much more enjoyable than before.

We also take advantage of this PR to fix all the Clippy lint warnings that existed to date and enable this check in the CI.


Pierrci:
Ok so I survived and didn't drown! I fixed a couple of things, very nice refactoring!

Issue #187, opened 2020-03-04
python version hangs on encode_batch when run in subprocess


kretes:
Hello,

When I run encode_batch with at least two texts in an array in python subprocess - the call hangs.
This happens in real-life when used inside pytorch dataloaders with multiple workers.
A self-contained reproducible  script is here: https://gist.github.com/kretes/1a51bb8b936fc4e6277f71931b886bed

epwalsh:
Hey @kretes, thanks for bringing this up and thanks for the reproducible example. I ran the gist and sure enough I saw the same results.

What I'm guessing is happening (and keep in mind I am no expert in multiprocessing so I could be way off) is that when `tokenizers` is imported at the top level, the Rust code behind the scenes is creating some resources that it needs for parallel processing (like a thread pool and I don't know what else - this stuff comes from https://github.com/rayon-rs/rayon and I'm not super familiar with it). Then when the process is forked and `encode_batch` is called, the child process now needs to access those resources, but they are locked by the parent process.

If instead you import tokenizers within the `encode` function (so after the forking), you'll see this actually works:

```python
from multiprocessing import Process

def encode():
    from tokenizers.implementations import ByteLevelBPETokenizer
    tok = ByteLevelBPETokenizer()
    print(tok.encode_batch(['ala']))
    print(tok.encode_batch(['ala', 'kot']))

p = Process(target=encode)
p.start()
p.join()
```

ivanjacobsec:
Encode batch hangs for me in python even when not ran as a subprocess.

epwalsh:
@ivanjacobsec can you provide an example?

ivanjacobsec:
I tried 2 approaches:


1. 

 ```
    data_path="../data/"
    dest_path='../data/models/'
    model_name='my-tokenizer'
    paths = [str(x) for x in Path(data_path).glob("**/*.txt")]
    from tokenizers import Tokenizer, models, pre_tokenizers, decoders, processors

    tokenizer = Tokenizer(BPE.empty())

    tokenizer.pre_tokenizer = pre_tokenizers.PreTokenizer.custom(MyPretokenizer())
    tokenizer.decoder=decoders.Decoder.custom(MoleculePretokenizer())
    tokenizer.pre_tokenizer = pre_tokenizers.PreTokenizer.custom(MoleculePretokenizer())
    tokenizer.decoder = decoders.Decoder.custom(MoleculePretokenizer())

[...]

PR #306, opened 2020-06-15
add ability to disable parallelism

epwalsh:
Addresses #187. Python tokenizers will now work within a sub-process as long as you disable parallelism:

```python
import sys
from multiprocessing import Process
from tokenizers.implementations import ByteLevelBPETokenizer

tok = ByteLevelBPETokenizer()
tok.encode("hi")
tok.encode_batch(["hi", "there"])


def encode():
    # Disabling parallelism is key to avoiding a dead lock within a sub-process.
    tok.parallelism = False
    print(tok.encode("hi"))
    print(tok.encode_batch(["hi", "there"]))


p = Process(target=encode)
p.start()

p.join(timeout=1)
if p.is_alive():
    print("process is hanging, killing now")
    p.terminate()
    sys.exit(1)
```

---

My original idea of having a custom trait for parallel iterators that would wrap `rayon` turned out to be way too complicated.

So instead I just added a `bool` option called `parallelism` to the `Tokenizer` struct, and threw in a check for this everywhere that `into_par_iter()` (or anything else like that from `rayon`) is used. So when `parallelism` is `false`, the non-parallel version (such as `into_iter()`) is used.

### DIFF
--- a/bindings/node/native/src/encoding.rs
+++ b/bindings/node/native/src/encoding.rs
@@ -268,7 +268,7 @@ declare_types! {
                     params.pad_id,
                     params.pad_type_id,
                     &params.pad_token,
-                    params.direction
+                    params.direction,
                     true
                 );
             });
--- a/bindings/node/native/src/encoding.rs
+++ b/bindings/node/native/src/encoding.rs
@@ -355,7 +355,7 @@ declare_types! {
             let mut this = cx.this();
             let guard = cx.lock();
             this.borrow_mut(&guard).encoding.execute_mut(|encoding| {
-                encoding.unwrap().pad(length, pad_id, pad_type_id, &pad_token, direction);
+                encoding.unwrap().pad(length, pad_id, pad_type_id, &pad_token, direction, true);
             });
 
             Ok(cx.undefined().upcast())
--- a/bindings/python/CHANGELOG.md
+++ b/bindings/python/CHANGELOG.md
@@ -17,6 +17,8 @@ This adds some methods to easily save/load an entire tokenizer (`from_str`, `fro
 activation of the Tensor Cores, while ensuring padding to a multiple of 8. Use with
 `enable_padding(pad_to_multiple_of=8)` for example.
 - [#298]: Ability to get the currently set truncation/padding params
+- [#187]: Added the ability to disable parallelism in the Rust lib, which will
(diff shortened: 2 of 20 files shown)


epwalsh:
I don't know if this issue would arise in Node (and I don't know my way around Node anyway) so I haven't add anything to those bindings.

n1t0:
Thank you for the PR @epwalsh! I'm sorry I didn't get a chance to work on this for now, as I wanted to check a few things, but I'll do my best to get it done asap.

Can you elaborate a bit on the difficulties you encountered while trying to wrap `.par_iter` and others in our own traits? I really liked the idea and would love to explore this a bit more if at all possible. I fear that we end up having to think about wrapping this manually everywhere otherwise, and having to carry the `parallelism` property everywhere might make things difficult in some cases.

epwalsh:
No worries, take your time!

> Can you elaborate a bit on the difficulties you encountered while trying to wrap `.par_iter` and others in our own traits? I really liked the idea and would love to explore this a bit more if at all possible. I fear that we end up having to think about wrapping this manually everywhere otherwise, and having to carry the `parallelism` property everywhere might make things difficult in some cases.

The main complication I saw with trying to creating a custom `ParalleIterator` trait is that we'd have to implement it on every iterator struct that we use from both https://doc.rust-lang.org/std/iter/trait.Iterator.html and https://docs.rs/rayon/1.3.1/rayon/iter/trait.ParallelIterator.html. And then our custom trait methods would have to `Box` their output. This could get pretty ugly.

So since we only use the rayon trait method in a handfull of locations I figured - although not ideal - just manually wrapping would be much more tractable.

The other thing I realized is that the dead lock issue is not just with `rayon`. In fact, any attempt to access a `lazy_static` var after forking results in a dead lock. And so it seems like passing `parallelism` around is necessary, as opposed to having a global `STATIC` flag to disable parallelism that gets it's value from an environment variable.

(I also tried using a `STATIC` flag with `once_cell`, but saw the same issue)

n1t0:
Indeed, I saw problems with `lazy_static` too, when digging #183. There was a deadlock when using the byte-level `PreTokenizer` even with `encode` since it uses it to generate the regex and the byte<->characters maps.

I'm surprised that it seems to work here, even though you added a test for byte-level BPE

epwalsh:
Closing in favor of #311

PR #311, opened 2020-06-19
Make parallelism optional

n1t0:
Related to #187 

An attempt at implementing your first idea @epwalsh. I just didn't manage to have `map_with` working with `maybe_par_bridge` but everything else seems to be good!

Let me know what you think! And if you have an idea for the `map_with`, that'd be awesome, I couldn't make sense of the error...

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -104,7 +104,7 @@ fn normalizers(_py: Python, m: &PyModule) -> PyResult<()> {
 #[pymodule]
 fn tokenizers(_py: Python, m: &PyModule) -> PyResult<()> {
     // Register the fork callback
-    #[cfg(target_os = "linux")]
+    #[cfg(target_family = "unix")]
     unsafe {
         if !REGISTERED_FORK_CALLBACK {
             libc::pthread_atfork(None, None, Some(child_after_fork));
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -1,3 +1,5 @@
+extern crate tokenizers as tk;
+
 mod decoders;
 mod encoding;
 mod error;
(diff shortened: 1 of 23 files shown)


epwalsh:
Oof, that error message has me stumped.

I do like this solution though! I hope we can get it working.

tokenizers/src/utils/parallelism.rs
```diff
@@ -0,0 +1,179 @@
...
+pub const ENV_VARIABLE: &str = "TOKENIZERS_PARALLELISM";
+
+/// Check if the TOKENIZERS_PARALLELISM env variable has been explicitly set
+pub fn is_parallelism_configured() -> bool {
+    std::env::var(ENV_VARIABLE).is_ok()
+}
+
+/// Get the currently set value for `TOKENIZERS_PARALLELISM` env variable
+pub fn get_parallelism() -> bool {
+    match std::env::var(ENV_VARIABLE) {
```
epwalsh, line 18: Is the cost of looking up this environment variable every time significant?

bindings/python/src/lib.rs
```diff
@@ -84,6 +103,15 @@ fn normalizers(_py: Python, m: &PyModule) -> PyResult<()> {
 /// Tokenizers Module
 #[pymodule]
 fn tokenizers(_py: Python, m: &PyModule) -> PyResult<()> {
+    // Register the fork callback
+    #[cfg(target_os = "linux")]
+    unsafe {
+        if !REGISTERED_FORK_CALLBACK {
+            libc::pthread_atfork(None, None, Some(child_after_fork));
```
epwalsh, line 110: 💯 💯 🎉 🎉 !!!

epwalsh: TIL

n1t0: I didn't know about it either, and it seems to be working really well. This should let us have a very flawless experience. :tada:

n1t0:
Ok, everything should be good now. `map_with` now also works when using `maybe_par_bridge`.

So, the behavior in Python is now the following:
 - When the library is imported for the first time, we register a listener to make sure we catch a potential `fork`
 - If we detect a `fork` and `TOKENIZERS_PARALLELISM` is not set, we disable the parallelism and warn the user
 - If `TOKENIZERS_PARALLELISM` has been set, we use the provided value

Except for the `fork` part, everything works the same on any OS/platform. The environement variable can be used to disable parallelism in any situation.

I think we might wait for https://github.com/cuviper/rayon-cond/pull/2 to be merged before we merge this, but this is now ready for review!

n1t0:
Nice catch! I think it should work fine now.

epwalsh:
I just confirmed it works! 💯

n1t0:
Ok, let's merge this to include it in the upcoming release. We'll see what to do with the `git` dependency when we'll release the Rust version.

PR #327, opened 2020-07-05
Use supplied UNK token even when vocab absent

aalok-sathe:
If a vocab file isn't provided the supplied unk token (different from [UNK]) gets ignored and later throws an error:
Exception: WordPiece error: Missing [UNK] token from the vocabulary
when trying to encode an input string with an unknown token.


n1t0:
Thanks @aalok-sathe

PR #334, opened 2020-07-08
Fix Rust example.

junnlikestea:
Hi, the Rust example wasn't working for me so I updated it.

### DIFF
--- a/tokenizers/README.md
+++ b/tokenizers/README.md
@@ -84,14 +84,20 @@ fn main() -> Result<()> {
 Deserialize a pretrained Tokenizer.
 
 ```Rust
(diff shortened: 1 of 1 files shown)


n1t0:
Hi @junnlikestea. The example and the doc in the repository should be up-to-date with `master`. We didn't make a rust release recently so if you didn't run from `master`, it surely is different. At the moment we have a dependency that is not on crates.io and prevents us from releasing, but you can absolutely test it by using:
```toml
tokenizers = { version = "*", git = "https://github.com/huggingface/tokenizers" }
```

junnlikestea:
Ok no worries, closing the PR.

PR #344, opened 2020-07-16
Clean up various normalization methods.

sebpuetz:
Rust 1.45 added a new clippy lint for empty ranges, that led to failing CI builds which are now addressed by the first commit in this PR.

**edit:** Looks like are some additional new lints in the node bindings...

----------
Clean up various normalization methods.

* `char` is `Copy`, therefore filter can take `char` by value
* `Char` is double ended, no need to collect before `rev()`
* clearer name for filter condition
* reverse iterator instead of reversing a vec in-place
* replace `filter(|o| o.is_some()).map(|o| o.unwrap())` with single
  `filter_map`
* simplify logic in `transform`
* directly collect transformation into `String`, skipping the
  `Vec<char>` first

--------------

Simplify some iterators.

* replace `.filter(|x| x.is_some()).map(|x| x.unwrap())` with
  `filter_map` calls
* replace some reducers with `fold` instead of mapping, filtering
  and summing

---------------

Make BERT pre tokenizer closure take chars by value.

char is `Copy`, can be passed by value.

### DIFF
--- a/tokenizers/src/normalizers/bert.rs
+++ b/tokenizers/src/normalizers/bert.rs
@@ -90,7 +90,7 @@ impl BertNormalizer {
 
     fn do_clean_text(&self, normalized: &mut NormalizedString) {
         normalized
-            .filter(|c| !(*c as usize == 0 || *c as usize == 0xfffd || is_control(*c)))
+            .filter(|c| !(c as usize == 0 || c as usize == 0xfffd || is_control(c)))
             .map(|c| if is_whitespace(c) { ' ' } else { c });
     }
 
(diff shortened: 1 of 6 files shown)


sebpuetz:
I rebased this PR on master and removed commit to address the new clippy lints.

n1t0:
Love these! It's way better indeed. 

I don't remember exactly why it didn't work initially in `filter` but I had a very weird behavior where instead of getting the expected result `vec![(a1,  b1),  (a2, b2), ...]`, we had something like `vec![(a1, b2), (a2,  b3), ...]`. It didn't make sense at all. Anyway, it's so much better like this, and I think we should now have enough tests to be confident about the result.

sebpuetz:
I think I can shed some light on why the double `rev`ing didn't work:

`rev` reverses the `Chars` iter - at that point, the iterator moves from the back through the byte-slice backing the `&str` and returns the `char`s as it encounters them. Calling `rev` again after `map`ping does not trigger a new iteration in the other direction after finishing iteration, Rust iterators are evaluated lazily, so effectively the double `rev`ing cancels out.

[Playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3905c05f7e923eb96c5d39284f5ab13f)
~~~Rust
"test".chars().rev().map(|c| {
    println!("{}", c);
    c
    })
    .rev()
    .collect::<String>();
~~~

prints `"t" "e" "s" "t"`

Issue #352, opened 2020-07-21
How to apply both pre-tokenizer and post-processor


joverwey:
the output of    

```
pub fn with_pre_tokenizer(&mut self, pre_tokenizer: Box<dyn PreTokenizer>) -> &Self
```
is *&Self* which is no longer mutable.

Shouldn't this be *Self* so that you can chain multiple mutating operations like

```
Tokenizer::new(Box::new(bpe))
                .with_post_processor(post_processor)
                .with_pre_tokenizer(pre_tokenizer)

```

n1t0:
You are totally right! It should be returning a `&mut Self` so that we can chain them. Feel free to make a PR if you want to!

joverwey:
I don't think you can return &mut Self because you would end up with multiple borrows.
The builder pattern usually looks more like:

```
fn with_option(mut self, value: usize) -> Self
```
 
I'll take a stab at it and send a pull request.

n1t0:
Yes, you can return `&mut Self`, and I think that's what we want. We don't want to consume `self` otherwise it becomes a pain when we don't chain.

PR #356, opened 2020-07-25
Added ability to chain setup calls for tokenizer

joverwey:
This is a stab at a solution to #352


n1t0:
Thank you for your PR @joverwey. I updated everything with https://github.com/huggingface/tokenizers/pull/407, adding what you suggested slightly differently: There is now a `TokenizerBuilder` that does exactly what you wanted (taking `mut self` with ownership), while the `Tokenizer` just returns a `&mut Self`.

PR #357, opened 2020-07-25
Move Python source to subdirectory.

sebpuetz:
This allows testing versions not built in-place. Otherwise
importing (or testing) in the package root fails without develop
builds.
Replace maturin with setuptools_rust since maturin fails with
proper project structure.

-----------------

Otherwise the package does not get tested under the circumstances it is installed as. I.e. running `pytest` under `tokenizers/bindings/python/` will import `tokenizers` from `tokenizers/bindings/python/tokenizers`. It's not possible to run the tests outside of `develop` mode without an in-place build since `import tokenizers` fails to find `tokenizers.tokenizers`.


n1t0:
Looks good thanks! I didn't know we could name it differently than `src`. It fixes all the issues with setuptools_rust too so that's really nice.

PR #362, opened 2020-08-03
Fix deadlocks with custom python components.

### LINKED ISSUE #359: Custom chinese pre-tokenizer
I am trying to use a custom pre-tokenizer based on a [jieba](https://github.com/fxsjy/jieba) library. It is a tool that allows splitting strings into meaningful words. 
Here is the code that I wrote in order to combine jieba tokens with tokenizers.  

```python 
import jieba

class JiebaPreTokenizer:
    """This class is a jieba adapter that mimic the interface of PreTokenizer, 
    which is a component responsible for a initial spliting a sentence into tokens.
    """
    def __init__(self, jieba_init_fn=None):
        """
        :param jieba_init_fn: is a function pointner that is executed in order to initialize jieba tokenizer engine, 
        see https://github.com/fxsjy/jieba#initialization
        """
        if jieba_init_fn is not None:
            jieba_init_fn()
    
    def pre_tokenize(self, sentence):
        """this function return tokenized version of the sentence, 
        it has to adjust the format of the output generated by jieba to be consistent with the one required by the tokenizers """
        return [(token[0],(token[1],token[2])) for token in jieba.tokenize(sentence)]
    
    def decode(self, tokens):
        return "".join(tokens)
    
    
def build_tokenizer():
        
    tokenizer = Tokenizer(BPE())
    jieba_pre_tokenizer = JiebaPreTokenizer(jieba.disable_parallel)
    tokenizer.pre_tokenizer = pre_tokenizers.PreTokenizer.custom(jieba_pre_tokenizer)
    tokenizer.decoder = decoders.Decoder.custom(jieba_pre_tokenizer)
        
    return tokenizer

def create_trainer():
    return BpeTrainer(vocab_size=15000, show_progress=True,
                    min_frequency=0, special_tokens=[
                    "<s>",
                    "<pad>",
                    "</s>",
                    "<unk>",
                    "<mask>",
                    ]
          )
               
        
if __name__ =="__main__":
    tokenizer = build_tokenizer()
    trainer = create_trainer()
    tokenizer.train(trainer, ["./descriptions_cn.txt"])
    tokenizer.model.save('./tokenizer_cn')
```

When running the code, I encounter the deadlock issue. Is there any way to make it work?

sebpuetz:
This should solve the training deadlock with custom components. (This is also compatible with the changes in #355)

Fixes #359

### DIFF
--- a/bindings/python/CHANGELOG.md
+++ b/bindings/python/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - [#333]: Fix deserialization of `AddedToken`, where the content was not restored properly
+- [#362]: Fix training deadlock with Python components.
 
 ### Changed
 - [#329]: Improved warning and behavior when we detect a fork
--- a/bindings/python/src/tokenizer.rs
+++ b/bindings/python/src/tokenizer.rs
@@ -583,12 +583,15 @@ impl Tokenizer {
                 Ok(input)
             })
             .collect::<PyResult<Vec<tk::EncodeInput>>>()?;
-        ToPyResult(
-            self.tokenizer
-                .encode_batch(input, add_special_tokens)
-                .map(|encodings| encodings.into_iter().map(|e| e.into()).collect()),
-        )
-        .into()
+        let gil = Python::acquire_gil();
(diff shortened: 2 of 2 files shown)


n1t0:
Nice! Thank you @sebpuetz

PR #363, opened 2020-08-03
Fix training panic.

### LINKED ISSUE #322: tokenizer.train encounter core dump when the file does not exist.
The tokenizer.train encounter core dump when the file does not exist. I hope it could be more elegant to show the error.

sebpuetz:
Fixes #322


n1t0:
Great! Thanks

PR #364, opened 2020-08-03
Cache based on Cargo.lock.

sebpuetz:
The lock-file should be more accurate in caching the correct dependencies than `Cargo.toml`.

I think the failing Rust MacOS jobs are due to incorrect caching. As long as `Cargo.toml` isn't changed, the dependencies stay the same for every build since the same cache hits are returned. While I'm not sure why the builds started to fail, I think they kept failing because the same (faulty) dependencies kept being retrieved. Removing the cache for the library should resolve such issues in the future. 

The previously suggestion to use `Cargo.lock`  as a key for the Rust library would only make sense if `cargo update` is executed as the first step to ensure `Cargo.lock` exists with up-to-date dependencies.

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -31,7 +31,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ~/.cargo/registry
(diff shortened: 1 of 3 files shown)


n1t0:
Make sense! Thanks @sebpuetz

PR #366, opened 2020-08-04
Add benches for WordPiece Tokenizers.

sebpuetz:
This adds a benchmark for WordPiece tokenizers and factors out some common methods between the two benches.


n1t0:
Thank you @sebpuetz. Good addition!

PR #368, opened 2020-08-05
Rust - NormalizedString - Fix offsets when adding new chars

### LINKED ISSUE #367: Unicode problem, Exception: Split pre-tokenized string must represent entire orignal String
Hi there,

I got an issue when tokenizing the following sentence:

Rika Noguchi ( 野口里佳 Noguchi Rika, born 1971 in Tokyo, Japan) is a Japanese photographic artist

Exception: Split pre-tokenized string must represent the entire original string. 

The code is shown as follows,
#####################################################
from tokenizers import (ByteLevelBPETokenizer,
                            CharBPETokenizer,
                            SentencePieceBPETokenizer,
                            BertWordPieceTokenizer)
tokenizer = BertWordPieceTokenizer("bert-base-uncased-vocab.txt", lowercase=True)
original_str = "Rika Noguchi ( 野口里佳 Noguchi Rika, born 1971 in Tokyo, Japan) is a Japanese photographic artist "
output = tokenizer.encode(original_str)

I think the problem comes from the chinese unicode 野口里佳, as the tokenizer works fine if these words are removed.

Is there any way to deal with it? 

Thank you. Sorry if the question should be post somewhere else. 

Best

n1t0:
New characters should have alignments of length 0 since they don't correspond to anything from the original string. This led to a bug where slicing the normalized string gave multiple times the same part of the original string.

Fix #367

### DIFF
--- a/tokenizers/src/normalizers/bert.rs
+++ b/tokenizers/src/normalizers/bert.rs
@@ -136,3 +136,22 @@ impl Normalizer for BertNormalizer {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn handle_chinese_chars() {
+        let mut normalized = NormalizedString::from("野口里佳 Noguchi Rika");
+        let normalizer = BertNormalizer {
+            handle_chinese_chars: true,
+            lowercase: false,
+            clean_text: false,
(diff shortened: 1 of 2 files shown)


n1t0:
Closing in favor of #377

Issue #10, opened 2019-12-27
Rust documentation


n1t0:
The main README is completely out-of-date.

We also want to provide documentation in order to prepare for the crate release. This covers the Rust documentation only, not bindings.

ropottnik:
However, it's a great exercise for people who want to understand the code: Trying to get the example from the README.md to work as I'm currently doing 😛 

I could do a PR with an updated Readme example once I'm done.

n1t0:
Another thing to consider is that the README is now completely out of sync with the last version on crates.io. We didn't release a Rust version for quite some time as we have unpublished dependencies. So you may want to pull `master`

ropottnik:
Yeah, I noticed that when first using the `crates.io` version and the `master` code for reference. Quite a change.

I could also try putting the example into a testable API documentation string and have the Readme just refer to it, such that there are as few things as possible in the Readme that are likely to quickly become obsolete again.

n1t0:
Sure, being able to test what's in the README would help a lot!
Everything that is related to the Rust documentation can fit in this issue, but feel free to open a new one if you prefer.

ropottnik:
@n1t0 could you have a look at #373 if it is fine like this?

strategy155:
Hello guys, how is the status for that issue? The README seems alright to me...

ArthurZucker:
Let's close it 😉

PR #373, opened 2020-08-15
testable example docs for training-serialization

ropottnik:
#10 

Updated example from the `Readme.md` to the newest API and converted it into testable docs inside `lib.rs` (actually with _```no_run_ to test if compilation works).

The example for deserialization and encoding was removed from the `Readme.md` since the `lib.rs` already contains a documentation test for that.
 
The `Readme.md` file now contains only references to these testable docs which should prevent them ever to be outdated ever again.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -25,7 +25,7 @@ jobs:
           architecture: "x64"
 
       - name: Install dependencies
-        run: pip install black
+        run: pip install black==19.10b0
 
       - name: Check style
         working-directory: ./bindings/python
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -29,6 +29,13 @@ jobs:
       - if: matrix.os == 'ubuntu-latest'
         run: sudo chown -R $(whoami):$(id -ng) ~/.cargo/
 
+      - name: Install cargo-readme for Ubuntu
+        if: matrix.os == 'ubuntu-latest'
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: cargo-readme
+
       - name: Build
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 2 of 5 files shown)


Narsil:
Food for thought [cargo-readme](https://github.com/livioribeiro/cargo-readme)

n1t0:
Thank you @ropottnik!

I agree with @Narsil that it is important to keep the examples in the `README.md` as well if possible. The main reason is that the version on `doc.rs` is not necessarily in sync with `master`.

`cargo-readme` seems nice to do this, and we could include it in the CI probably. I guess with the template file we can even keep the banner with the logo.

ropottnik:
Makes sense! So would the pipeline then generate the readme on every PR and make an additional commit?

n1t0:
I think I would probably instead make the pipeline generate it, and check that it is similar to what's in the tested commit, failing if not. Just like we do with linters and tests for example.

Do you want to try updating everything with a template and `cargo-readme`? No need to have the CI pipeline ready, (unless you really want to do it), we can add it later.

ropottnik:
@n1t0 sure, I'll give it a try.

ropottnik:
@n1t0 I added the `README.tpl` to generate the `README.md` as @Narsil suggested. I also added the pipeline step to check equality. a60fb5a and d83c177 show that the pipeline fails on a minor difference and then passes once that diff is taken care of. 

however, it seems as if some pipeline steps regarding `python` fail. this can't be in connection with my PR since no `python` files were touched, right? or am I missing something?

Narsil:
No you are right it seems the failing test are not linked to you.

Did you rebase ? It seems the problem lies in what you rebase against but master seems ok so I'm not sure.

ropottnik:
Pretty sure I rebased from `master`.

Could you try rerunning the github pipeline on `master` to be absolutely sure that there is a discrepancy in pipeline results? 
The last commit to `master` was 7 days ago and `black` might have changed something since then.

Narsil:
Yes it seems black got updated.

You can run
```
pip install -U black
cd bindings/python/
make style
git add ....
```

And see if that fixes it.

ropottnik:
I thought fixing the `black` version to what it was in the last master build (`19.10b0`) might be the cleaner option. What do you guys think @Narsil @n1t0 ?

n1t0:
Great thank you @ropottnik!

PR #378, opened 2020-08-21
Adding a 3 new PreTokenizers:

Narsil:
- Punctuation: Splits punctuation characters as isolated tokens
- Sequence: Applies a list of pretokenizers iteratively


n1t0:
Nice!

PR #383, opened 2020-08-31
Fix `Tokenizer::from_file`

n1t0:


n1t0:
Had to re-run all jobs because of a temporary issue with the CI, and it appears that the actions are not run on specific commits, but the entire branch. So the failing tests for `Add failing test for from_file` now appear to be green, but they ran on `Fix from_file`.
Expected result was: https://github.com/huggingface/tokenizers/pull/383/checks?check_run_id=1053181239#step:6:516
```
thread 'tokenizer::test::serialize' panicked at 'called `Result::unwrap()` on an `Err` value: Error("data did not match any variant of untagged enum PyNormalizerWrapper", line: 1, column: 153)', src/tokenizer.rs:864:36
```

I don't know why this actually fails but apparently `from_reader` and `from_str/from_slice` behave differently.

sebpuetz:
I think I have an explanation why this happens. `from_file` owns the Strings while deserializing, `from_str` can borrow from the in-memory String. The macro to implement the deserialization for unit structs assumes that it can extract a tuple of `&str`s for the `type: TypeName` entry. I'm looking into this, I'll let you know if it's really the issue.

sebpuetz:
Yup, changing the macro code in `tokenizers::utils::mod::impl_serde_unit_struct` fixes the bug:

```
        struct $visitor;
        impl<'de> serde::de::Visitor<'de> for $visitor {
            type Value = $self_ty;
            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
                write!(formatter, stringify!($self_ty))
            }

            fn visit_map<A>(self, mut map: A) -> std::result::Result<Self::Value, A::Error> where
                A: serde::de::MapAccess<'de>, {
                let self_ty_str = stringify!($self_ty);
                println!("visito {}", self_ty_str);
                let k = map.next_entry::<String, String>();
                let (k, v) = match k {
                    Ok(Some((k, v))) => (k, v),
                    Ok(None) => return Err(serde::de::Error::custom(&format!("Expected type : {}", self_ty_str))),
                    Err(e) => return Err(e),
                };
                match (k.as_str(), v.as_str()) {
                    ("type", stringify!($self_ty)) => Ok($self_ty),
                    (_, ty) => Err(serde::de::Error::custom(&format!("Expected {}, got {}", self_ty_str, ty))),
                }
            }
        }
```

That's just something I hacked together to verify that this causes the issue. I can make this a proper PR tomorrow or you can just address this yourself!

PR #384, opened 2020-09-01
Node bindings for the new pre_tokenizers.

Narsil:
Sequence
Punctuation

### DIFF
--- a/bindings/node/native/src/normalizers.rs
+++ b/bindings/node/native/src/normalizers.rs
@@ -241,8 +241,7 @@ mod test {
 
         let js_seq: JsNormalizerWrapper = Sequence::new(vec![NFC.into(), NFKC.into()]).into();
         let js_wrapper_ser = serde_json::to_string(&js_seq).unwrap();
-        let rs_wrapped =
-            NormalizerWrapper::Sequence(Sequence::new(vec![NFC.into(), NFKC.into()]).into());
+        let rs_wrapped = NormalizerWrapper::Sequence(Sequence::new(vec![NFC.into(), NFKC.into()]));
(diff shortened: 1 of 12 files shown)


n1t0:
Nice! You might want to add a `JsPreTokenizerWrapper` just like we have for the normalizers though. The reason we added it is because it lets you clone the `Arc` directly instead of the content. We need it if we want the users to be able to modify the normalizers as they might expect:
```javascript
let normalizer = BertNormalizer({ clean_text: true });
tokenizer.normalizer = normalizer;
normalizer.clean_text = false;
```

bindings/node/native/src/normalizers.rs
```diff
@@ -213,3 +213,46 @@ pub fn register(m: &mut ModuleContext, prefix: &str) -> NeonResult<()> {
...
+    #[test]
+    fn serialize() {
+        let js_wrapped: JsNormalizerWrapper = NFKC.into();
+        let js_ser = serde_json::to_string(&js_wrapped).unwrap();
+
+        let rs_wrapped = NormalizerWrapper::NFKC(NFKC);
+        let rs_ser = serde_json::to_string(&rs_wrapped).unwrap();
+        assert_eq!(js_ser, rs_ser);
+
+        // let js_norm: Normalizer = serde_json::from_str(&rs_ser).unwrap();
```
n1t0, line 233: Any reason for keeping these?

Narsil: Hmm I meant to ask you. 

I borrowed these from the python bindings, however deserialization here failed (js_norm.normalizer is None).

I am not sure why though.

PR #385, opened 2020-09-01
Fix deserialization from readers for unit structs

sebpuetz:
Deserialization from readers works with owned strings, the custom
deserialize implementations for unit structs assumed that they can
extract the type entry as &str. This commit changes the behaviour
to extracting Strings.

--------

Fixes #383. If you prefer to write your own solution, that's fine too! @n1t0


n1t0:
Perfect, thank you @sebpuetz!

PR #395, opened 2020-09-04
[WIP] Adding a new PreTokenizer for UnicodeScripts.

Narsil:
The core idea is to detect the "Script" of every char in a String
and split whenever there is a change of Script.

For instance

"どこで生れ。Yes" -> "どこで生れ", "。", "Yes"

The first chars are all considered `Han`, "。" is `Common` and the rest are
`Latin`.
This PreTokenizer will be helpful to disentangle pre-processing done in
`spm` and be able to reuse it better when needed.


Narsil:
We're currently missing a whitelist for some characters to be treated as Any and be allowed in any token (useful for spaces and continuation characters)

Narsil:
Waiting on this PR to stabilize Training mode. It should come along with it, not before IMO.

Narsil:
Fixed by #401

PR #400, opened 2020-09-08
Making cache layer common to BPE and Unigram.

Narsil:
And add a cache layer to Unigram.

It will probably have to be removed (or bypassed) when sampling is enabled.


n1t0:
LGTM!

PR #403, opened 2020-09-09
Template processor

### LINKED ISSUE #79: Add Template post-processor

n1t0:
Fix #79

### DIFF
--- a/bindings/python/tests/bindings/test_processors.py
+++ b/bindings/python/tests/bindings/test_processors.py
@@ -5,7 +5,13 @@ from ..utils import data_dir, roberta_files
 from tokenizers import Tokenizer
 from tokenizers.models import BPE
 from tokenizers.pre_tokenizers import ByteLevel as ByteLevelPreTokenizer
-from tokenizers.processors import PostProcessor, BertProcessing, RobertaProcessing, ByteLevel
+from tokenizers.processors import (
+    PostProcessor,
+    BertProcessing,
+    RobertaProcessing,
+    ByteLevel,
+    TemplateProcessing,
+)
 
 
 class TestBertProcessing:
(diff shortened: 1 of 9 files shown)


tokenizers/src/processors/template.rs
```diff
@@ -0,0 +1,704 @@
...
+
+    fn validate(&self) -> std::result::Result<(), String> {
+        let check = |sp| {
+            let exist = self
+                .special_tokens
+                .as_ref()
+                .map_or(false, |map| map.0.contains_key(sp));
+            match exist {
+                true => Ok(()),
+                false => Err(format!("The special token `{}` is missing", sp)),
```
Narsil, line 342: Could we delay the error to gather all missing tokens ?

As a user I found that having the change tokens one by one was tedious. (Maybe only display the first 10 max to prevent flooding though if case can happen.)

bindings/python/tests/bindings/test_processors.py
```diff
@@ -73,3 +79,78 @@ def test_processing(self, roberta_files):
...
+        tokenizer = Tokenizer(BPE())
+        tokenizer.add_special_tokens(["[SEP]", "[CLS]"])
+        tokenizer.add_tokens(["my", "name", "is", "john", "pair"])
+        tokenizer.post_processor = BertProcessing(("[SEP]", 0), ("[CLS]", 1))
+
+        original = tokenizer.encode("my name", "pair")
+
+        tokenizer.post_processor = self.get_bert()
+        template = tokenizer.encode("my name", "pair")
+        assert original.ids == template.ids
```
Narsil, line 145: I am a bit confused about ("[CLS]", 0) and the `.add_special_tokens`.

I tried to create a similar tokenizer, but it seems doing `tokenizer.add_special_tokens(["[SEP]", "[CLS]"])` add "[SEP]" with id=0. However I defined my PostProcessor with `special_tokens=[("[CLS]", 0), ("[SEP]", 1)]`. I will end up with [0, 2, 3, 1, 6, 1]. I think it should be incorrect, or something or something should yell at me. However I'm not sure how all this is supposed to work. Maybe the two "[SEP]" are different ?

Narsil (commented):
I am unsure about the use case of this PR, so take my comments with a grain of salt.
 
There are a few NITs and I'm not sure I understand something about the behavior.

tokenizers/src/processors/template.rs
```diff
@@ -0,0 +1,704 @@
...
+}
+
+impl From<String> for Piece {
+    fn from(s: String) -> Self {
+        // Try to extract `$(n)?` first
+        if s.starts_with('$') {
+            let rest = &s['$'.len_utf8()..];
+
+            // If the id is just `$`, we use 0 as type_id
+            if rest == "" {
```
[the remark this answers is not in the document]
n1t0: This means that when you specify `$`, it is equivalent to `$0`. It is just a shortcut for when you don't care about the `type_id`.

tokenizers/src/processors/template.rs
```diff
@@ -0,0 +1,704 @@
...
+                Piece::Sequence { type_id } => {
+                    default_type_id = *type_id;
+                    encoding.len()
+                }
+                Piece::SpecialToken { id } => {
+                    if add_special_tokens {
+                        self.special_tokens
+                            .0
+                            .get(id)
+                            .ok_or(format!("Missing SpecialToken with id {}", id))?
```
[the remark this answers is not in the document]
n1t0: It does trigger an error, but you are right, we should use `ok_or_else`. I don't know why Clippy didn't catch it this time.

bindings/python/tests/bindings/test_processors.py
n1t0: :+1: Opened #406 to track this.

PR #416, opened 2020-09-16
Adding a new normalizer that strips accents by removing combining

### LINKED ISSUE #413: Add a Strip accents Normalizer
Necessary at least for Unigram Albert tokenizer.

Current python code in transformers:
```python
outputs = unicodedata.normalize("NFKD", outputs)
outputs = "".join([c for c in outputs if not unicodedata.combining(c)])
```

Narsil:
characters in unicode strings.

Fixes #413

### DIFF
--- a/tokenizers/src/normalizers/strip.rs
+++ b/tokenizers/src/normalizers/strip.rs
@@ -38,9 +38,9 @@ impl Normalizer for Strip {
     }
 }
 
-/// This normalizer removes combining marks from a normalized string
-/// It's different from unidecode as it does not attempt to modify
-/// non ascii languages.
+// This normalizer removes combining marks from a normalized string
+// It's different from unidecode as it does not attempt to modify
+// non ascii languages.
 #[derive(Copy, Clone, Debug)]
 pub struct StripAccents;
 impl_serde_unit_struct!(StripAccentsVisitor, StripAccents);
--- a/tokenizers/src/normalizers/strip.rs
+++ b/tokenizers/src/normalizers/strip.rs
@@ -48,17 +48,7 @@ impl_serde_unit_struct!(StripAccentsVisitor, StripAccents);
 impl Normalizer for StripAccents {
     /// Strip the normalized string inplace
     fn normalize(&self, normalized: &mut NormalizedString) -> Result<()> {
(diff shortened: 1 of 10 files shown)


tokenizers/src/normalizers/strip.rs
```diff
@@ -36,3 +37,91 @@ impl Normalizer for Strip {
         Ok(())
     }
 }
+
+/// This normalizer removes combining marks from a normalized string
+/// It's different from unidecode as it does not attempt to modify
```
n1t0, line 42: nit: I don't know what is `unidecode` in this context, and it's probably irrelevant to have this in the doc.

tokenizers/src/normalizers/strip.rs
```diff
@@ -36,3 +37,91 @@ impl Normalizer for Strip {
...
+/// non ascii languages.
+#[derive(Copy, Clone, Debug)]
+pub struct StripAccents;
+impl_serde_unit_struct!(StripAccentsVisitor, StripAccents);
+
+impl Normalizer for StripAccents {
+    /// Strip the normalized string inplace
+    fn normalize(&self, normalized: &mut NormalizedString) -> Result<()> {
+        let mut transforms = Vec::with_capacity(normalized.get().len());
```
n1t0, line 51: You can use `NormalizedString::filter` to avoid having to do this manually:
```rust
normalized.filter(|c| !is_combining_mark(c))
```

n1t0 (commented):
LGTM! We should probably add node bindings at the same time if possible.

tokenizers/src/normalizers/strip.rs
Narsil: Unidecode is a way to get ~ascii from a unicode string:
https://pypi.org/project/Unidecode/
https://github.com/chowdhurya/rust-unidecode

I put it here, because googling "Strip accents unicode" yield unidecode as a first answer, and could be a source of confusion for users/readers. Could be a regular comment instead of doc comment.

tokenizers/src/normalizers/strip.rs
Narsil: That actually fixes a bug. :D

PR #417, opened 2020-09-16
Adding a new "Replace" normalizer that takes a string and replaces it

### LINKED ISSUE #414: Add a normalizer for removing double single quotes
Transformers tokenizer_albert code contains this line:
```python
line = line.replace("``", '"').replace("''", '"')
```

We currently have no way to handle this with normalizers. We need it to be able to emulate this tokenizer.

Either a specialized one, or regexp ?

Narsil:
with another String (for now).
Fixes #414

### DIFF
--- a/bindings/python/src/normalizers.rs
+++ b/bindings/python/src/normalizers.rs
@@ -8,13 +8,8 @@ use crate::error::ToPyResult;
 use serde::ser::SerializeStruct;
 use serde::{Deserialize, Serialize, Serializer};
 use tk::normalizers::{
-<<<<<<< HEAD
-    BertNormalizer, Lowercase, Nmt, NormalizerWrapper, Precompiled, Strip, StripAccents, NFC, NFD,
-    NFKC, NFKD,
-=======
-    BertNormalizer, Lowercase, Nmt, NormalizerWrapper, Precompiled, Replace, Strip, NFC, NFD, NFKC,
-    NFKD,
->>>>>>> Adding a new "Replace" normalizer that takes a string and replaces it
+    BertNormalizer, Lowercase, Nmt, NormalizerWrapper, Precompiled, Replace, Strip, StripAccents,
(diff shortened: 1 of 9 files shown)


bindings/python/src/normalizers.rs
```diff
@@ -310,6 +312,18 @@ impl PyPrecompiled {
     }
 }
 
+#[pyclass(extends=PyNormalizer, module = "tokenizers.normalizers", name=Replace)]
+pub struct PyReplace {}
+#[pymethods]
+impl PyReplace {
+    #[new]
+    fn new(pattern: &PyString, content: &PyString) -> PyResult<(Self, PyNormalizer)> {
```
n1t0, line 320: No need to convert manually, you can do:
```rust
fn new(pattern: String, content: String) -> PyResult<(Self, PyNormalizer)> {
```

tokenizers/src/normalizers/replace.rs
```diff
@@ -0,0 +1,43 @@
...
+impl Replace {
+    pub fn new(pattern: String, content: String) -> Self {
+        Self { pattern, content }
+    }
+}
+
+impl Normalizer for Replace {
+    fn normalize(&self, normalized: &mut NormalizedString) -> Result<()> {
+        let pattern: &str = &self.pattern;
+        normalized.replace(pattern, &self.content)?;
```
n1t0, line 22: You can directly return `normalized.replace(...)` here.

n1t0 (changes requested):
Nice! We can probably already make it generic over `Pattern` in Rust, we just miss some stuff to implement the Python bindings at the moment.

Narsil:
Given that it's relatively hard to put directly a pattern here, can we move this forward for now so we can get `tokenizers` fast into transformers first, then figure out how to make it generic in another issue/PR ?

PR #426, opened 2020-09-21
Update PyO3 to 0.12

n1t0:

PR #427, opened 2020-09-21
Custom components

n1t0:
Opened as draft because it is based on `update-pyo3` but will be merged in master after #426. Ready for review nonetheless.

- Adds `RefMutGuard` and `RefMutContainer` that can be used to safely pass an `&mut T` to Python. The pointer gets invalidated when the `RefMutGuard` gets dropped.
- Adds bindings for `NormalizedString` and `PreTokenizedString`, as well as their "ref mut" counterpart.
- Adds bindings for `Regex`, `Token`, and many other small helper types.
- Adds support for custom `Normalizer`, `PreTokenizer` and `Decoder`. `PostProcessor` is not necessary at the moment since `TemplateProcessing` should be able to achieve most use-cases, and we don't provide any way to modify `Encoding` for now.
- Adds bindings for methods on most of the components (`PostProcessor::process`, `Normalizer::normalize`, ...) since a lot of them were missing.


n1t0:
@sebpuetz I'd love to get your review too if you have a moment!

sebpuetz:
I'll try to take a look later today

bindings/python/py_src/tokenizers/__init__.pyi
```diff
@@ -26,6 +27,232 @@ PreTokenizedEncodeInput = Union[
...
+    def for_each(self, func: Callable[[str], None]):
+        """ Calls the given function for each character of the string """
+        pass
+    def map(self, func: Callable[[str], str]):
+        """ Calls the given function for each character of the string
+
+        Replaces each character of the string using the returned value. Each
+        returned value **must** be a str of length 1 (ie a character).
+        """
+        pass
```
Narsil, line 222: I think that's way too many primitives IMHO.
People should not use our internal API this way.

`replace` seems ok, could it be general enough that it can call `transform` automatically ? or maybe seems like `transform` is the way to go, no ?

bindings/python/src/error.rs
```diff
@@ -7,6 +7,7 @@ use tokenizers::tokenizer::Result;
 #[derive(Debug)]
 pub struct PyError(pub String);
 impl PyError {
+    #[allow(dead_code)]
```
Narsil, line 10: That's odd, no ?

bindings/python/src/models.rs
```diff
@@ -105,6 +106,22 @@ impl PyModel {
         }
     }
 
+    fn tokenize(&self, tokens: &str) -> PyResult<Vec<PyToken>> {
```
Narsil, line 109: Why to we need this ? `encode` is enough, no ?

bindings/python/src/utils/normalization.rs
```diff
@@ -0,0 +1,504 @@
...
+    }
+}
+
+impl From<PySplitDelimiterBehavior> for SplitDelimiterBehavior {
+    fn from(v: PySplitDelimiterBehavior) -> Self {
+        v.0
+    }
+}
+
+fn filter(normalized: &mut NormalizedString, func: &PyAny) -> PyResult<()> {
```
Narsil, line 102: Again that seems a bit too much api here. one very generic function seems better to lower the surface we have to maintain. Rest of the helper functions can be in pure python I feel.

bindings/python/src/utils/pretokenization.rs
```diff
@@ -0,0 +1,256 @@
...
+        }))
+        .into()
+    }
+}
+
+fn tokenize(pretok: &mut PreTokenizedString, func: &PyAny) -> PyResult<()> {
+    if !func.is_callable() {
+        Err(exceptions::PyTypeError::new_err(
+            "`tokenize` expect a callable with the signature: \
+            `fn(str) -> List[Token]`",
```
Narsil, line 53: You had the index of the token in the example, no ?

Narsil: Like with the `normalizer` the api surface seems a bit too much. `pretok.split(Fn[PretokenizedString] -> List[PretokenizedString])` is general enough, no ?

bindings/python/tests/bindings/test_normalizers.py
```diff
@@ -93,3 +94,41 @@ def test_full_strip(self):
...
+
+class TestCustomNormalizer:
+    class BadCustomNormalizer:
+        def normalize(self, normalized, wrong):
+            pass
+
+    class GoodCustomNormalizer:
+        def normalize(self, normalized):
+            self.kept_normalized = normalized
+            normalized.replace("there", "you")
```
Narsil, line 107: Careful with this, it doesn't respect the `replace` of Python which is *not* inplace, but returns a new string.

It will be very confusing for users.

bindings/python/tests/bindings/test_normalizers.py
```diff
@@ -93,3 +94,41 @@ def test_full_strip(self):
...
+        ):
+            good_custom.use_after_normalize()
+
+    def test_normalizer_interface(self):
+        normalizer = Normalizer.custom(TestCustomNormalizer.GoodCustomNormalizer())
+
+        normalized = NormalizedString("Hey there!")
+        normalizer.normalize(normalized)
+
+        assert repr(normalized) == 'NormalizedString(original="Hey you!", normalized="Hey there!")'
```
Narsil, line 133: Isn't the repr reversed here ?

`NormalizedString(original="Hey there!", normalized="Hey you!")`

bindings/python/tests/bindings/test_pre_tokenizers.py
```diff
@@ -119,3 +119,29 @@ def test_instantiate(self):
...
+
+    def test_instantiate(self):
+        bad = PreTokenizer.custom(TestCustomPreTokenizer.BadCustomPretok())
+        good = PreTokenizer.custom(TestCustomPreTokenizer.GoodCustomPretok())
+
+        assert isinstance(bad, PreTokenizer)
+        assert isinstance(good, PreTokenizer)
+        with pytest.raises(Exception, match="TypeError: pre_tokenize()"):
+            bad.pre_tokenize_str("Hey there!")
+        assert good.pre_tokenize_str("Hey there!") == [
```
Narsil, line 144: I think a more realistic example would be better, here I find it a bit confusing, and it would be hard to change this test into creating my own CustomPretok.

Narsil (changes requested):
Nice one.

However I think the API surface for `CustomPretokenizer` and `CustomNormalizer` is too large (hence confusing as a user). Keeping it small to ideally one function (`transform` for normalizer + `split` to pretokenizer`), should make it obvious to users how to do their job. Also exposing `NormalizedString` and `PretokenizedString` fully seems a bit too much (that's arguable).

It would also help, because we definitely need more documentation later (example + test is already a good start)

We should also be careful because the names `split` and `replace` are not inplace in Python by default, that's a confusing spot for users. Making it more pythonesque in the binding is probably a good idea.

bindings/python/examples/custom_components.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class CustomDecoder:
+    def decode(self, tokens):
+        return "".join(tokens)
+
+
+class CustomNormalizer:
+    def normalize(self, normalized):
+        normalized.nfkc()
```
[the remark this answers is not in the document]
n1t0: Agreed the example is not very meaningful. The only API available there is the one from `NormalizedString` though, so using raw Python is not an option.

bindings/python/py_src/tokenizers/__init__.pyi
n1t0: This is not internal API but public API. `transform` is too hard to be used on its own, and each method here make use of it for simplicity.

bindings/python/src/models.rs
n1t0: These serve different purpose. `PreTokenizedString::tokenize` is deep inside the `encode` pipeline, and some use-cases require customization.

bindings/python/src/utils/normalization.rs
n1t0: Nothing can be in pure python unfortunately, for normalization, we have to use the `NormalizedString`. Its goals is to keep track of the alignment, while pure python doesn't. This is just another method on `NormalizedString`.

bindings/python/src/utils/pretokenization.rs
n1t0: Indeed, split takes the index also, nice catch!

bindings/python/tests/bindings/test_normalizers.py
n1t0: Sure, but this is really hard to achieve in Rust without a lot of unnecessary unsafe code, or without cloning. We'll have to be clear about this in our guides/documentation.

bindings/python/src/error.rs
n1t0: We need to get rid of `PyError` and `ToPyResult` altogether, but this is for another PR.

Narsil (approved):

PR #428, opened 2020-09-22
[WIP] Temp work to make the APIs uniform (build from memory by default).

Narsil:
In Python,

We deprecated `BPE(filename, filename)` for `BPE.from_files(filename, filename)` and BPE(vocab, merges) is now the standard constructor.

Same for WordLevel and WordPiece.
Unigram just gets the default treatment as it was never was part of the api in a release before.

### DIFF
--- a/bindings/node/lib/bindings/models.js
+++ b/bindings/node/lib/bindings/models.js
@@ -2,10 +2,12 @@ const native = require("./native");
 
 module.exports = {
   BPE: {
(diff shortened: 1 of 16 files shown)


Narsil:
@n1t0 Do you know why I can't have my staticmethod `from_file(s)` ? The error message is a bit confusing, I could just switch the return result to be something else, but I figure there's a reason of the return type in #[new] in the first place.

PR #429, opened 2020-09-22
Going back for `not` fuse_unk by default for BPE, but add a flag to enable it.

Narsil:


n1t0:
That's better indeed! Thanks!

PR #433, opened 2020-09-23
Replace Normalizer can use Regex

n1t0:
Adds the ability to replace using regular expressions on the `Replace` normalizer.

There are no bindings for the regex pattern for node at the moment, but a they need a big refresh anyway. Let's keep it outside of the scope for now.


Narsil (approved):
LGTM

PR #434, opened 2020-09-24
Upgrading to black 20.8b1

Narsil:


n1t0:
Great, I hope we won't have any unwanted black magic in our PRs now.

PR #441, opened 2020-09-28
Improve template processing

n1t0:
The current version of the `TemplateProcessing` lets us define two sequences `seq_a` and `seq_b`, that end up being concatenated when both are provided. This is actually quite limiting when the template for a pair of sequences is more complicated. (ie XLNet)

This PR modifies the API to actually provide two templates `single` and `pair`, that will be used as relevant, thus allowing for a lot more freedom.


Narsil (approved):
LGTM. I changed a missed replace in `pyi` file, (Not sure why it did not "suggest" instead.).

I think we should have a bit more tests in failure modes, and at least one test to check that type_ids are correctly set.

PR #452, opened 2020-10-09
Fixing Trainer with u8 instead of chars.

Narsil:
Now check both optimized and unoptimized encodings schemes for Unigram.


tokenizers/src/models/unigram/lattice.rs
```diff
@@ -220,11 +226,14 @@ impl<'a> Lattice<'a> {
     }
 
     pub fn piece(&self, node: &Node) -> String {
-        self.sentence
-            .chars()
-            .skip(node.pos)
-            .take(node.length)
-            .collect()
+        String::from_utf8(
```
n1t0, line 229: Can't we do `self.sentece[node.pos..node.pos+length].to_owned()`?

tokenizers/src/models/unigram/model.rs
```diff
@@ -24,6 +24,7 @@ pub struct Unigram {
     pub(super) eos_id: usize,
 
     fuse_unk: bool,
+    is_optimized: bool,
```
n1t0, line 27: This is going to be serialized, maybe you want to `#[serde(skip)]` it

tokenizers/tokenizer.json
```diff
@@ -0,0 +1,194 @@
+{
```
n1t0, line 1: Haha, nope!

n1t0 (approved):
Just a couple things to verify (see comments below), but otherwise LGTM.

tokenizers/tokenizer.json
Narsil: Ahah

PR #454, opened 2020-10-10
Proper fixing of new clippy lints.

Narsil:


tokenizers/src/tokenizer/normalizer.rs
```diff
@@ -490,11 +490,14 @@ impl NormalizedString {
...
+            .filter_map(|o| o)
             .collect::<Vec<_>>();
-        self.transform(
-            filtered.into_iter().rev().filter_map(|o| o),
-            removed as usize,
-        );
+        // We can't use double rev, it get optimized away and make the code wrong.
+        // Re-iterating with rev will trigger clippy error.
+        // Reversing here seems slow but should not be a problem as it should
+        // help cache locality.
```
sebpuetz, line 498: Rust's iterators are evaluated lazily, double `rev()`ing without collecting cancels out the reverse: https://github.com/huggingface/tokenizers/pull/344#issuecomment-661054887

Replacing the `into_iter().rev()` with the in-place `reverse()` is probably not optimal either since `into_iter().rev()` is essentially free as it just iterates from the end and in contrast to `reverse()` does not need to swap all elements of the vector.

Was there a clippy lint complaining about the `into_iter().rev()` part?

Narsil: Yes. It complains that `needless_collect`. Basically that we collect something that we iterate over afterwards.

We could definitely use `[allow(needless_collect)]` But I think we should try to fix them anyway. We probably also could have only one forward pass for this one.

But how does `iter().rev()` handle cache locality though ?

Narsil: I fixed it to only forward pass anyway, it's just better.

tokenizers/src/tokenizer/encoding.rs
```diff
@@ -144,6 +144,17 @@ impl Encoding {
         std::mem::replace(&mut self.overflowing, vec![])
     }
 
+    pub fn process_tokens_with_offsets_mut<F>(&mut self, func: F)
```
n1t0, line 147: Let's make this `pub(crate)` for now.

n1t0 (commented):
LGTM! The forward pass is better indeed than the reverse in-place.

PR #455, opened 2020-10-10
Moving to manylinux2010 and remove nightly on Windows.

Narsil:


Narsil:
@n1t0  Au fait on pourrait la passer celle là, ca permet de revenir en 1.47 au moins

n1t0:
Yes, I just wanted to make sure it works as expected before merging. I don't like surprises when doing a release :slightly_smiling_face: 

Also I think we should check if we can add support for PowerPC and aarch64 using the appropriate manylinux2014.

PR #459, opened 2020-10-13
Fixing a bug where long tokenizer files would be incorrectly deserialized

Narsil:
- Add a bunch of tests to check deserialization behaviour
- One tests also confirms current Single deserialization of Sequence.

### DIFF
--- a/bindings/python/tests/utils.py
+++ b/bindings/python/tests/utils.py
@@ -75,6 +75,16 @@ def train_files(data_dir):
     }
 
 
+@pytest.fixture(scope="session")
+def precompiled_files(data_dir):
+    albert_base = download(
+        "https://s3.amazonaws.com/models.huggingface.co/bert/albert-base-v1-tokenizer.json"
+    )
+    return {
+        "albert_base": albert_base,
+    }
+
+
 def multiprocessing_with_parallelism(tokenizer, enabled: bool):
     """
     This helper can be used to test that disabling parallelism avoids dead locks when the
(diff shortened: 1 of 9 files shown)


bindings/python/tests/utils.py
```diff
@@ -75,6 +75,16 @@ def train_files(data_dir):
     }
 
 
+@pytest.fixture(scope="session")
+def precompiled_files(data_dir):
```
n1t0, line 79: What about `albert_tokenizer` instead of `precompiled_files` which seems off? Also, we can probably return `albert_base` directly since its a single file in this case.

n1t0 (approved):
Nice catch! It should be faster like this anyway. Just added a couple comments about naming.

bindings/python/tests/utils.py
Narsil: I'd like to keep the dictionary format, at least for uniformity within the tests fixtures that all use dictionaries.

In terms of naming, because we might add follow ups saved tokenizers, what about `serialized_files` it makes `serialized_files["albert_base"]` more understandable. I'd rather add another tokenizer in the tests than naming specifically for albert IMO. (We probably should have a handful of serialized tokenizers in those tests to make sure will load the old ones, with BertNormalizer, BPE, and so on). What do you think ?

n1t0: Yes I agree that we'll probably add more, but it might be better to keep each set of files separated though. When you require `albert-base` you don't necessarily want to download all the serialized files, and having logically separated fixtures let us do that.
I don't mind keeping a dict, but I felt that with serialized tokenizers it would just add repetition (as opposed to the vocab/merges used for the others).

Narsil: Good argument for downloading separately, and that changes the names, which makes your solution the best in the end.

PR #462, opened 2020-10-13
Fixing sampling test heuristics to fail less often.

### LINKED ISSUE #461: Test fails in `--release` mode
The test `test_sample` from `tests/unigram.rs` fails if compiled in release mode.

```
cargo test --test unigram --release

thread 'test_sample' panicked at 'assertion failed: `(left !== right)` (left: `0.294`, right: `0.2496817458757616`, expect diff: `0.03`, real diff: `0.04431825412423837`)', tests/unigram.rs:100:13
```

Two things to note:
- This only happens if compiled with the `--release` flag and excluded otherwise, because the test is annotated with `#[cfg(not(debug_assertions))]`, that's why the CI doesn't catch this.
- The test seems to be non-deterministic, it fails for roughly 9 out of 10 cases and succeeds occasionally.

Narsil:
Fixes #461

### DIFF
--- a/tokenizers/tests/unigram.rs
+++ b/tokenizers/tests/unigram.rs
@@ -88,7 +88,7 @@ fn test_sample() {
             *p /= z;
         }
 
-        let n_trials = 1_000;
+        let n_trials = 10_000;
         let mut freq: HashMap<String, u32> = HashMap::new();
         for _ in 0..n_trials {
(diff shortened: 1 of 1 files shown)


janpetschexain (approved):
it actually fixes #461, not #459.

Narsil:
Thanks @janpetschexain

n1t0:
:ok_hand:

Issue #157, opened 2020-02-19
How can we output the progress bars within a jupyter notebook?


ohmeow:
... in particular when we run `tokenizer.train(...)`

n1t0:
Indeed, this is a known limitation of the library we use to show progress. Unfortunately, it won't show anything if not attached to a "real" terminal, so it does not work in Notebooks.

jonaspleyer:
Has there been any update on this? What would be necessary steps to resolve this issue. I have a large library and have so far been very satified with indicatif but this might be a dealbreaker in the long run.

n0kovo:
> Indeed, this is a known limitation of the library we use to show progress

So uh, could this "known limitation" maybe be considered, you know, an... issue?
Considering  that a pretty huge chunk of `tokenizers` users primarily use Jupyter/Colab etc.?
Just thinking out loud 🤷

PR #463, opened 2020-10-14
Add progress bar feature flag

### LINKED ISSUE #270: indicatif compilation feature flag?
We are using Flutter/Dart for cross compiling for Desktop / iOS / Droid applications, and I had to fork HuggingFace tokenizers in order to get Cargo Lipo to work.  Indicatif progress bar has a dependency on termios (https://docs.rs/termios/0.3.2/termios/) that only supports the following Operating System platforms:
- linux
- macos
- freebsd
- openbsd
- netbsd
- dragonfly
- solaris

Whenever running cargo lipo to target iOS platforms hugging face will bomb out everytime because it cannot find the target operating system.  
```
error[E0433]: failed to resolve: could not find `target` in `os`
   --> /Users/user/.cargo/registry/src/github.com-1ecc6299db9ec823/termios-0.3.2/src/lib.rs:210:31
    |
210 |     fn deref(&self) -> &::os::target::termios {
    |                               ^^^^^^ could not find `target` in `os`
```

When I forked and deleted the deps on indactif, everything compiled great.

janpetschexain:
Fixes #270.

We ran into the same issue as #270 while compiling for ios targets and decided to provide a pr. This pr introduces the `progressbar` feature flag, which makes the `indicatif` dependencies and its usage optional (but it's still set as default to avoid breaking changes).

### DIFF
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -47,7 +47,7 @@ clap = "2.33"
 unicode-normalization-alignments = "0.1"
 unicode_categories = "0.1"
 unicode-segmentation = "1.6"
-indicatif = "0.14"
+indicatif = { version = "0.15", optional = true }
 itertools = "0.9"
 log = "0.4"
 esaxx-rs = "0.1"
@@ -58,3 +58,7 @@ spm_precompiled = "0.1"
 criterion = "0.3"
 tempfile = "3.1"
 assert_approx_eq = "1.1"
+
+[features]
+default = ["progressbar"]
+progressbar = ["indicatif"]
--- a/tokenizers/README.md
+++ b/tokenizers/README.md
@@ -72,7 +72,7 @@ fn main() -> Result<()> {
     let vocab_size: usize = 100;
 
     let trainer = BpeTrainerBuilder::new()
-        .show_progress(true)
+        .show_progress(true) // requires "progressbar" feature
         .vocab_size(vocab_size)
         .min_frequency(0)
         .special_tokens(vec![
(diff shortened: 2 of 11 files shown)


janpetschexain:
regarding the readme mismatch:
```
Run cargo readme > must_match_readme.md && diff must_match_readme.md README.md
72,73d71
<     # #[cfg(feature = "progressbar")]
<     # {
107d104
<    # }
123,124c120,121
< compilation for certain targets is not supported by the [termios](https://crates.io/crates/termios)
< dependency of the [indicatif](https://crates.io/crates/indicatif) progress bar.
---
>   compilation for certain targets is not supported by the [termios](https://crates.io/crates/termios)
>   dependency of the [indicatif](https://crates.io/crates/indicatif) progress bar.
Error: Process completed with exit code 1.
```

do you prefer to:
- have the missing lines to be added to the readme.md, but those will not be hidden like in the generated docs from the lib.rs
- to mark the code section in the lib.rs as `ignore` instead of `no_run`, then the hidden cfg is not needed b/c the doc code is not compiled
- to make the hidden cfg line visible in both lib.rs and readme.md and to use `#[cfg(feature = "progressbar")]` and `#[cfg(not(feature = "progressbar"))]` properly, but this might clutter the actual example
?

Narsil:
I think this adds a bit too many compilation flag for us to support. 
Thanks for the PR anyway, we'll probably use it as a starting point.

Just FYI we also have issue #157 that we would like to fix at the same time.

What's probably going to happen is:

- should_progress flags will remain for everyone
- Progressbar is going to be internalized in the library with different backends support
  - Indicatif in Rust
  - Raw string in Rust with compilation flag.
  - tqdm/tqdm_notebook in Python
  - cli-progress in Node, maybe ?


NB: The Readme error you're encountering is just a sanity check that the README.md *must* come from the Rust doc, just run `cargo readme > README.md` to fix

PR #480, opened 2020-10-22
Last missing pieces for Unigram + ByteLevel

n1t0:
In order to fully support the byte-level and its idea of not having an unk token, we need to support the `initial_alphabet`.
Also adds the ability to add `special_tokens`.

### DIFF
--- a/bindings/python/py_src/tokenizers/trainers/__init__.pyi
+++ b/bindings/python/py_src/tokenizers/trainers/__init__.pyi
@@ -46,7 +46,7 @@ class BpeTrainer(Trainer):
             initial_alphabet: List[str]:
                 A list of characters to include in the initial alphabet, even
                 if not seen in the training dataset.
-                If the strings contains more than one character, only the first one
+                If the strings contain more than one character, only the first one
                 is kept.
 
             continuing_subword_prefix: Optional[str]:
@@ -98,7 +98,7 @@ class WordPieceTrainer(Trainer):
             initial_alphabet: List[str]:
                 A list of characters to include in the initial alphabet, even
                 if not seen in the training dataset.
-                If the strings contains more than one character, only the first one
+                If the strings contain more than one character, only the first one
                 is kept.
 
             continuing_subword_prefix: Optional[str]:
@@ -135,6 +135,12 @@ class UnigramTrainer(Trainer):
 
             special_tokens: List[Union[str, AddedToken]]:
                 A list of special tokens the model should know of.
+
+            initial_alphabet: List[str]:
+                A list of characters to include in the initial alphabet, even
+                if not seen in the training dataset.
+                If the strings contain more than one character, only the first one
+                is kept.
 
         Returns:
             Trainer
--- a/tokenizers/src/models/unigram/model.rs
+++ b/tokenizers/src/models/unigram/model.rs
@@ -53,8 +53,8 @@ impl Clone for Unigram {
 
 impl std::fmt::Debug for Unigram {
     fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
-        fmt.debug_struct("BPE")
-            .field("vocab", &self.vocab.len())
+        fmt.debug_struct("Unigram")
+            .field("vocab", &self.vocab)
             .field("unk_id", &self.unk_id)
             .finish()
     }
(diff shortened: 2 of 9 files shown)


bindings/python/py_src/tokenizers/trainers/__init__.pyi
```diff
@@ -136,6 +136,12 @@ class UnigramTrainer(Trainer):
             special_tokens: List[Union[str, AddedToken]]:
                 A list of special tokens the model should know of.
 
+            initial_alphabet: List[str]:
```
Narsil, line 139: Hmm I kind of think we should throw an error instead of silently taking the first char instead.  I understand it's to be coherent with other "initial_alphabets" but I think people don't read docs/docstrings as much as we would like, throwing an error would be more helpful to them.

tokenizers/src/models/unigram/trainer.rs
```diff
@@ -94,39 +96,40 @@ impl UnigramTrainer {
...
-        let mut pieces: HashMap<String, f64> = HashMap::new();
+        let mut pieces: Vec<(String, f64)> = vec![];
+        let mut inserted: HashSet<String> = HashSet::new();
         let existing_pieces: HashMap<String, f64> = model.iter().cloned().collect();
-        // XXX: Make sure bos, eos and unk exists and are ids 0, 1, 2
-        pieces.insert(self.unk_token.clone(), 0.0);
         for c in required_chars {
             if let Some(t) = existing_pieces.get(&c) {
-                pieces.insert(c, *t);
+                inserted.insert(c.clone());
```
Narsil, line 104: Is the hashSet `inserted` really necessary ? `required_chars` is already a HashMap, so we shouldn't get duplicates anyway, no ?

tokenizers/src/models/unigram/trainer.rs
```diff
@@ -421,6 +424,11 @@ impl UnigramTrainer {
             Vec::with_capacity(self.vocab_size.try_into().unwrap());
         // XXX: Make sure unk exists and are ids 0
```
Narsil, line 425: If we want to get rid of unk_token in the general case, I think we need to remove it from here no ? Or are we still keeping it for now, because we can't get rid of it in the trainer ?

tokenizers/src/models/unigram/trainer.rs
```diff
@@ -421,6 +424,11 @@ impl UnigramTrainer {
...
         pieces.push((self.unk_token.clone(), f64::NAN));
+        pieces.extend(
+            self.special_tokens
+                .iter()
+                .map(|tok| (tok.content.clone(), f64::NAN)),
```
Narsil, line 430: I would feel more confident if we had a test for this NaN behaviour.

It's supposed to be taken care of upward in the stack I know, but I don't think there's actually a test to train on content that contains that kind of token, *and* that the actual tokenization of the resulting tokenizer does handle this special tokens correctly.

tokenizers/src/models/unigram/model.rs
```diff
@@ -53,8 +53,8 @@ impl Clone for Unigram {
 
 impl std::fmt::Debug for Unigram {
     fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
-        fmt.debug_struct("BPE")
-            .field("vocab", &self.vocab.len())
+        fmt.debug_struct("Unigram")
+            .field("vocab", &self.vocab)
```
Narsil, line 57: That's quite a large log. There's only vocab.len() in BPE for instance. Is there a reason to output the whole vocab here ?

Narsil (changes requested):
I think my main concern is to have a bit more tests around the special tokens handling. Just to make sure all the NaNs don't destroy the algo.

tokenizers/src/models/unigram/model.rs
n1t0: Oops my bad, needed it for debug, but we should keep the `len` yes!

tokenizers/src/models/unigram/trainer.rs
n1t0: `inserted` has been added here because we are changing `pieces` to be a `Vec<(String, f64)>` instead of a `HashMap<String, f64>`, so we still need a way to check if a token has been inserted with `O(1)`

tokenizers/src/models/unigram/trainer.rs
n1t0: We are keeping it for now. It leads to breaking changes that are best left for another PR, for another release.

tokenizers/src/models/unigram/trainer.rs
Narsil: My bad, did not see that change to `Vec`

Narsil (approved):
LGTM ! Super nice

PR #481, opened 2020-10-23
Fixed `BPE.read_files` -> `BPE.read_file` in SentencePieceBPETokenizer

esceptico:
Fixed a line that was apparently forgotten to replace (`read_files` -> `read_file`) in #430


n1t0:
Thanks @esceptico, nice catch!

PR #492, opened 2020-10-31
makes from_file() method static

taufique74:
`from_file()` method should be static in `BerWordPieceTokenizer`.


n1t0:
Nice catch! Thank you @taufique74

PR #500, opened 2020-11-04
typo: from_files() renamed to from_file()

taufique74:
`BPE` model doesn't have any method called `from_files()`.
It should be changed to `from_file()` instead. This exists in Quicktour doc too.


n1t0:
Thanks @taufique74

PR #506, opened 2020-11-04
Improve Encoding mappings for pairs of sequence

### LINKED ISSUE #486: Improve `Encoding` mappings for pairs of sequence
The `words` indices, the `offsets`, and all the related methods don't provide information about which input sequence the index is based on.

Here is an example of the `Encoding.words` output when a pair of sequences was encoded:
```
[None, 0, 1, 2, 2, 2, 2, 3, 4, None, 0, 0, 0, 0, 1, 2, 3, 4, 5, None, None, None, None, None, None]
```

We should include a sequence index in the relevant outputs, and provide an argument to choose the target input sequence on the relevant methods.

n1t0:
Fix #486 

Improve the mappings `XX_to_XX` methods by adding support for pairs of sequence:
- `token_to_XX` methods now return a tuple with the sequence index and the expected result for pairs of sequence
- `char_to_XX` and `word_to_XX` now accept a new argument `sequence_index` to specify the target input sequence
- Also add a `token_to_sequence` method that can tell for any token, the associated input sequence.

### DIFF
--- a/bindings/node/native/src/encoding.rs
+++ b/bindings/node/native/src/encoding.rs
@@ -115,7 +115,7 @@ declare_types! {
         }
 
         method getWords(mut cx) {
-            // getWords(): number[]
+            // getWords(): (number | undefined)[]
 
             let this = cx.this();
             let guard = cx.lock();
@@ -123,6 +123,18 @@ declare_types! {
                 .encoding.as_ref().expect("Uninitialized Encoding")
                 .get_words()
                 .to_vec();
+
+            Ok(neon_serde::to_value(&mut cx, &ids)?)
+        }
+
+        method getSequences(mut cx) {
+            // getSequences(): (number | undefined)[]
+
+            let this = cx.this();
+            let guard = cx.lock();
+            let ids = this.borrow(&guard)
+                .encoding.as_ref().expect("Uninitialized Encoding")
+                .get_sequences();
 
             Ok(neon_serde::to_value(&mut cx, &ids)?)
         }
--- a/bindings/node/native/src/encoding.rs
+++ b/bindings/node/native/src/encoding.rs
@@ -27,6 +27,28 @@ declare_types! {
             Ok(cx.number(length as f64).upcast())
         }
 
+        method getNSequences(mut cx) {
+            let this = cx.this();
+            let guard = cx.lock();
+            let n = this.borrow(&guard)
+                .encoding.as_ref().expect("Uninitialized Encoding")
+                .n_sequences();
+
+            Ok(cx.number(n as f64).upcast())
+        }
+
+        method setSequenceId(mut cx) {
+            let seq_id = cx.extract::<usize>(0)?;
+
+            let mut this = cx.this();
+            let guard = cx.lock();
+            this.borrow_mut(&guard)
+                .encoding.as_mut().expect("Uninitialized Encoding")
+                .set_sequence_id(seq_id);
+
+            Ok(cx.undefined().upcast())
+        }
+
         method getIds(mut cx) {
             // getIds(): number[]
 
@@ -145,16 +167,17 @@ declare_types! {
         }
 
         method wordToTokens(mut cx) {
-            // wordToTokens(word: number): [number, number] | undefined
+            // wordToTokens(word: number, seqId: number = 0): [number, number] | undefined
 
             let word = cx.extract::<u32>(0)?;
+            let seq_id = cx.extract_opt::<usize>(1)?.unwrap_or(0);
 
             let this = cx.this();
             let guard = cx.lock();
 
             let res = this.borrow(&guard)
                 .encoding.as_ref().expect("Uninitialized Encoding")
-                .word_to_tokens(word);
+                .word_to_tokens(word, seq_id);
 
             if let Some(tokens) = res {
                 Ok(neon_serde::to_value(&mut cx, &tokens)?)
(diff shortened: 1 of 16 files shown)


bindings/node/native/src/encoding.rs
```diff
@@ -182,53 +206,92 @@ declare_types! {
...
+                let encoding = borrowed.encoding.as_ref().expect("Uninitialized Encoding");
+
+                let res = encoding.token_to_word(token);
+                let n_seq = encoding.n_sequences();
+                (res, n_seq)
+            };
+
+            if let Some((seq_id, index)) = res {
+                if n_seq > 1 {
+                    Ok(neon_serde::to_value(&mut cx, &(seq_id, index))?)
```
Narsil, line 275: TBH variant return types always give me shivers. It's usually so much better to have 2 different functions instead. Saves so much headaches for users....

bindings/python/py_src/tokenizers/__init__.pyi
```diff
@@ -368,68 +383,110 @@ class Encoding:
...
             token_index (:obj:`int`):
                 The index of a token in the encoded sequence.
 
         Returns:
-            :obj:`Tuple[int, int]`: The token offsets :obj:`(first, last + 1)`
+            :obj:`int`: The sequence id of the given token
         """
         pass
-    def token_to_word(self, token_index: int) -> Optional[int]:
+    def token_to_chars(self, token_index: int) -> Optional[Union[Offsets, Tuple[int, Offsets]]]:
```
Narsil, line 427: Same here for variant return types.

bindings/python/py_src/tokenizers/__init__.pyi
```diff
@@ -275,6 +275,21 @@ class Encoding:
...
         pass
     @property
+    def n_sequences(self) -> int:
+        """The number of sequences represented
+
+        Returns:
+            :obj:`int`: The number of sequences in this :class:`~tokenizers.Encoding`
+        """
+        pass
+    def set_sequence_id(self, sequence_index: int):
```
Narsil, line 285: Is that really something we need to expose ?

bindings/python/py_src/tokenizers/__init__.pyi
```diff
@@ -368,68 +383,110 @@ class Encoding:
...
+
+            - For a single sequence: the token offsets:
+              :obj:`Tuple[int, int]` of the form :obj:`(first, last + 1)`
+
+            - For pairs of sequence: A tuple with the sequence index, and the token offsets:
+              :obj:`Tuple[int, Tuple[int, int]]` with offsets of the form :obj:`(first, last + 1)`
+
+        """
+        pass
+    def token_to_word(self, token_index: int) -> Optional[Union[int, Tuple[int, int]]]:
```
Narsil, line 449: Another variant return type

tokenizers/src/tokenizer/encoding.rs
```diff
@@ -22,6 +24,9 @@ pub struct Encoding {
     attention_mask: Vec<u32>,
     /// A list of overflowing Encoding generated when we got truncated
     overflowing: Vec<Encoding>,
+    /// Ranges of tokens covered by each sequence. If this is empty we consider
+    /// there is only one sequence in this Encoding, and that it covers the entire range.
```
Narsil, line 28: Is that a speed optimization ? We probably don't need that optimization until proven otherwise, no ?

tokenizers/src/tokenizer/encoding.rs
```diff
@@ -155,11 +180,40 @@ impl Encoding {
...
+    /// given sequence id
+    fn sequence_range(&self, sequence_id: usize) -> Range<usize> {
+        self.sequence_ranges
+            .get(&sequence_id)
+            .cloned()
+            .unwrap_or(0..self.len())
+    }
+
+    /// Returns the index of the sequence containing the given token
+    pub fn token_to_sequence(&self, token: usize) -> Option<usize> {
```
Narsil, line 193: Removing empty hashmap is empty mean one sequence over the whole lenght would really simplify a lot of the code too.

tokenizers/src/processors/bert.rs
```diff
@@ -105,6 +115,9 @@ impl PostProcessor for BertProcessing {
                 [&vec![0u32; encoding.get_type_ids().len()][..], &[1]].concat();
             let pair_attention_mask = vec![1; pair_ids.len()];
 
+            // For compatibility with `TemplateProcessing`, the sequence_ranges shouldn't contain
+            // the special tokens.
+            let pair_sequence_ranges = HashMap::from_iter(vec![(1, 0..pair_ids.len() - 1)]);
```
Narsil, line 120: It's not exactly clear why this code needs to care about the ranges and those hardcoded values feel a bit scary (-1, +1 etc..).

Could be at least have a test that would exhibit the error in case this line was wrong ? It would be even better if we didn't need the hardcodedness.

Narsil (changes requested):
The choice I would really think needs to be changed it the variant return type, it's really messy for users I think. I gave an example of something which I feel would both keep backcompat and be cleaner for users.

`set_sequence_id` being exposed feels odd to me, I think we should probably add a realistic test where this might be used in production to convey why it's there (because destroying sequence sources being irreversible, we need a good reason to expose that IMO). Otherwise we should remove it until a convincing scenario is presented I feel.

bindings/node/native/src/encoding.rs
n1t0: As discussed, we'll remove the output with the sequence id, and will just keep the classic one. The user can call `token_to_sequence` when needed to determine it.

bindings/python/py_src/tokenizers/__init__.pyi
n1t0: Yes, it's just for consistency. The user can manually merge multiple `Encoding`s, and this allows to manually set the sequence id for each of them.

tokenizers/src/tokenizer/encoding.rs
n1t0: It just simplifies the code since we don't need to keep it updated when everything else changes.

PR #512, opened 2020-11-09
[WIP] train from memory.

### LINKED ISSUE #198: Training a model from in-memory data
Hi,
How could I change the code so that it is possible to train a model from  in-memory data instead of using files?
Basically, changing
`tokenizer.train(["wiki.test.raw"], vocab_size=20000)`
by 
`tokenizer.train(data_array, vocab_size=20000)`
considering that **data_array** is e.g. an array of sentences ["First sentence", "second sentence...].
Thanks for your work!
Best, Loic

Narsil:
First attempt to train from memory.

Nice things:
- It works
- It's relatively simple
- It enables shifting paradigm so that `feed_train_chunk` could start to
  directly feed some inner struct of `Trainer` (pairs for BPE, sentences
for Unigram) and bypass `word_counts` later.
- `datasets` training is somehow as fast as from raw file (~20% slowdown
on wikitext-103-raw ?).

Negatives:
- Puts burden on the caller to feed chunks
large enough that parallelization actually achieves something.
- Adds a new item in `Tokenizer` struct that does not really belong (it
belongs to the trainer and could be different from words, but this state
is temporary).
- In the case of `datasets` we are still having a slowdown because we
have to recreate a `string` chunk for a list of strings that gets
splitted later.
- We are accepting only strings, not raw bytes which does incur some
overhead but really not the focus here.
- `{train|train_and_replace}_{from_files|from_buffer}` is starting to get a bit silly.

Still missing:
- Adding the docs
- Rust tests.
- Node bindings

Fixes #198

### DIFF
--- a/bindings/python/examples/train_datasets.py
+++ b/bindings/python/examples/train_datasets.py
@@ -1,0 +1,33 @@
+import datasets
+from tokenizers import normalizers, pre_tokenizers, Tokenizer, models, trainers
+import tqdm
+
+
+def main():
+    dataset = datasets.load_dataset("wikitext", "wikitext-103-raw-v1")
+    bpe_tokenizer = Tokenizer(models.BPE())
+    bpe_tokenizer.pre_tokenizer = pre_tokenizers.Whitespace()
+    bpe_tokenizer.normalizer = normalizers.Lowercase()
+    trainer = trainers.BpeTrainer(show_progress=True)
+
+    def data_iterator():
+        L = len(dataset["train"])
+        B = 30000
+        for i in tqdm.tqdm(range(0, L, B)):
+            batch = dataset["train"][i : i + B]
+            # print("H", end="", flush=True)
+            yield "".join(batch["text"])
+
+    bpe_tokenizer.train_from_iterator(trainer, data_iterator())
+
+
+def main2():
+    trainer = trainers.BpeTrainer(show_progress=True)
+    bpe_tokenizer = Tokenizer(models.BPE())
+    bpe_tokenizer.pre_tokenizer = pre_tokenizers.Whitespace()
+    bpe_tokenizer.normalizer = normalizers.Lowercase()
+    bpe_tokenizer.train(trainer, ["/home/nicolas/data/wikitext-103-raw/wiki.train.raw"])
+
+
+if __name__ == "__main__":
+    main()
--- a/bindings/python/src/tokenizer.rs
+++ b/bindings/python/src/tokenizer.rs
@@ -1035,6 +1035,33 @@ impl PyTokenizer {
         Ok(self.tokenizer.add_special_tokens(&tokens))
     }
 
+    fn train_from_iterator(&mut self, trainer: &PyTrainer, iterator: &PyAny) -> PyResult<()> {
+        let gil = Python::acquire_gil();
+        let mut iterator: PyIterator = iterator.iter()?;
+        while let Ok(Some(content)) = {
+            if let Some(value) = iterator.next() {
+                let item: &PyAny = value?;
+                if let Ok(s) = item.downcast::<PyString>() {
+                    let content: &str = s.to_str()?;
+                    Ok(Some(content))
+                } else {
+                    Err(exceptions::PyTypeError::new_err(
+                        "Invalid string in python iterator",
+                    ))
+                }
+            } else {
+                Ok(None)
+            }
+        } {
+            self.tokenizer
+                .feed_train_chunk(trainer, content)
+                .map_err(|_| exceptions::PyException::new_err("Cannot feed chunk"))?;
+        }
+        gil.python().allow_threads(|| {
+            ToPyResult(self.tokenizer.train_and_replace_from_buffer(trainer)).into()
+        })
+    }
+
     fn train(&mut self, trainer: &PyTrainer, files: Vec<String>) -> PyResult<()> {
         let gil = Python::acquire_gil();
         gil.python()
--- a/bindings/python/tests/bindings/test_trainers.py
+++ b/bindings/python/tests/bindings/test_trainers.py
@@ -12,6 +12,41 @@ from tokenizers import (
 )
 from ..utils import data_dir, train_files
 
+ZEN_OF_PYTHON = """
+[CLS] The Zen of Python, by Tim Peters [SEP]
+[CLS] Beautiful is better than ugly. [SEP]
+[CLS] Explicit is better than implicit. [SEP]
+[CLS] Simple is better than complex. [SEP]
+[CLS] Complex is better than complicated. [SEP]
+[CLS] Flat is better than nested. [SEP]
+[CLS] Sparse is better than dense. [SEP]
+[CLS] Readability counts. [SEP]
+[CLS] Special cases aren't special enough to break the rules. [SEP]
+[CLS] Although practicality beats purity. [SEP]
+[CLS] Errors should never pass silently. [SEP]
+[CLS] Unless explicitly silenced. [SEP]
+[CLS] In the face of ambiguity, refuse the temptation to guess. [SEP]
+[CLS] There should be one-- and preferably only one --obvious way to do it. [SEP]
+[CLS] Although that way may not be obvious at first unless you're Dutch. [SEP]
+[CLS] Now is better than never. [SEP]
+[CLS] Although never is often better than *right* now. [SEP]
+[CLS] If the implementation is hard to explain, it's a bad idea. [SEP]
+[CLS] If the implementation is easy to explain, it may be a good idea. [SEP]
+[CLS] Namespaces are one honking great idea -- let's do more of those! [SEP]
+"""
+
+
+class TestTrain:
+    def test_train_from_memory(self):
+        bpe_tokenizer = Tokenizer(models.BPE())
+        bpe_tokenizer.normalizer = normalizers.Lowercase()
+        trainer = trainers.BpeTrainer(show_progress=False)
+
+        def data_iterator():
+            yield ZEN_OF_PYTHON
+
+        bpe_tokenizer.train_from_iterator(trainer, data_iterator())
+
 
 class TestUnigram:
     def test_train(self, train_files):
@@ -46,30 +81,7 @@ class TestUnigram:
     def test_train_with_special_tokens(self):
         filename = "tests/data/dummy-unigram-special_tokens-train.txt"
         with open(filename, "w") as f:
-            f.write(
-                """
-[CLS] The Zen of Python, by Tim Peters [SEP]
-[CLS] Beautiful is better than ugly. [SEP]
-[CLS] Explicit is better than implicit. [SEP]
-[CLS] Simple is better than complex. [SEP]
-[CLS] Complex is better than complicated. [SEP]
-[CLS] Flat is better than nested. [SEP]
-[CLS] Sparse is better than dense. [SEP]
-[CLS] Readability counts. [SEP]
-[CLS] Special cases aren't special enough to break the rules. [SEP]
-[CLS] Although practicality beats purity. [SEP]
-[CLS] Errors should never pass silently. [SEP]
-[CLS] Unless explicitly silenced. [SEP]
-[CLS] In the face of ambiguity, refuse the temptation to guess. [SEP]
-[CLS] There should be one-- and preferably only one --obvious way to do it. [SEP]
-[CLS] Although that way may not be obvious at first unless you're Dutch. [SEP]
-[CLS] Now is better than never. [SEP]
-[CLS] Although never is often better than *right* now. [SEP]
-[CLS] If the implementation is hard to explain, it's a bad idea. [SEP]
-[CLS] If the implementation is easy to explain, it may be a good idea. [SEP]
-[CLS] Namespaces are one honking great idea -- let's do more of those! [SEP]
-            """
-            )
+            f.write(ZEN_OF_PYTHON)
 
         tokenizer = Tokenizer(models.Unigram())
         trainer = trainers.UnigramTrainer(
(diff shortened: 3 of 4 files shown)


n1t0:
Nice! I've been thinking about your comments and experimented a bit, and I think I might have something that can avoid almost/all the negative points stated above, without complicating things too much and even simplifying some others.

> It enables shifting paradigm so that `feed_train_chunk` could start to
  directly feed some inner struct of `Trainer` (pairs for BPE, sentences
for Unigram) and bypass `word_counts` later.

I'm not sure to understand exactly how you would feed types specific to each `Trainer` without being too specific about each of them in the `Tokenizer`. I think it should stay the responsibility of each `Trainer` to process the data, and as such, having them process a bunch of `&str` seems the most straightforward way to do it for me.

> Puts burden on the caller to feed chunks
large enough that parallelization actually achieves something.

I think the way to avoid this is to actually parallelize over the iterator instead of each element. We could absolutely be training over an array with small sentences, and I even feel like it will be the case most of the time. Plus, we can expect to get at least as many elements in the iterator as the number of threads in the pool, while we can't really expect each string to be long enough and to include some `\n`.

> - Adds a new item in `Tokenizer` struct that does not really belong (it
belongs to the trainer and could be different from words, but this state
is temporary).
> - In the case of `datasets` we are still having a slowdown because we
have to recreate a `string` chunk for a list of strings that gets
splitted later.

These are not necessary with the solution below. `datasets` can probably directly feed a NumPy array, without the need for any sort of pre-processing.

> `{train|train_and_replace}_{from_files|from_buffer}` is starting to get a bit silly.

With #519, we remove `train_and_replace` to keep only `train`, that will be training the attached `Model` in-place. Makes everything easier.

So in order to achieve this, I think we can have something along the lines of: 
```rust
pub trait Trainer {
    [...]

    // No need to give anything else than the Model to train
    fn train(&self, model: &mut Self::Model) -> Result<Vec<AddedToken>>;

    // We feed the `Trainer` with an iterator, and give him a way to pre-process each sequence
    fn feed<I, S, F>(&mut self, iterator: I, process: F) -> Result<()>
    where
        I: Iterator<Item = S> + Send,
        S: AsRef<str> + Send,
        F: Fn(&str) -> Result<Vec<String>> + Sync;
}

// Example of impl to handle word counts. We can factorize the inside to easily
// re-use it in the various Trainer for now.
fn feed<I, S, F>(&mut self, iterator: I, process: F) -> Result<()>
where
    I: Iterator<Item = S> + Send,
    S: AsRef<str> + Send,
    F: Fn(&str) -> Result<Vec<String>> + Sync,
{
    let words: Result<HashMap<String, u32>> = iterator
        .maybe_par_bridge()
        .map(|sequence| {
            let words = process(sequence.as_ref())?;
            let mut map = HashMap::new();
            for word in words {
                map.entry(word).and_modify(|c| *c += 1).or_insert(1);
            }
            Ok(map)
        })
        .reduce(
            || Ok(HashMap::new()),
            |acc, ws| {
                let mut acc = acc?;
                for (k, v) in ws? {
                    acc.entry(k).and_modify(|c| *c += v).or_insert(v);
                }
                Ok(acc)
            },
        );

    self.words = words?;
    Ok(())
}

// On the Tokenizer, the default train now deals with iterators directly
pub fn train<T, I, S>(&mut self, trainer: &mut T, sequences: I) -> Result<&mut Self>
where
    T: Trainer<Model = M> + Sync,
    I: Iterator<Item = S> + Send,
    S: AsRef<str> + Send,
{
    let (lower, upper) = sequences.size_hint();
    let len = upper.unwrap_or(lower) as u64;
    let progress = if trainer.should_show_progress() {
        let progress = ProgressBar::new(len);
        progress.set_style(
            ProgressStyle::default_bar()
                .template("[{elapsed_precise}] {msg:<40!} {wide_bar} {pos:<9!}/{len:>9!}"),
        );
        progress.set_message("Pre-processing sequences");
        progress.set_draw_delta(len / 100); // Redraw only every 2%
        Some(progress)
    } else {
        None
    };

    trainer.feed(
        sequences.map(|s| {
            if let Some(progress) = &progress {
                progress.inc(1)
            }
            s
        }),
        |seq| {
            let normalized = self.do_normalize(seq.as_ref())?;
            let pre_tokenized = self.do_pre_tokenize(normalized)?;
            Ok(pre_tokenized
                .get_splits(OffsetReferential::Original, OffsetType::Byte)
                .into_iter()
                .map(|(s, _, _)| s.to_owned())
                .collect())
        },
    )?;
    if let Some(pbar) = progress {
        pbar.finish();
    }

    // let words = self.word_count(trainer, files)?;
    let special_tokens = trainer.train(&mut self.model)?;
    self.add_special_tokens(&special_tokens);

    Ok(self)
}

// And we can have a version that trains on files
pub fn train_from_files<T>(&mut self, trainer: &mut T, files: Vec<String>) -> Result<&mut Self>
where
    T: Trainer<Model = M> + Sync,
{
    let max_read = 1_000_000;
    use crate::utils::iter::ResultShunt;
    ResultShunt::process(
        files.into_iter().flat_map(|filename| {
            match File::open(filename) {
                Ok(file) => {
                    let file = BufReader::with_capacity(max_read, file);
                    // We read new lines using this API instead of the Lines Iterator
                    // on purpose. We want to keep the `\n` and potential `\r` between each lines
                    // We use an iterator to be able to chain with par_bridge.
                    itertools::Either::Left(file.lines_with_ending())
                }
                Err(e) => itertools::Either::Right(std::iter::once(Err(e))),
            }
        }),
        |iter| self.train(trainer, iter).map(|_| {}),
    )??;
    Ok(self)
}
```

Narsil:
> Nice! I've been thinking about your comments and experimented a bit, and I think I might have something that can avoid almost/all the negative points stated above, without complicating things too much and even simplifying some others.

I like your proposal, my main concerns are:

1. How feasible/simple it is to implement this `Send` Iterator.
2. Be careful about the accumulating costs of casting from native string to rust string for each sentence might incur a bigger cost that we want (to be checked)

Other than that it's great. And with the inversion of control proposed in #519 I think most of the problems will be gone.

> 
> > It enables shifting paradigm so that `feed_train_chunk` could start to
> > directly feed some inner struct of `Trainer` (pairs for BPE, sentences
> > for Unigram) and bypass `word_counts` later.
> 
> I'm not sure to understand exactly how you would feed types specific to each `Trainer` without being too specific about each of them in the `Tokenizer`. I think it should stay the responsibility of each `Trainer` to process the data, and as such, having them process a bunch of `&str` seems the most straightforward way to do it for me.

It is the `tokenizer` responsability to process the data (pre_tokenize and normalize). But yes it's the trainer's responsability to store the struct it needs to do that actual training, and that what I was referring to with that sentence.

> 
> > Puts burden on the caller to feed chunks
> > large enough that parallelization actually achieves something.
> 
> I think the way to avoid this is to actually parallelize over the iterator instead of each element. We could absolutely be training over an array with small sentences, and I even feel like it will be the case most of the time. Plus, we can expect to get at least as many elements in the iterator as the number of threads in the pool, while we can't really expect each string to be long enough and to include some `\n`.

I don't think we can have a python iterator to be `Send`, it's stateful by nature. 
The problem with that solution is that for every language we need to find a way to make a natural iterator become a `Send` iterator in Rust. That's why I thought moving large String chunks around felt more natural (We don't have to create that `Send` structure for each langage). We also have to take into account the cost of Casting a binding language string into a Rust string, the more we have to do that slower it gets (we probably need to check if this is a real bottleneck or not).


> 
> > * Adds a new item in `Tokenizer` struct that does not really belong (it
> >   belongs to the trainer and could be different from words, but this state
> >   is temporary).
> > * In the case of `datasets` we are still having a slowdown because we
> >   have to recreate a `string` chunk for a list of strings that gets
> >   splitted later.
> 
> These are not necessary with the solution below. `datasets` can probably directly feed a NumPy array, without the need for any sort of pre-processing.
> 
> > `{train|train_and_replace}_{from_files|from_buffer}` is starting to get a bit silly.
> 
> With #519, we remove `train_and_replace` to keep only `train`, that will be training the attached `Model` in-place. Makes everything easier.

Yup that's nice.

> 
> So in order to achieve this, I think we can have something along the lines of:
> 
> ```rust
> pub trait Trainer {
>     [...]
> 
>     // No need to give anything else than the Model to train
>     fn train(&self, model: &mut Self::Model) -> Result<Vec<AddedToken>>;
> 
>     // We feed the `Trainer` with an iterator, and give him a way to pre-process each sequence
>     fn feed<I, S, F>(&mut self, iterator: I, process: F) -> Result<()>
>     where
>         I: Iterator<Item = S> + Send,
>         S: AsRef<str> + Send,
>         F: Fn(&str) -> Result<Vec<String>> + Sync;
> }
> 
> // Example of impl to handle word counts. We can factorize the inside to easily
> // re-use it in the various Trainer for now.
> fn feed<I, S, F>(&mut self, iterator: I, process: F) -> Result<()>
> where
>     I: Iterator<Item = S> + Send,
>     S: AsRef<str> + Send,
>     F: Fn(&str) -> Result<Vec<String>> + Sync,
> {
>     let words: Result<HashMap<String, u32>> = iterator
>         .maybe_par_bridge()
>         .map(|sequence| {
>             let words = process(sequence.as_ref())?;
>             let mut map = HashMap::new();
>             for word in words {
>                 map.entry(word).and_modify(|c| *c += 1).or_insert(1);
>             }
>             Ok(map)
>         })
>         .reduce(
>             || Ok(HashMap::new()),
>             |acc, ws| {
>                 let mut acc = acc?;
>                 for (k, v) in ws? {
>                     acc.entry(k).and_modify(|c| *c += v).or_insert(v);
>                 }
>                 Ok(acc)
>             },
>         );
> 
>     self.words = words?;
>     Ok(())
> }
> 
> // On the Tokenizer, the default train now deals with iterators directly
> pub fn train<T, I, S>(&mut self, trainer: &mut T, sequences: I) -> Result<&mut Self>
> where
>     T: Trainer<Model = M> + Sync,
>     I: Iterator<Item = S> + Send,
>     S: AsRef<str> + Send,
> {
>     let (lower, upper) = sequences.size_hint();
>     let len = upper.unwrap_or(lower) as u64;
>     let progress = if trainer.should_show_progress() {
>         let progress = ProgressBar::new(len);
>         progress.set_style(
>             ProgressStyle::default_bar()
>                 .template("[{elapsed_precise}] {msg:<40!} {wide_bar} {pos:<9!}/{len:>9!}"),
>         );
>         progress.set_message("Pre-processing sequences");
>         progress.set_draw_delta(len / 100); // Redraw only every 2%
>         Some(progress)
>     } else {
>         None
>     };
> 
>     trainer.feed(
>         sequences.map(|s| {
>             if let Some(progress) = &progress {
>                 progress.inc(1)
>             }
>             s
>         }),
>         |seq| {
>             let normalized = self.do_normalize(seq.as_ref())?;
>             let pre_tokenized = self.do_pre_tokenize(normalized)?;
>             Ok(pre_tokenized
>                 .get_splits(OffsetReferential::Original, OffsetType::Byte)
>                 .into_iter()
>                 .map(|(s, _, _)| s.to_owned())
>                 .collect())
>         },
>     )?;
>     if let Some(pbar) = progress {
>         pbar.finish();
>     }
> 
>     // let words = self.word_count(trainer, files)?;
>     let special_tokens = trainer.train(&mut self.model)?;
>     self.add_special_tokens(&special_tokens);
> 
>     Ok(self)
> }
> 
> // And we can have a version that trains on files
> pub fn train_from_files<T>(&mut self, trainer: &mut T, files: Vec<String>) -> Result<&mut Self>
> where
>     T: Trainer<Model = M> + Sync,
> {
>     let max_read = 1_000_000;
>     use crate::utils::iter::ResultShunt;
>     ResultShunt::process(
>         files.into_iter().flat_map(|filename| {
>             match File::open(filename) {
>                 Ok(file) => {
>                     let file = BufReader::with_capacity(max_read, file);
>                     // We read new lines using this API instead of the Lines Iterator
>                     // on purpose. We want to keep the `\n` and potential `\r` between each lines
>                     // We use an iterator to be able to chain with par_bridge.
>                     itertools::Either::Left(file.lines_with_ending())
>                 }
>                 Err(e) => itertools::Either::Right(std::iter::once(Err(e))),
>             }
>         }),
>         |iter| self.train(trainer, iter).map(|_| {}),
>     )??;
>     Ok(self)
> }
> ```

n1t0:
> I don't think we can have a python iterator to be Send, it's stateful by nature.
The problem with that solution is that for every language we need to find a way to make a natural iterator become a Send iterator in Rust. That's why I thought moving large String chunks around felt more natural (We don't have to create that Send structure for each langage). We also have to take into account the cost of Casting a binding language string into a Rust string, the more we have to do that slower it gets (we probably need to check if this is a real bottleneck or not).

We can convert and stream everything in Python by doing something like this:
```rust
#[args(trainer = "None")]
fn train_from_iterator(
    &mut self,
    trainer: Option<&mut PyTrainer>,
    iterator: &PyAny,
) -> PyResult<()> {
    let mut trainer =
        trainer.map_or_else(|| self.tokenizer.get_model().get_trainer(), |t| t.clone());
    let (send, recv) = std::sync::mpsc::sync_channel(256);
    let mut sender = Some(send);
    let iterator: PyIterator = iterator.iter()?;

    crossbeam::thread::scope(|s| {
        let _train_handle = s.spawn(|_| {
            self.tokenizer
                .train(&mut trainer, recv.into_iter())
                .map(|_| {})
        });

        ResultShunt::process(iterator.map(|seq| seq?.extract::<&str>()), |iter| {
            if let Some(send) = sender.take() {
                for seq in iter {
                    send.send(seq)
                        .map_err(|e| exceptions::PyException::new_err(e.to_string()))?;
                }
            }
            Ok(())
        })?
    })
    .unwrap()
}
```

The only problem I see with this is that it might get tricky to make it work with custom Python components, but It does not seem impossible though. It would incur some bad GIL locking contention but that's the case wherever we use custom Python components at the moment, and something we can improve in many places.

marcglobality:
I'm very interested in this feature. What's the state of this? :)

n1t0:
Superseded by #544

PR #513, opened 2020-11-09
Rename words -> word_ids & sequences -> sequence_ids

n1t0:
`words` and `sequences` are misleading because they actually represent indices and not actual words/sequences.


thomwolf:
You don't need to also update the `pyi`?
https://github.com/huggingface/tokenizers/blob/master/bindings/python/py_src/tokenizers/__init__.pyi#L314

n1t0:
I didn't do it, to avoid a conflict in #509. It will get updated automatically then.

Issue #504, opened 2020-11-04
Custom tokenizer not returning UNK on tokens outside the vocabulary


talolard:
I was fooling around with a custom tokenizer and when I passed it text that wasn't in the vocabulary it simply ignored it :-( 
```python
    from tokenizers.trainers import BpeTrainer
    from tokenizers import Tokenizer
    from tokenizers.models import BPE
    import tokenizers

    tokenizer = Tokenizer(BPE())
    tokenizer.pre_tokenizer = tokenizers.pre_tokenizers.Whitespace()

    trainer = BpeTrainer(special_tokens=["[UNK]", "[CLS]", "[SEP]", "[PAD]", "[MASK]"])
    tokenizer.train(trainer, ["path to any file without hebrew"])
    text = "i am tal  and I wrote this שלום לברים"
    encoding = tokenizer.encode(text)
    shalom_index = text.find('שלום')
    assert shalom_index>0
    assert encoding.char_to_token(shalom_index) is not None, "The tokenizer wont say Shalom"
```
![image](https://user-images.githubusercontent.com/5352830/98120071-cb110d80-1ead-11eb-92c3-48f38b49db12.png)

And 
```python
print(encoding.tokens) 
```
Gives back 
>['i', 'am', 'tal', 'and', 'I', 'w', 'ro', 'te', 'this']

What am I doing wrong?

talolard:
I found the answer in the docs
![image](https://user-images.githubusercontent.com/5352830/98157038-07a62e80-1ed9-11eb-86d5-625af9c5e330.png)

The README is inconsistent, it doesn't say to save the model and for UNK tokens to work. 
![image](https://user-images.githubusercontent.com/5352830/98157105-1f7db280-1ed9-11eb-8154-c8ea9f324b93.png)

Any objections to me adding that to the readme ?

PR #518, opened 2020-11-10
Adding ways to flow `unk_token` from trainer to new model.

### LINKED ISSUE #517: Unknown token missing from BPE tokenizer trained with BPETrainer
I took a look at the short tutorial (https://huggingface.co/transformers/tokenizer_summary.html) and tried to replicate it (the BPE part of it).

My initial corpus looks like this: 

```
hug
hug
hug
hug
hug
hug
hug
hug
hug
hug
pug
pug
pug
pug
pug
pun
pun
pun
pun
pun
pun
pun
pun
pun
pun
pun
pun
bun
bun
bun
bun
hugs
hugs
hugs
hugs
hugs
```

Which is same as `('hug', 10), ('pug', 5), ('pun', 12), ('bun', 4), ('hugs', 5)` as presented in the tutorial.

Then I train the BPE tokenizer as follows:

```

tokenizer = Tokenizer(BPE())
tokenizer.pre_tokenizer = Whitespace()
from tokenizers.trainers import BpeTrainer

trainer = BpeTrainer(vocab_size=10)
tokenizer.train(trainer, ["test.txt"])
```

and test it on words given in the example: bug and mug

```
In [12]: tokenizer.encode("bug").tokens
Out[12]: ['b', 'ug']

In [13]: tokenizer.encode("mug").tokens
Out[13]: ['ug']
```

Shouldn't it be `['UNK', 'ug']` ?

Another example:

```
In [17]: tokenizer.encode("mug 😀").tokens
Out[17]: ['ug']
```

Is this a bug?

Narsil:
Fixes #517

### DIFF
--- a/bindings/node/lib/bindings/tokenizer.test.ts
+++ b/bindings/node/lib/bindings/tokenizer.test.ts
@@ -132,7 +132,7 @@ describe("Tokenizer", () => {
 
       const model = BPE.empty();
       tokenizer = new Tokenizer(model);
-      tokenizer.preTokenizer = whitespacePreTokenizer();
+      tokenizer.setPreTokenizer(whitespacePreTokenizer());
       tokenizer.addTokens(["my", "name", "is", "john", new AddedToken("pair", false)]);
 
       encode = promisify(tokenizer.encode.bind(tokenizer));
@@ -384,7 +384,7 @@ describe("Tokenizer", () => {
     beforeAll(() => {
       const model = BPE.empty();
       tokenizer = new Tokenizer(model);
-      tokenizer.preTokenizer = whitespacePreTokenizer();
+      tokenizer.setPreTokenizer(whitespacePreTokenizer());
       tokenizer.addTokens(["my", "name", "is", "john", "pair"]);
 
       encode = promisify(tokenizer.encode.bind(tokenizer));
--- a/bindings/node/lib/bindings/tokenizer.test.ts
+++ b/bindings/node/lib/bindings/tokenizer.test.ts
@@ -5,6 +5,7 @@ import { promisify } from "util";
 
 import { PaddingDirection, TruncationStrategy } from "./enums";
 import { BPE } from "./models";
+import { whitespacePreTokenizer } from "./pre-tokenizers";
 import { RawEncoding } from "./raw-encoding";
 import {
   AddedToken,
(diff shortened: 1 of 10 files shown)


bindings/node/lib/bindings/tokenizer.test.ts
```diff
@@ -131,6 +132,7 @@ describe("Tokenizer", () => {
 
       const model = BPE.empty();
       tokenizer = new Tokenizer(model);
+      tokenizer.setPreTokenizer(whitespacePreTokenizer());
```
n1t0, line 135: I don't understand why you want to add these in all the tests. The pre-tokenizer won't get called anyway because all the tokens in the tests come are handled by the `AddedVocabulary`.

tokenizers/src/models/bpe/trainer.rs
```diff
@@ -43,6 +43,7 @@ struct Config {
     initial_alphabet: HashSet<char>,
     continuing_subword_prefix: Option<String>,
     end_of_word_suffix: Option<String>,
+    unk_token: Option<String>,
```
n1t0, line 46: I think it doesn't make sense to add the `unk_token` to the trainer. I thought about it too, but since it will never get used, this would just be misleading for the user and actually requires more work on their part. I'm fixing this with #519.

n1t0 (changes requested):
Thank you for taking a look at this!

I think we should just keep the modifications in `tokenizers/src/models/bpe/mod.rs` and `tokenizers/src/models/bpe/model.rs` that fix the unexpected behavior when no `unk_token` is set. This also fixes #504

bindings/node/lib/bindings/tokenizer.test.ts
Narsil: That's not True, at least it was not in my case, all the tests were failing because the test string was not splitted through whitespace, so the added vocabulary was not handled, and those tests were failing because UNK_TOKEN was not defined.

We probably should really test the output of those methods in the full to make sure it's consistent, I'm pretty sure all those tests would have caught the missing unk tokens otherwise.

n1t0: The added vocabulary is always handled since it happens first, so no need for a pre-tokenizer. The tests actually started to fail because the BPE is receiving the whitespace between each word, and it doesn't have an `unk_token`. With the `Whitespace` pre-tokenizer, these whitespaces are just removed.

> We probably should really test the output of those methods in the full to make sure it's consistent, I'm pretty sure all those tests would have caught the missing unk tokens otherwise.

The unk token was treated this way by design. When provided we use it, otherwise the unknown tokens just get ignored. This is not a bug we didn't catch. That being said, feel free to add all the tests you find necessary, we never have enough of them :slightly_smiling_face:

PR #519, opened 2020-11-10
Trainer improvements

### LINKED ISSUE #201: BPE dropout not working as expected
I'm using `BytelvelBPETokenizer` together with `fastai.text` to make a text classifier in Python, and experimenting with BPE dropout.

When I create an empty model and train on my own text corpus, BPE dropout doesn't seem to work.

```python
import json
import pathlib

import tokenizers


# Make an empty tokeniser then train on some text

with open('train.txt', 'w') as f:
    f.writelines(line + '\n' for line in ['hi, how are you?', 'fine, how about you?', 'ababab'])

bpe_tokeniser_from_empty = tokenizers.ByteLevelBPETokenizer(dropout=0.5)
bpe_tokeniser_from_empty.train('train.txt', vocab_size=500)

print('dropout =', bpe_tokeniser_from_empty._parameters['dropout'])
for _ in range(20):
    print(bpe_tokeniser_from_empty.encode('ab').tokens)
```

output:
```
dropout = 0.5
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
['ab']
```

However, when I create a `ByteLevelBPETokenizer` from vocab and merges files, BPE dropout does work as I expect (probabilistic merging). Did I get something wrong with the first method, or is this a bug?

```python
# Create dummy vocab and merges files to build tokeniser from

with open('vocab.json', 'w') as f:
    json.dump({'a': 0, 'b': 1, 'ab': 2}, f)

with open('merges.txt', 'w') as f:
    f.write('#version 1\na b')

bpe_tokeniser_from_files = tokenizers.ByteLevelBPETokenizer(
    vocab_file='vocab.json',
    merges_file='merges.txt',
    dropout=0.5,
)

print('dropout =', bpe_tokeniser_from_files._parameters['dropout'])
for _ in range(20):
    print(bpe_tokeniser_from_files.encode('ab').tokens)
```

output:
```
dropout = 0.5
['a', 'b']
['ab']
['ab']
['ab']
['a', 'b']
['ab']
['a', 'b']
['a', 'b']
['ab']
['a', 'b']
['a', 'b']
['ab']
['ab']
['ab']
['ab']
['a', 'b']
['a', 'b']
['ab']
['a', 'b']
['ab']
```

I came up with this workaround: train on the source text, then output the merges and vocab files and use them to make a new tokeniser

```python
# My workaround

# Train a tokeniser on my data (BPE dropout will not work for this one)
bpe_tokeniser_from_empty = tokenizers.ByteLevelBPETokenizer(dropout=0.5)
bpe_tokeniser_from_empty.train('train.txt', vocab_size=500)

# Export the files
vocab_file, merges_file = bpe_tokeniser_from_empty.save('.')

# Create from files
bpe_tokeniser_from_files = tokenizers.ByteLevelBPETokenizer(
    vocab_file=pathlib.Path(vocab_file).name,
    merges_file=pathlib.Path(merges_file).name,
    dropout=0.5,
)

print('dropout =', bpe_tokeniser_from_files._parameters['dropout'])
for _ in range(20):
    print(bpe_tokeniser_from_files.encode('ab').tokens)
```

output:
```
dropout = 0.5
['a', 'b']
['a', 'b']
['a', 'b']
['ab']
['ab']
['a', 'b']
['a', 'b']
['a', 'b']
['a', 'b']
['a', 'b']
['ab']
['ab']
['ab']
['a', 'b']
['a', 'b']
['ab']
['a', 'b']
['ab']
['a', 'b']
['ab']
```

### LINKED ISSUE #265: Add the WordLevelTrainer
Currently, it is not possible to train a `WordLevel` model.

### LINKED ISSUE #526: Trainer trains the Model in-place
### Current state

When training a `Tokenizer`, its `Model` gets replaced after training since the `Trainer` generates a new `Model`. This has several limitations:
- When the `Model` being replaced has been customized (dropout, unk_token, ...), we lose all of this when we replace it (cf #201)
- In Python, if we keep a reference to the model added to the `Tokenizer`, this reference does not point to the actual model used by the `Tokenizer` after training.

### Goal

Change the `Trainer` to actually train to `Model` in-place.

### LINKED ISSUE #527: Make the `trainer` optional when training
### Current state

When training a `Tokenizer`, we must provide a `Trainer`:
```python
trainer = BpeTrainer(...)
tokenizer.train(files=[ ... ], trainer=trainer)
```
This is true even if we instantiate the `BpeTrainer` without changing any defaults.

### Goal

Makes the `Trainer` optional:
```python
tokenizer.train(files=[ ... ])
```

### How

The `Model` should be able to return a default trainer:
```python
trainer = model.get_trainer()
```

n1t0:
- Add a `WordLevelTrainer` (Fix #265)
- Add a default impl for `Trainer::process_tokens`
- Each `Model` can now return its associated trainer with `get_trainer`. This removes the need to provide a trainer in Python (could be optional on Rust too) (Fix #527)
- The `Model` on the `Tokenizer` now gets trained in-place. This allows keeping its customization (dropout, unk_token, ...) while just training it. That's also more in line with the role of the trainer, which is to train a Model, not to build one. (Fix #201 & Fix #526)
- This makes us wrap the `Model` in a `RwLock` on the Python's side. This is something that we'll have to do anyway very soon to provide the ability to customize the components properties in Python.

Missing before merge:
 - [x] Node bindings
 - [x] Update parts of the docs where we reload the `Model` after training
 - [x] Update parts of the docs where `Tokenizer.train` is used

### DIFF
--- a/bindings/python/py_src/tokenizers/implementations/bert_wordpiece.py
+++ b/bindings/python/py_src/tokenizers/implementations/bert_wordpiece.py
@@ -115,4 +115,4 @@ class BertWordPieceTokenizer(BaseTokenizer):
         )
         if isinstance(files, str):
             files = [files]
-        self._tokenizer.train(trainer, files)
+        self._tokenizer.train(files, trainer=trainer)
--- a/bindings/python/src/models.rs
+++ b/bindings/python/src/models.rs
@@ -581,12 +581,12 @@ mod test {
         match *py_model.model.as_ref().read().unwrap() {
             ModelWrapper::BPE(_) => (),
             _ => panic!("Expected Bert postprocessor."),
-        }
+        };
 
         let py_model: PyModel = serde_json::from_str(&rs_wrapper_ser).unwrap();
         match *py_model.model.as_ref().read().unwrap() {
             ModelWrapper::BPE(_) => (),
             _ => panic!("Expected Bert postprocessor."),
-        }
+        };
     }
 }
--- a/bindings/python/src/models.rs
+++ b/bindings/python/src/models.rs
@@ -3,6 +3,7 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use crate::token::PyToken;
+use crate::trainers::PyTrainer;
 use pyo3::exceptions;
 use pyo3::prelude::*;
 use pyo3::types::*;
@@ -45,6 +46,8 @@ impl PyModel {
 }
 
 impl Model for PyModel {
+    type Trainer = PyTrainer;
+
     fn tokenize(&self, tokens: &str) -> tk::Result<Vec<Token>> {
         self.model.tokenize(tokens)
     }
@@ -67,6 +70,10 @@ impl Model for PyModel {
 
     fn save(&self, folder: &Path, name: Option<&str>) -> tk::Result<Vec<PathBuf>> {
         self.model.save(folder, name)
+    }
+
+    fn get_trainer(&self) -> Self::Trainer {
+        self.model.get_trainer().into()
     }
 }
 
@@ -141,6 +148,10 @@ impl PyModel {
             .into_iter()
             .map(|path| path.to_string_lossy().into_owned())
             .collect())
+    }
+
+    fn get_trainer(&self) -> PyResult<PyObject> {
+        PyTrainer::from(self.model.get_trainer()).get_as_subtype()
     }
 }
 
--- a/bindings/python/src/models.rs
+++ b/bindings/python/src/models.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 
 use crate::token::PyToken;
 use crate::trainers::PyTrainer;
@@ -24,11 +24,11 @@ use super::error::{deprecation_warning, ToPyResult};
 #[derive(Clone, Serialize, Deserialize)]
 pub struct PyModel {
     #[serde(flatten)]
-    pub model: Arc<ModelWrapper>,
+    pub model: Arc<RwLock<ModelWrapper>>,
 }
 
 impl PyModel {
-    pub(crate) fn new(model: Arc<ModelWrapper>) -> Self {
+    pub(crate) fn new(model: Arc<RwLock<ModelWrapper>>) -> Self {
         PyModel { model }
     }
 
@@ -36,7 +36,7 @@ impl PyModel {
         let base = self.clone();
         let gil = Python::acquire_gil();
         let py = gil.python();
-        Ok(match self.model.as_ref() {
+        Ok(match *self.model.as_ref().read().unwrap() {
             ModelWrapper::BPE(_) => Py::new(py, (PyBPE {}, base))?.into_py(py),
             ModelWrapper::WordPiece(_) => Py::new(py, (PyWordPiece {}, base))?.into_py(py),
             ModelWrapper::WordLevel(_) => Py::new(py, (PyWordLevel {}, base))?.into_py(py),
@@ -49,31 +49,31 @@ impl Model for PyModel {
     type Trainer = PyTrainer;
 
     fn tokenize(&self, tokens: &str) -> tk::Result<Vec<Token>> {
-        self.model.tokenize(tokens)
+        self.model.read().unwrap().tokenize(tokens)
     }
 
     fn token_to_id(&self, token: &str) -> Option<u32> {
-        self.model.token_to_id(token)
+        self.model.read().unwrap().token_to_id(token)
     }
 
-    fn id_to_token(&self, id: u32) -> Option<&str> {
-        self.model.id_to_token(id)
+    fn id_to_token(&self, id: u32) -> Option<String> {
+        self.model.read().unwrap().id_to_token(id)
     }
 
-    fn get_vocab(&self) -> &HashMap<String, u32> {
-        self.model.get_vocab()
+    fn get_vocab(&self) -> HashMap<String, u32> {
+        self.model.read().unwrap().get_vocab()
     }
 
     fn get_vocab_size(&self) -> usize {
-        self.model.get_vocab_size()
+        self.model.read().unwrap().get_vocab_size()
     }
 
     fn save(&self, folder: &Path, name: Option<&str>) -> tk::Result<Vec<PathBuf>> {
-        self.model.save(folder, name)
+        self.model.read().unwrap().save(folder, name)
     }
 
     fn get_trainer(&self) -> Self::Trainer {
-        self.model.get_trainer().into()
+        self.model.read().unwrap().get_trainer().into()
     }
 }
 
@@ -84,7 +84,7 @@ impl PyModel {
         // Instantiate a default empty model. This doesn't really make sense, but we need
         // to be able to instantiate an empty model for pickle capabilities.
         Ok(PyModel {
-            model: Arc::new(BPE::default().into()),
+            model: Arc::new(RwLock::new(BPE::default().into())),
         })
     }
 
@@ -116,7 +116,7 @@ impl PyModel {
     /// Tokenize the given sequence
     #[text_signature = "(self, tokens)"]
     fn tokenize(&self, tokens: &str) -> PyResult<Vec<PyToken>> {
-        Ok(ToPyResult(self.model.tokenize(tokens))
+        Ok(ToPyResult(self.model.read().unwrap().tokenize(tokens))
             .into_py()?
             .into_iter()
             .map(|t| t.into())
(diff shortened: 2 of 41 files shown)


bindings/python/src/tokenizer.rs
```diff
@@ -1165,15 +1171,15 @@ mod test {
...
     use crate::normalizers::{PyNormalizer, PyNormalizerTypeWrapper};
-    use std::sync::Arc;
+    use std::sync::{Arc, RwLock};
     use tempfile::NamedTempFile;
     use tk::normalizers::{Lowercase, NFKC};
 
     #[test]
     fn serialize() {
-        let mut tokenizer = Tokenizer::new(PyModel::new(Arc::new(
+        let mut tokenizer = Tokenizer::new(PyModel::new(Arc::new(RwLock::new(
```
Narsil, line 1180: Could you explain why all those RwLock are need ? I'm not sure why we would need them.

bindings/python/py_src/tokenizers/implementations/bert_wordpiece.py
```diff
@@ -115,4 +115,4 @@ def train(
         )
         if isinstance(files, str):
             files = [files]
-        self._tokenizer.train(trainer, files)
+        self._tokenizer.train(files, trainer=trainer)
```
Narsil, line 118: As long as we're breaking signature, I would argue we have a different signature like

`train(files, options=bpe_train_options)`, or `train(files, vocab_size=X, ....)` what do you think ?

I like the second version better, the only trouble is the exact description of those options if going to get fuzzy pretty fast and error handling a bit hard. But it "feels" more pythonic, what do you think ?

Either that, or if we keep the `trainer` concept, we should stick to something closer to Rust, with `trainer.train(tokenizer, files)` I actually like that last version better at this moment, the control flow feels more natural.

bindings/python/src/models.rs
```diff
@@ -45,28 +46,34 @@ impl PyModel {
 }
 
 impl Model for PyModel {
+    type Trainer = PyTrainer;
+
     fn tokenize(&self, tokens: &str) -> tk::Result<Vec<Token>> {
-        self.model.tokenize(tokens)
+        self.model.read().unwrap().tokenize(tokens)
```
Narsil, line 52: I'm always a bit scared by adding that much `unwrap` everywhere... Do you think there's a way we could avoid them ?

Narsil (commented):
I really like the inversion of control !
Mostly seems fine, but we should probably have a discussion between available syntax for training if we're changing it.

Options:

1. `tokenizer.train(trainer, files)` (current)
2. `tokenizer.train(files, trainer=trainer)` `tokenizer.train(files)` (proposed)
3. `tokenizer.train(files, vocab_size=X, ...)`
4. `trainer.train(tokenizer, files)`

I like 2. because it can drop completely the concept of `trainer` in the default case, but I think it makes the case were you need it a bit awkward, 3. Fixes that, but then knowing the exact list of options is going to be tricky (as it depends on `tokenizer.model` now)

Option 4. does not remove the concept of `trainer` and the inversion of flow makes it closer to the rust version. But it does feel as simple as it could be.

Right now, I'm leaning on `4.` that would get called automatically by a syntax like `2.` in the default case, and we would drop the non default case. So:

`tokenizer.train(files)` works, and it equivalent to `trainer = tokenizer.default_trainer(); trainer.train(tokenizer, files)`.

bindings/python/src/tokenizer.rs
n1t0: Sure, we need them because anything in an `Arc` is immutable. These `RwLock` provides us with a way to actually mutate the `Model` here.

bindings/python/src/models.rs
n1t0: The `unwrap` here is for the `std::sync::LockResult` that is returned by the `RwLock` when you try to access its content. The `Err` case happens when a thread that was holding the lock panicked. So since this shouldn't happen, and we don't want to recover from it, I think the `unwrap` should be ok.

n1t0:
Sure! Happy to discuss the possible options.

With `2. tokenizer.train(files, trainer=trainer) & tokenizer.train(files)`, the expected usage is:
```python
# Default training, for example when you just want to train a pre-trained tokenizer on your own dataset
tokenizer = Tokenizer.from_pretrained("bert-base-uncased") # Not available as we speak, but soon
tokenizer.train(files=[ ... ])

# Same case but with some custom parameters
tokenizer = Tokenizer.from_pretrained("bert-base-uncased")
trainer = tokenizer.model.get_trainer()
print(trainer) # inspect the current training params
trainer.vocab_size = 32000
trainer.special_tokens = [ "[UNK]", "[PAD]" ]
tokenizer.train(files=[ ... ], trainer=trainer)

# Or building your own
trainer = BpeTrainer(...)
tokenizer.train(files=[ ... ], trainer)
```

I think we should avoid `3. tokenizer.train(files, vocab_size=X, ...)`. It looks good at first sight, but it would be pretty impossible to document accurately.

I'm not entirely sure about what you have in mind with `4. trainer.train(tokenizer, files)`. Can you provide some examples of the various usage in Python (and maybe the expected signatures in Rust too)?

Jess0-0:
Hi,

I'm using the WordLevelTrainer in this branch. I'm able to import the package and initialize a trainer. However, when I tried to use it in:
`trainer = WordLevelTrainer(special_tokens=["[UNK]", "[CLS]", "[SEP]", "[PAD]", "[MASK]"])`
`tokenizer.train(trainer, ["wiki.train.raw", "wiki.valid.raw", "wiki.test.raw"])`
it generates 

> TypeError: Can't convert <WordLevelTrainer object at 0x00000194EC6DBD90> to Sequence

error. Wondering if there is any way to fix this? Thanks!

n1t0:
In this PR we make the trainer optional when calling `train`, so you need to provide the files first:
```python
trainer = WordLevelTrainer(special_tokens=["[UNK]", "[CLS]", "[SEP]", "[PAD]", "[MASK]"])
tokenizer.train(["wiki.train.raw", "wiki.valid.raw", "wiki.test.raw"], trainer)
```

Narsil:
> Sure! Happy to discuss the possible options.
> 
> With `2. tokenizer.train(files, trainer=trainer) & tokenizer.train(files)`, the expected usage is:
> 
> ```python
> # Default training, for example when you just want to train a pre-trained tokenizer on your own dataset
> tokenizer = Tokenizer.from_pretrained("bert-base-uncased") # Not available as we speak, but soon
> tokenizer.train(files=[ ... ])
> 
> # Same case but with some custom parameters
> tokenizer = Tokenizer.from_pretrained("bert-base-uncased")
> trainer = tokenizer.model.get_trainer()
> print(trainer) # inspect the current training params
> trainer.vocab_size = 32000
> trainer.special_tokens = [ "[UNK]", "[PAD]" ]
> tokenizer.train(files=[ ... ], trainer=trainer)
> ```

I think this shows there's a problem in the naming of stuff. This example is using `trainer` as glorified `options` (or `params`). It's even mentionned in the comment. The user seem to be passing pure arguments, but the name `trainer` suggests that the object does something, namely it should `train`, which feels odd. That's the main reason why I like `trainer.train(tokenizer, files)` better, the trainer does train as he's supposed to.
The other way to think about it is, if we want to stick to `2.` we should rename everything to something more passive like options or params. That choice should also be reflected in the Rust API.

```python
tokenizer = Tokenizer.from_pretrained("bert-base-uncased")
train_options = tokenizer.model.get_train_options()
print(train_options) # it's actually a raw Python dict
train_options["vocab_size"] = 32000
train_options["special_tokens"] = [ "[UNK]", "[PAD]" ]
tokenizer.train(files=[ ... ], train_options=train_options)
# or 
tokenizer.train(files=[...], **train_options)
```

Small nit:
```python
train_options = tokenizer.model.get_train_options()
tokenizer.train(..., train_options)
``` 
feels strange, why not
```python
train_options = tokenizer.get_train_options() # No need to call the underlying model.
tokenizer.train(..., train_options)
``` 

**The main issue I have with that API is for training a tokenizer from scratch:**
```python
tokenizer = Tokenizer(BPE())
train_options = tokenizer.get_train_options()
```
should not work as there's no way this code can know what vocabulary size I want. This code should raise an Exception IMO saying the user should define the vocab_size.
I'm unsure what's the best way to fix it.
- `Tokenizer(BPE(vocab_size=X))` seems the most reasonable. You should get an error while trying to encode which would be the case anyway (unk token is not defined).
- `tokenizer.get_train_options(vocab_size=XX)` feels too odd (why it needs to be defined sometimes and not others and so on.)






BpeTrainer with Unigram model does not make sense, we should avoid at all cost enabling the mixup.
We also probably don't want to make choices in the user stead where it's not necessary (for instance vocab_size is a core argument, we should never choose one by default if it's not already available)

n1t0:
> I think this shows there's a problem in the naming of stuff. This example is using trainer as glorified options (or params). It's even mentionned in the comment. The user seem to be passing pure arguments, but the name trainer suggests that the object does something, namely it should train, which feels odd. That's the main reason why I like trainer.train(tokenizer, files) better, the trainer does train as he's supposed to.
The other way to think about it is, if we want to stick to 2. we should rename everything to something more passive like options or params. That choice should also be reflected in the Rust API.

I think you might have the wrong idea about the way things currently work. The `trainer` actually trains, it trains a `Model`, and it is not some sort of glorified params. We didn't expose its methods yet, because we are still changing them and it is not the main path so there's no rush on this side, but we will. It could actually even be used to train multiple `Model` while only ingesting the data once later on. All of this wouldn't be possible with a `dict`.
Also, I don't think the `Trainer` should somehow have a dependency on the `Tokenizer`. It trains a `Model` given some input text, and it should stay that way. The fact that the `Tokenizer` does pre-process this text before feeding it to the `Trainer` is and should stay the responsibility of the `Tokenizer`.

> BpeTrainer with Unigram model does not make sense, we should avoid at all cost enabling the mixup.

Agreed, and we don't allow it. It makes me realize that I treated this path as `unreachable` though, and will need to change it to raise instead.

PR #542, opened 2020-11-23
Split Pre-Tokenizer

### LINKED ISSUE #468: Add a `Split` PreTokenizer
We are missing a very versatile `PreTokenizer` that can cover the capabilities of `NormalizedString::split`. I propose to add a `Split` Pretokenizer that would have the following parameters:
  - A pattern: either a string or a `tokenizers.Regex`
  - A split behavior: using the `SplitDelimiterBehavior`
  - Whether to invert the behavior of the pattern (cf [here](https://github.com/huggingface/tokenizers/blob/6899618/tokenizers/src/tokenizer/pattern.rs#L124-L137))

patrickvonplaten:
Fix #468 

This is very much work in progress and a first version on how a `Split` pre-tokenizer can be implemented.

I'm still very inexperienced in Rust and accordingly, this first PR is a very unclean first version of how I think the pre tokenizer could work.

I tested the `Split` tokenizer by running:

```python 
use tokenizers::{OffsetReferential, OffsetType, PreTokenizedString, PreTokenizer, Result, SplitDelimiterBehavior};
use regex::Regex;
use tokenizers::pre_tokenizers::split::Split;

fn main() -> Result<()> {
    let pre_tokenizer = Split::new(Regex::new(r"\w+|[^\w\s]+").unwrap(), SplitDelimiterBehavior::Isolated, true);
    let mut pre_tokenized = PreTokenizedString::from("Hello! How are you? I'm fine, thank you.");

    pre_tokenizer.pre_tokenize(&mut pre_tokenized)?;

    let _result = pre_tokenized;

    println!(
        "{:?}",
        _result.get_splits(OffsetReferential::Original, OffsetType::Byte)
    );
    Ok(())
}
```

This code compiles and seems to work.

It would be awesome if I could get some help on the following points:
- First, is the general logic correct, *i.e.* is it correct that the `Split` struct should have the three fields `behavior` of type `SplitDelimiterBehavior`, `regex` of type `Regex` (or should that be changed to `String`?) and bool for `Invert` or not?
- I have some problems making the code clean for the `pre_tokenize` function => I don't think I need a match pattern here, but I didn't manage to simply do something like:

```python
pretokenized.split(|_, normalized| {
                        normalized.split(Invert(&self.re), self.behavior)
                                                                             ^
                                                                            # this throw move errors, I didn't manage to come around
})
```
similarly, I didn't manage to process the regex via `Invert`:

```python
let re: &Regex = if self.invert { Invert(&self.re) } { &self.re };
```

didn't work. 

- I'm not at all sure whether the serialization and deserialization is ok here 

@n1t0, @Narsil It would be awesome if you guys could give me some pointers/tips on how to tackle the issues above. I'm still reading the rust-lang book and following the rustlings course => sorry for these problem very basic questions

### DIFF
--- a/bindings/node/lib/bindings/pre-tokenizers.test.ts
+++ b/bindings/node/lib/bindings/pre-tokenizers.test.ts
@@ -47,7 +47,7 @@ describe("punctuationPreTokenizer", () => {
 
 describe("splitPreTokenizer", () => {
   it("instantiates correctly with invert parameter", () => {
-    const processor = splitPreTokenizer(" ", "isolated", false);
+    const processor = splitPreTokenizer(" ", "mergedWithPrevious", false);
     expect(processor.constructor.name).toEqual("PreTokenizer");
   });
 });
--- a/bindings/node/lib/bindings/pre-tokenizers.test.ts
+++ b/bindings/node/lib/bindings/pre-tokenizers.test.ts
@@ -3,6 +3,7 @@ import {
   metaspacePreTokenizer,
   punctuationPreTokenizer,
   sequencePreTokenizer,
+  splitPreTokenizer,
   whitespaceSplitPreTokenizer,
 } from "./pre-tokenizers";
 
@@ -40,6 +41,13 @@ describe("metaspacePreTokenizer", () => {
 describe("punctuationPreTokenizer", () => {
   it("instantiates correctly without any parameter", () => {
     const processor = punctuationPreTokenizer();
+    expect(processor.constructor.name).toEqual("PreTokenizer");
+  });
+});
+
+describe("splitPreTokenizer", () => {
+  it("instantiates correctly with invert parameter", () => {
+    const processor = splitPreTokenizer(" ", "isolated", false);
     expect(processor.constructor.name).toEqual("PreTokenizer");
   });
 });
--- a/bindings/node/native/src/pre_tokenizers.rs
+++ b/bindings/node/native/src/pre_tokenizers.rs
@@ -19,12 +19,12 @@ impl FromJsValue for JsSplitDelimiterBehavior {
         Ok(Self(match s.as_ref() {
             "removed" => Ok(SplitDelimiterBehavior::Removed),
             "isolated" => Ok(SplitDelimiterBehavior::Isolated),
-            "merged_with_previous" => Ok(SplitDelimiterBehavior::MergedWithPrevious),
-            "merged_with_next" => Ok(SplitDelimiterBehavior::MergedWithNext),
+            "mergedWithPrevious" => Ok(SplitDelimiterBehavior::MergedWithPrevious),
+            "mergedWithNext" => Ok(SplitDelimiterBehavior::MergedWithNext),
             "contiguous" => Ok(SplitDelimiterBehavior::Contiguous),
             _ => Err(Error(
                 "Wrong value for SplitDelimiterBehavior, expected one of: \
-                 `removed, isolated, merged_with_previous, merged_with_next, contiguous`"
+                 `removed, isolated, mergedWithPrevious, mergedWithNext, contiguous`"
                     .into(),
             )),
         }?))
@@ -188,7 +188,7 @@ fn metaspace(mut cx: FunctionContext) -> JsResult<JsPreTokenizer> {
 /// split(invert: bool = false)
 fn split(mut cx: FunctionContext) -> JsResult<JsPreTokenizer> {
     let pattern: String = cx.extract::<String>(0)?;
-    let behavior: JsSplitDelimiterBehavior = cx.extract::<JsSplitDelimiterBehavior>(1)?;
+    let behavior: SplitDelimiterBehavior = cx.extract::<JsSDB>(1)?.0;
     let invert: bool = cx.extract_opt::<bool>(2)?.unwrap_or(false);
 
     let mut pretok = JsPreTokenizer::new::<_, JsPreTokenizer, _>(&mut cx, vec![])?;
--- a/bindings/node/native/src/pre_tokenizers.rs
+++ b/bindings/node/native/src/pre_tokenizers.rs
@@ -188,7 +188,7 @@ fn metaspace(mut cx: FunctionContext) -> JsResult<JsPreTokenizer> {
 /// split(invert: bool = false)
 fn split(mut cx: FunctionContext) -> JsResult<JsPreTokenizer> {
     let pattern: String = cx.extract::<String>(0)?;
-    let behavior: SplitDelimiterBehavior = cx.extract::<JsSDB>(1)?.0;
+    let behavior: JsSplitDelimiterBehavior = cx.extract::<JsSplitDelimiterBehavior>(1)?;
     let invert: bool = cx.extract_opt::<bool>(2)?.unwrap_or(false);
 
     let mut pretok = JsPreTokenizer::new::<_, JsPreTokenizer, _>(&mut cx, vec![])?;
--- a/bindings/node/native/src/pre_tokenizers.rs
+++ b/bindings/node/native/src/pre_tokenizers.rs
@@ -7,6 +7,34 @@ use std::sync::Arc;
 use serde::{ser::SerializeStruct, Serialize, Serializer};
 use tk::pre_tokenizers::PreTokenizerWrapper;
 use tk::PreTokenizedString;
+use tk::normalizer::SplitDelimiterBehavior;
+
+#[derive(Clone)]
+struct JsSplitDelimiterBehavior(SplitDelimiterBehavior);
+
+impl FromJsValue for JsSplitDelimiterBehavior {
+    fn from_value<'c, C: Context<'c>>(from: Handle<'c, JsValue>, _cx: &mut C) -> LibResult<Self> {
+        let s = from.downcast::<JsString>()?.value();
+
+        Ok(Self(match s.as_ref() {
+            "removed" => Ok(SplitDelimiterBehavior::Removed),
+            "isolated" => Ok(SplitDelimiterBehavior::Isolated),
+            "merged_with_previous" => Ok(SplitDelimiterBehavior::MergedWithPrevious),
+            "merged_with_next" => Ok(SplitDelimiterBehavior::MergedWithNext),
+            "contiguous" => Ok(SplitDelimiterBehavior::Contiguous),
+            _ => Err(Error(
+                 "Wrong value for SplitDelimiterBehavior, expected one of: \
+                 `removed, isolated, merged_with_previous, merged_with_next, contiguous`".into()
+            )),
+        }?))
+    }
+}
+
+impl<'s> From<JsSplitDelimiterBehavior> for SplitDelimiterBehavior {
+    fn from(v: JsSplitDelimiterBehavior) -> Self {
+        v.0
+    }
+}
 
 #[derive(Clone, Debug, Deserialize)]
 #[serde(untagged)]
@@ -156,6 +184,22 @@ fn metaspace(mut cx: FunctionContext) -> JsResult<JsPreTokenizer> {
     Ok(pretok)
 }
 
+/// split(invert: bool = false)
+fn split(mut cx: FunctionContext) -> JsResult<JsPreTokenizer> {
+    let pattern: String = cx.extract::<String>(0)?;
+    let behavior: JsSplitDelimiterBehavior = cx.extract::<JsSplitDelimiterBehavior>(1)?;
+    let invert: bool = cx.extract_opt::<bool>(2)?.unwrap_or(false);
+
+    let mut pretok = JsPreTokenizer::new::<_, JsPreTokenizer, _>(&mut cx, vec![])?;
+    let guard = cx.lock();
+    pretok.borrow_mut(&guard).pretok = Some(
+        tk::pre_tokenizers::split::Split::new(pattern, behavior.into(), invert)
+            .map_err(|e| Error(e.to_string()))?
+            .into(),
+    );
+    Ok(pretok)
+}
+
 /// punctuation()
 fn punctuation(mut cx: FunctionContext) -> JsResult<JsPreTokenizer> {
     let mut pretok = JsPreTokenizer::new::<_, JsPreTokenizer, _>(&mut cx, vec![])?;
@@ -231,6 +275,7 @@ pub fn register(m: &mut ModuleContext, prefix: &str) -> NeonResult<()> {
     m.export_function(&format!("{}_WhitespaceSplit", prefix), whitespace_split)?;
     m.export_function(&format!("{}_BertPreTokenizer", prefix), bert_pre_tokenizer)?;
     m.export_function(&format!("{}_Metaspace", prefix), metaspace)?;
+    m.export_function(&format!("{}_Split", prefix), split)?;
     m.export_function(
         &format!("{}_CharDelimiterSplit", prefix),
         char_delimiter_split,
--- a/bindings/node/native/src/pre_tokenizers.rs
+++ b/bindings/node/native/src/pre_tokenizers.rs
@@ -5,9 +5,9 @@ use neon::prelude::*;
 use std::sync::Arc;
 
 use serde::{ser::SerializeStruct, Serialize, Serializer};
+use tk::normalizer::SplitDelimiterBehavior;
 use tk::pre_tokenizers::PreTokenizerWrapper;
 use tk::PreTokenizedString;
-use tk::normalizer::SplitDelimiterBehavior;
 
 #[derive(Clone)]
 struct JsSplitDelimiterBehavior(SplitDelimiterBehavior);
@@ -23,8 +23,9 @@ impl FromJsValue for JsSplitDelimiterBehavior {
             "merged_with_next" => Ok(SplitDelimiterBehavior::MergedWithNext),
             "contiguous" => Ok(SplitDelimiterBehavior::Contiguous),
             _ => Err(Error(
-                 "Wrong value for SplitDelimiterBehavior, expected one of: \
-                 `removed, isolated, merged_with_previous, merged_with_next, contiguous`".into()
+                "Wrong value for SplitDelimiterBehavior, expected one of: \
+                 `removed, isolated, merged_with_previous, merged_with_next, contiguous`"
+                    .into(),
             )),
         }?))
     }
--- a/bindings/python/py_src/tokenizers/pre_tokenizers/__init__.pyi
+++ b/bindings/python/py_src/tokenizers/pre_tokenizers/__init__.pyi
@@ -205,7 +205,7 @@ class Split(PreTokenizer):
             Whether to invert the pattern.
     """
 
-    def __init__(self, pattern, behavior, invert):
+    def __init__(self, pattern, behavior, invert=False):
         pass
     def pre_tokenize(self, pretok):
         """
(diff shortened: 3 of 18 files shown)


tokenizers/src/pre_tokenizers/split.rs
```diff
@@ -0,0 +1,145 @@
...
+impl<'de> Deserialize<'de> for Split {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserializer.deserialize_map(SplitVisitor)
+    }
+}
+
+struct SplitVisitor;
```
patrickvonplaten, line 125: I copied this functionality from the `Whitespace` pre-tokenizer. Not sure at all if this should be the same for `Split` to be honest.

tokenizers/src/pre_tokenizers/split.rs
```diff
@@ -0,0 +1,145 @@
...
+            invert
+        }
+    }
+}
+
+impl PreTokenizer for Split {
+    fn pre_tokenize(&self, pretokenized: &mut PreTokenizedString) -> Result<()> {
+        use SplitDelimiterBehavior::*;
+        match self.behavior {
+            Removed => {
```
patrickvonplaten, line 54: this is far from being clean. I encountered problems with passing the correct `SplitDelimiterBehavior` without having to do a match. 

```
pretokenized.split(|_, normalized| {
                        normalized.split(Invert(&self.re), self.behavior)
})
```

throws a "move" error. Would be super happy for some pointers to resources on how this can be better handled!!!

n1t0:
Awesome! Thank you @patrickvonplaten for looking at this. You clearly are in the right direction, and it shouldn't require too many changes. 

> It would be awesome if I could get some help on the following points:
> - First, is the general logic correct, *i.e.* is it correct that the `Split` struct should have the three fields `behavior` of type `SplitDelimiterBehavior`, `regex` of type `Regex` (or should that be changed to `String`?) and bool for `Invert` or not?

Yes it is! We can allow `Split` to be constructed with both raw strings and regex, while internally keeping a regex in all cases.

We'll want to use `onig::Regex` probably instead of `regex::Regex` because it allows using more RegExp features like negative lookahead for example, but that is a detail that should be kept inside `Split` so that we can change it as we see fit, without impacting the end-user. You can have a look at the `Replace` normalizer that actually works pretty similarly: https://github.com/huggingface/tokenizers/blob/master/tokenizers/src/normalizers/replace.rs

> - I have some problems making the code clean for the `pre_tokenize` function => I don't think I need a match pattern here, but I didn't manage to simply do something like:
> ```python
> pretokenized.split(|_, normalized| {
>                         normalized.split(Invert(&self.re), self.behavior)
>                                                                              ^
>                                                                             # this throw move errors, I didn't manage to come around
> })
> ```

Great! That's a good example to understand how the borrow checker works:
The `split` method on the `NormalizedString` expects an owned `SplitDelimiterBehavior`. So you can't lend it by doing:
```rust
normalized.split(Invert(&self.re), &self.behavior)
```
Now you can't give yours (the one on `Split`) either for two reasons:
 1. in `pre_tokenize`, we are taking self by shared reference `&self`, so we can't modify `self`.
 2. `Split` is defined with `behavior: SplitDelimiterBehavior` so it must have a `behavior` at all times. So even if we could modify `self`, we couldn't consume the `behavior` by giving it, otherwise, we wouldn't have one left for the next call to `pre_tokenize`.

So the only thing you can do here is actually to clone/copy your own `behavior` to hand it to `normalized.split`:
```rust
normalized.split(Invert(&self.re), self.behavior.clone())
```
But since `SplitDelimiterBehavior` is our own type and we can modify it, the best thing to do is to make it [`Copy`](https://doc.rust-lang.org/std/marker/trait.Copy.html) in `src/tokenizer/normalizer.rs`:
```rust
#[derive(Debug, Clone, Copy, PartialEq)]
pub enum SplitDelimiterBehavior {
    Removed,
    Isolated,
    MergedWithPrevious,
    MergedWithNext,
    Contiguous,
}
```
And then the code you showed above will work!

> I tested the `Split` tokenizer by running:
> ```python 
> use tokenizers::{OffsetReferential, OffsetType, PreTokenizedString, PreTokenizer, Result, SplitDelimiterBehavior};
> use regex::Regex;
> use tokenizers::pre_tokenizers::split::Split;
> 
> fn main() -> Result<()> {
>     let pre_tokenizer = Split::new(Regex::new(r"\w+|[^\w\s]+").unwrap(), SplitDelimiterBehavior::Isolated, true);
>     let mut pre_tokenized = PreTokenizedString::from("Hello! How are you? I'm fine, thank you.");
> 
>     pre_tokenizer.pre_tokenize(&mut pre_tokenized)?;
> 
>     let _result = pre_tokenized;
> 
>     println!(
>         "{:?}",
>         _result.get_splits(OffsetReferential::Original, OffsetType::Byte)
>     );
>     Ok(())
> }
> ```

Nice! I think you can even turn this into a test, directly in `split.rs`. You can check the tests we have for the `Replace` normalizer for an example.

> - I'm not at all sure whether the serialization and deserialization is ok here 

I think it should probably be very similar to how we handle it for `Replace` too.

tokenizers/src/pre_tokenizers/split.rs
```diff
@@ -0,0 +1,145 @@
...
+            re,
+            behavior,
+            invert
+        }
+    }
+}
+
+impl PreTokenizer for Split {
+    fn pre_tokenize(&self, pretokenized: &mut PreTokenizedString) -> Result<()> {
+        use SplitDelimiterBehavior::*;
```
Narsil, line 52: Usually imports are at the beginning of the file. It should always be at the top level module
Exception is tests within a file (where imports are still at the beginning of the module and not in functions)
Tests is usually a module at the end of a file.

Narsil (commented):
Nicely done ! Not easy for a first Rust PR :).

Let me know if there's anything unclear in my comments.

tokenizers/src/normalizers/replace.rs
```diff
@@ -21,7 +21,7 @@ impl From<&str> for ReplacePattern {
     }
 }
 
-/// We use this custom deserializer to provided the value for `regex` for `Replace`
+/// We use this custom deserializer to provide the value for `regex` for `Replace`
```
patrickvonplaten, line 24: just a typo

patrickvonplaten:
@n1t0 @Narsil - thanks a lot for your reviews! I'm learning a lot from it :-) 
I think the `tokenizers/src/pre_tokenizers/split.rs` is in a better state now! I will fix the CI tomorrow and will also add some documentation. I'll ping you guys again, then if that's fine.

tokenizers/src/pre_tokenizers/split.rs
n1t0: Actually, this is one of those things Rust handles very well. You can't make a type `Copy` if it contains anything that shouldn't be copied, like `Vec`, `HashMap`, ...

So, as the Rust doc says:
> **When should my type be Copy?**
Generally speaking, if your type can implement Copy, it should. Keep in mind, though, that implementing Copy is part of the public API of your type. If the type might become non-Copy in the future, it could be prudent to omit the Copy implementation now, to avoid a breaking API change.

(source [the doc for `std::marker::Copy`](https://doc.rust-lang.org/std/marker/trait.Copy.html))

tokenizers/src/pre_tokenizers/split.rs
n1t0: That's a good way to avoid polluting the global scope with enum variants though, and something you'll see very often in Rust. There are multiple places in the lib where we use it.

tokenizers/src/pre_tokenizers/split.rs
```diff
@@ -1,145 +1,231 @@
...
+                vec![
+                    ("How", (0, 3)),
+                    (" are", (3, 7)),
+                    (" you", (7, 11)),
+                    (" doing", (11, 17)),
+                    ("?", (17, 18)),
+                ],
+            ),
+            (
+                Contiguous,
```
Narsil, line 155: Just to make sure that Isolated and Contiguous are different because:

Isolated : "I  have this"  -> "I", " ", " ", "have", " ", "this"
Contiguous : "I  have this"  -> "I", "  ", "have", " ", "this"
(Not sure it's how it's going to behave that way, and even if it's the correct behavior with this regexp, but fyi)

tokenizers/src/pre_tokenizers/split.rs
```diff
@@ -1,145 +1,231 @@
...
+    fn eq(&self, other: &Split) -> bool {
+        self.pattern == other.pattern && self.behavior == other.behavior && self.invert == other.invert
     }
 }
 
 impl Split {
-    pub fn new(re: Regex, behavior: SplitDelimiterBehavior, invert: bool) -> Self {
-        Split {
-            re,
+    pub fn new<I: Into<SplitPattern>>(pattern: I, behavior: SplitDelimiterBehavior, invert: bool) -> Result<Self> {
```
Narsil, line 68: Solid!

Narsil (approved):
Super solid !

tokenizers/src/pre_tokenizers/split.rs
```diff
@@ -0,0 +1,145 @@
...
+// use crate::pattern::Pattern;
+use crate::tokenizer::{
+    pattern::{Invert, Pattern}, PreTokenizedString, PreTokenizer, Result, SplitDelimiterBehavior,
+};
+use serde::de::{Error, Visitor};
+
+#[derive(Clone, Debug, Serialize)]
+#[serde(tag = "type")]
+pub struct Split {
+    #[serde(default = "default_regex", skip)]
```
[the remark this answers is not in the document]
patrickvonplaten: Yes, I corrected this I think

patrickvonplaten:
## Update:

- **TODO**
- [ ] bindings/node
- [ ] docs
- [ ] final check

- Ready for Review:
- [ ] bindings/python
- [ ] tokenizers/src/

bindings/python/src/pre_tokenizers.rs
```diff
@@ -220,6 +222,48 @@ impl PyWhitespaceSplit {
...
+///
+///     behavior: SplitDelimiterBehavior:
+///         The behavior to use when splitting.
+///         Choices: "removed", "isolated", "merged_with_previous", "merged_with_next",
+///         "contiguous"
+///
+///     invert: bool:
+///         Whether to invert the pattern.
+#[pyclass(extends=PyPreTokenizer, module = "tokenizers.pre_tokenizers", name=Split)]
+#[text_signature = "(self, pattern, behavior, invert=False)"]
```
patrickvonplaten, line 243: not sure if this default signature is as you expected @n1t0 - I thought `pattern` and `behavior` is too general to give it default values, but `invert` could well default to `False`...

bindings/python/src/utils/normalization.rs
```diff
@@ -44,6 +44,15 @@ impl From<PyPattern<'_>> for tk::normalizers::replace::ReplacePattern {
     }
 }
 
+impl From<PyPattern<'_>> for tk::pre_tokenizers::split::SplitPattern {
```
patrickvonplaten, line 47: is it fine that `PyPattern` is implemented for `pre_tokenizers` in this file or should I move this to `utils/pretokenization.rs` ?

bindings/python/pyproject.toml
```diff
@@ -1,5 +1,6 @@
 [build-system]
 requires = ["setuptools", "wheel", "setuptools-rust"]
+build-backend = "setuptools.build_meta"
```
patrickvonplaten, line 3: @Narsil @n1t0 - I had to add this `build-system` dependency to make `pip install -e .` work for me (which in turn I needed to make `make style` work...).

If I don't add this `pip install -e .` fails with this error: https://github.com/pypa/setuptools/issues/1694

bindings/python/py_src/tokenizers/pre_tokenizers/__init__.pyi
```diff
@@ -184,6 +184,40 @@ class Sequence(PreTokenizer):
         """
         pass
 
+class Split(PreTokenizer):
```
patrickvonplaten, line 187: @n1t0 - what is the best way to re-generate this code? 

At first I ran `make style` which ran `python stub.py` which generated this code. Afterward I made changes to bindings/python/src/pre_tokenizers.rs and ran `make style` again - BUT this code wasn't changed accordingly...I had to manually delete the code and then after re-running `make style` it worked again. Not sure if you have encountered a similar problem before.

patrickvonplaten:
@n1t0 @n1t0 - I see that the CI complains about `AssertionError: The content of py_src/tokenizers/pre_tokenizers/__init__.pyi seems outdated, please run `python stub.py`. When I run `python stub.py` locally however, nothing changes...not sure what might be going on here - I'll see after I've done the rest of the PR

bindings/node/lib/bindings/pre-tokenizers.test.ts
```diff
@@ -44,6 +45,13 @@ describe("punctuationPreTokenizer", () => {
   });
 });
 
+describe("splitPreTokenizer", () => {
```
patrickvonplaten, line 48: This test fails -> not sure what's going on. I didn't manage to see a good error message :-/

bindings/node/native/src/pre_tokenizers.rs
```diff
@@ -7,6 +7,34 @@ use std::sync::Arc;
 use serde::{ser::SerializeStruct, Serialize, Serializer};
 use tk::pre_tokenizers::PreTokenizerWrapper;
 use tk::PreTokenizedString;
+use tk::normalizer::SplitDelimiterBehavior;
+
+#[derive(Clone)]
+struct JsSplitDelimiterBehavior(SplitDelimiterBehavior);
```
patrickvonplaten, line 13: I thought the best way is to add a `JsSplitDelimiterBehavior` struct to be able to convert `string` to `SplitDelimiterBehavior`.

The rust code compiles without error, but not very sure if everything is correct here

patrickvonplaten:
@n1t0 - it would be awesome if you could take over from here. I think the rust implementation is ok, the Python bindings seem to work, but there is a problem with the node bindings. The native rust code in `bindings/node` compiles without error, but the node tests fail (I'm having trouble getting good error messages out of these tests...).

bindings/python/src/pre_tokenizers.rs
n1t0: Yes, it looks great! I was wondering if we could find a good default for `behavior` but I think you are right and it's better like this.

bindings/python/pyproject.toml
n1t0: Awesome thanks! I'm used to directly use `python setup.py install|develop` so I didn't even know about this problem.

bindings/python/src/utils/normalization.rs
n1t0: Yes, I think it's the best place to have it. I like having the `From` defined in the same place that one of the involved types, so here just after `PyPattern` seems great.
I see you finally moved it to `utils/pretokenization.rs` so I'll move it back here.

bindings/python/py_src/tokenizers/pre_tokenizers/__init__.pyi
n1t0: That's weird. `python stub.py` needs to have the version of `tokenizers` in the env up-to-date, so if you compiled everything before running it, it should have worked fine.

bindings/node/lib/bindings/pre-tokenizers.test.ts
n1t0: How did you compile the node bindings? Did you run `npm run dev-rs`? I'm asking because it seems fine for the CI, and on my local install too.

n1t0 (approved):
Awesome! You did a great job for your first PR @patrickvonplaten! Both Python and Node bindings look great.

I just changed a few details, mainly about the docs and naming in node (we use camelCase there) but otherwise, everything seems to work perfectly!

bindings/node/lib/bindings/pre-tokenizers.test.ts
patrickvonplaten: ah yeah I didn't - didn't know how to compile the bindings, now I know :-)

bindings/node/native/src/pre_tokenizers.rs
```diff
@@ -19,12 +19,12 @@ impl FromJsValue for JsSplitDelimiterBehavior {
         Ok(Self(match s.as_ref() {
             "removed" => Ok(SplitDelimiterBehavior::Removed),
             "isolated" => Ok(SplitDelimiterBehavior::Isolated),
-            "merged_with_previous" => Ok(SplitDelimiterBehavior::MergedWithPrevious),
-            "merged_with_next" => Ok(SplitDelimiterBehavior::MergedWithNext),
+            "mergedWithPrevious" => Ok(SplitDelimiterBehavior::MergedWithPrevious),
```
patrickvonplaten, line 22: I see!

docs/source/components.rst
```diff
@@ -204,14 +234,22 @@ the ByteLevel)
...
    * - Split
-     - Versatile pre-tokenizer that splits on provided pattern and according to provided behavior. Optionally, invert flag can be provided to invert the pattern.
+     - Versatile pre-tokenizer that splits on provided pattern and according to provided behavior.
+       The pattern can be inverted if necessary.
 
          - pattern should be either a custom string or regexp.
-         - behavior should be one of :obj:`"removed"`, :obj:`"isolated"`, :obj:`"merged_with_previous"`, :obj:`"merged_with_next"`, or :obj:`"contiguous"`.
+         - behavior should be one of:
+
+            * :entity:`SplitDelimiterBehavior.removed`
```
patrickvonplaten, line 243: much nicer :-)

bindings/node/native/src/pre_tokenizers.rs
```diff
@@ -188,7 +188,7 @@ fn metaspace(mut cx: FunctionContext) -> JsResult<JsPreTokenizer> {
 /// split(invert: bool = false)
 fn split(mut cx: FunctionContext) -> JsResult<JsPreTokenizer> {
     let pattern: String = cx.extract::<String>(0)?;
-    let behavior: SplitDelimiterBehavior = cx.extract::<JsSDB>(1)?.0;
+    let behavior: JsSplitDelimiterBehavior = cx.extract::<JsSplitDelimiterBehavior>(1)?;
```
patrickvonplaten, line 191: Wasn't sure if that's the best way to do it with `JsSplitDelimiterBehavior` ...

n1t0: It is! I tried having `serde` use the original enum, and rename the fields to camelCase automatically but it didn't work!

n1t0: Doesn't seem to be implemented: https://github.com/serde-rs/serde/issues/1061

n1t0:
@Narsil it's really weird, the CI is not happy when running `python stub.py --check` while in local it works as expected. Do you have an idea what could be happening?

patrickvonplaten:
> @Narsil it's really weird, the CI is not happy when running `python stub.py --check` while in local it works as expected. Do you have an idea what could be happening?

Could it maybe be due to https://github.com/huggingface/tokenizers/pull/542#discussion_r530324322 ? Not sure...

PR #543, opened 2020-11-25
Update Python bindings README.md

JavClaude:
>>> tokenizer.train(["./path/to/dataset/1.txt"], trainer=trainer)

Fix the "PyTokenizer.train() got multiple values for argument: trainer" error when the "files" parameter is missing

### DIFF
--- a/bindings/python/README.md
+++ b/bindings/python/README.md
@@ -138,7 +138,7 @@ tokenizer.post_processor = processors.ByteLevel(trim_offsets=True)
 
 # And then train
 trainer = trainers.BpeTrainer(vocab_size=20000, min_frequency=2)
-tokenizer.train([
+tokenizer.train(files=[
 	"./path/to/dataset/1.txt",
 	"./path/to/dataset/2.txt",
 	"./path/to/dataset/3.txt"


n1t0:
I'm not sure to understand. The `files` parameter is not optional so it cannot be there.

JavClaude:
The parameter is not mentioned in the documentation. What induces the error when trying the piece of code

marcglobality:
I just came here because was having the mentioned error when running the example in the docs. Any reason not to merge this @n1t0 ?

marcglobality:
I think the error appeared since https://github.com/huggingface/tokenizers/pull/544

n1t0:
The README reflects the most up to date state of the library, not the latest stable. If you try the `0.10.0rc1`, I'm pretty sure this error does not exist.
Maybe we should make this clearer in the readme though.

marcglobality:
> The README reflects the most up to date state of the library, not the latest stable. If you try the `0.10.0rc1`, I'm pretty sure this error does not exist.
> Maybe we should make this clearer in the readme though.

For the future, I think the README should be updated to the latest stable, not the latest rc. Given that eventually the RC will becomes stable and the example in the README will work, I think @JavClaude can close this PR then?

If you feel motivated, it'd be great to get in the README the example here https://github.com/huggingface/tokenizers/pull/544/commits/723ee7c9ee3b1619463d411784e358166098c691#diff-810f2de0965b0a169a00cee33465f4165347b006b5360ae3b1001723a6e9ebd3R19 to train with in-memory data

PR #549, opened 2020-12-04
CI - Fix docs deployment

n1t0:
I have no idea why this stopped working, or why it used to work in the first place, but had to specify the ssh user. Using a ssh host to make it easier.

cc @julien-c


julien-c:
👍  and sorry for breaking this :)

PR #593, opened 2021-01-12
Cleans up a few pattern-matches into their Option/Result equivalent

huitseeker:
Drive-by tweaks, opinions on this loosely held.


n1t0:
:+1: Thank you!

PR #686, opened 2021-04-21
Fix SPM conversions

LysandreJik:
This PR fixes an issue with the SPM converters (ALBERT and XLNet) where it would replace some characters by whitespace - after removing double whitespace occurrences. This meant that if double whitespace were to appear thanks to this replacement, they would be kept until the end of encoding, leading to a mismatch between SentencePiece and tokenizers.

Fixes https://github.com/huggingface/transformers/pull/11358


n1t0 (approved):
Thank you very much for this @LysandreJik!

Issue #667, opened 2021-03-24
[Feature] : Wav2VecTokenizer


Ierezell:
It's possible to export a Wav2VecForCTC Huggingface model to onnx and to use it in other programming stack. 

<details>
<summary>Code for exporting a Wav2Vec Model</summary>
<pre>
<code>
from transformers import Wav2Vec2Tokenizer, Wav2Vec2ForCTC
from datasets import load_dataset
import torch
import os
import shutil
import soundfile as sf

MODEL_NAME = "facebook/wav2vec2-base-960h"
MODEL_FOLDER = f"./asr/{MODEL_NAME.replace('/','_')}"
OUTPUT_PATH = f"{MODEL_FOLDER}/model.onnx"

if os.path.exists(MODEL_FOLDER):
    shutil.rmtree(MODEL_FOLDER)

os.makedirs(MODEL_FOLDER)

tokenizer = Wav2Vec2Tokenizer.from_pretrained(MODEL_NAME)
model = Wav2Vec2ForCTC.from_pretrained(MODEL_NAME)
model = model.eval()

def map_to_array(batch):
    speech, _ = sf.read(batch["file"])
    batch["speech"] = speech
    return batch

ds = load_dataset("patrickvonplaten/librispeech_asr_dummy", "clean", split="validation")
ds = ds.map(map_to_array)

dummy_input = tokenizer(ds["speech"][:2], return_tensors="pt",
                        padding="longest").input_values

torch.onnx.export(model, dummy_input, OUTPUT_PATH, verbose=True,
                  input_names=["input"], output_names=["output"],
                  dynamic_axes={'input': {1: 'sequence'}})

tokenizer.save_pretrained(MODEL_FOLDER)
</code>
</pre>
</details>

<details>
<summary>Code for using the Wav2Vec model in nodejs</summary>
<pre>
<code>
import * as tf from '@tensorflow/tfjs-node';
import fs from 'fs';
import * as ort from "onnxruntime";
import path from 'path';

export class ASR
{
  private _session?: ort.InferenceSession
  private tokToId: { [key: string]: number }
  private idToTok: { [key: number]: string } = {}

  constructor(private modelPath: string)
  {
    this.tokToId = JSON.parse(fs.readFileSync(path.join(this.modelPath, 'vocab.json'), 'utf-8'))
    for (const [key, value] of Object.entries(this.tokToId)) { this.idToTok[value] = key }
  }

[...]

PR #691, opened 2021-05-02
Rust - add a CTCDecoder as a seperate mod

ThomAub:
A first implementation to help on #667 

It's currently a separate module that impl the trait `Decoder`


Narsil:
@ThomAub You can check out a followup PR:

https://github.com/huggingface/tokenizers/pull/693

This should include Node + Python bindings. 

I didn't check that we could create a real Wav2Vec2 Tokenizer and that is serializes deserializes correctly but it should be doable.

@n1t0 anything else I might have forgotten ?

n1t0:
Superseded by https://github.com/huggingface/tokenizers/pull/693

PR #693, opened 2021-05-04
Adding bindings to Node + Python.

Narsil:


ThomAub:
Thanks ! it looks nice :)

Narsil:
I think it's good, we just need to fix the tests and we're good to go.

Narsil:
@n1t0 do you think we can merge this ?

n1t0 (approved):
Thank you very much for your contribution @ThomAub, it looks very good. Thank you @Narsil for adding the bindings too.

Sorry for the delay in my review. I just updated the name of the struct to follow the current naming convention (`BPEDecoder` is an exception that we didn't fix for backward compatibility reasons).

PR #718, opened 2021-05-26
Fix word level tokenizer determinism

### LINKED ISSUE #717: WordLevelTrainer not deterministic
Hi everyone and thanks for the great work.
I'm using the library to create a word level tokenizer. Unfortunately, given the exact same data, the output given by the class `WordLevelTrainer` is not deterministic. I suspect this happens when the count of two words is exactly the same. Their order in the resulting json (and thus the corresponding token id) is sometimes inverted.

I don't know much Rust, but I believe it would be an easy fix [at line 43 of `src/models/wordlevel/trainer.rs`](https://github.com/huggingface/tokenizers/blob/755e5f5c1e5b955fd41b72635701ffb60cb0d726/tokenizers/src/models/wordlevel/trainer.rs#L43). If understand correctly, here the hashmap (dictionary) is ordered by values, with ignored keys (_). It would be sufficient to order it by keys in case of same counts.

Reproducibility of the tokenizer given the same data can be very useful in some scenarios.
Let me know if I should look a bit into rust and create a pull request for this or if I understood the code wrong and the source non-determinism is somewhere else.

Thanks a lot for the help!
Luca

Using  `tokenizers==0.10.3`

lucacampanella:
Fixes #717 
If two word level tokens have the same counts, their order is determined by the token itself alphabetically. This allows for determinism in the resulting tokenizer.

I have tested the code only in the online playground of rust ([here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=14e462cded705bd9f8d590624004e88b)), as setting up everything on my system would take too much time. I thus didn't run any tests.

Thanks for taking a look at this. :)

### DIFF
--- a/tokenizers/src/models/wordlevel/trainer.rs
+++ b/tokenizers/src/models/wordlevel/trainer.rs
@@ -1,6 +1,7 @@
 use super::WordLevel;
 use crate::utils::parallelism::*;
 use crate::{AddedToken, Result, Trainer};
+use std::cmp::Ordering;
 use std::collections::HashMap;
 
 #[non_exhaustive]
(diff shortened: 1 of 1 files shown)


lucacampanella:
Hi @n1t0, did you have a chance to have a look at this? 
Thanks a lot :)

n1t0 (approved):
Awesome, thank you for taking care of this @lucacampanella!

I just fixed a few Clippy warnings and added a missing import at the beginning. Everything is working perfectly.

PR #728, opened 2021-06-04
Switch git dependencies in Cargo.toml back to regular versions

geofft:
rayon-cond turned out to be a rustc bug that has been fixed for a while
(see cuviper/rayon-cond#2), so we can revert the git dependency.

numpy has released the commit in question as part of 0.12.


n1t0 (approved):
Great news! Thank you for taking care of this @geofft!

PR #758, opened 2021-07-16
update lexical-core because 0.7.4 doesn't compile

KoichiYasuoka:
fix the issue as described in https://github.com/rust-lang/rust/issues/81654


n1t0 (approved):
:+1: Thank you @KoichiYasuoka

PR #762, opened 2021-07-19
[WIP] Add a way to specify the unknown token in `SentencePieceUnigramTokenizer` python implem

SaulLu:
# What does this PR do?

This PR proposes a way to specify the unknown token in the `SentencePieceUnigramTokenizer` object (python implementation). :slightly_smiling_face: 

This PR also adds tests to ensure that the new tokenizer does not return errors when asked to tokenize a sentence containing tokens never seen in the training set. I'll leave some comments below to explain the few additional changes.

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. I would love to hear your opinion @n1t0 and @Narsil. :smile:


bindings/python/py_src/tokenizers/implementations/sentencepiece_unigram.py
```diff
@@ -49,13 +49,29 @@ def train(
...
+            files (:obj:`List[str]`):
+                A list of path to the files that we should use for training
+            vocab_size (:obj:`int`):
+                The size of the final vocabulary, including all tokens and alphabet.
+            show_progress (:obj:`bool`):
+                Whether to show progress bars while training.
+            special_tokens (:obj:`List[Union[str, AddedToken]]`):
+                A list of special tokens the model should know of.
+            unk_token (:obj:`str`, `optional`):
+                The unknown token to be used by the model.
```
SaulLu, line 67: Not all implementations in the same folder describe method arguments (I think because these objects are not in the documentation). Nevertheless, here I thought it could be really interesting to have them so that the user can understand what the new added `unk_token` argument is for.

bindings/python/setup.py
```diff
@@ -2,7 +2,7 @@
 from setuptools_rust import Binding, RustExtension
 
 extras = {}
-extras["testing"] = ["pytest"]
+extras["testing"] = ["pytest", "requests", "numpy", "datasets"]
```
SaulLu, line 5: I propose to add these dependencies that I had to install to run the python tests.

bindings/python/tests/implementations/test_sentencepiece.py
```diff
@@ -14,10 +16,40 @@ def test_train_from_iterator(self):
...
+
+    def test_train_with_unk_token(self, tmpdir):
+        p = tmpdir.mkdir("tmpdir").join("file.txt")
+        p.write("A first sentence\nAnother sentence\nAnd a last one")
+
+        tokenizer = SentencePieceUnigramTokenizer()
+        tokenizer.train(
+            files=str(p), show_progress=False, special_tokens=["<unk>"], unk_token="<unk>"
+        )
+        output = tokenizer.encode("A sentence 🤗")
```
SaulLu, line 37: The `"🤗"` of the sentence hasn't been seen by the tokenizer during the training

bindings/python/py_src/tokenizers/implementations/sentencepiece_unigram.py
Narsil: Yes, totally fine.

bindings/python/setup.py
Narsil: Fine for me.

bindings/python/tests/implementations/test_sentencepiece.py
```diff
@@ -14,10 +16,40 @@ def test_train_from_iterator(self):
...
 class TestSentencePieceUnigram:
+    def test_train(self, tmpdir):
+        p = tmpdir.mkdir("tmpdir").join("file.txt")
+        p.write("A first sentence\nAnother sentence\nAnd a last one")
+
+        tokenizer = SentencePieceUnigramTokenizer()
+        tokenizer.train(files=str(p), show_progress=False)
+
+        output = tokenizer.encode("A sentence")
+        assert output.tokens == ["▁A", "▁", "s", "en", "t", "en", "c", "e"]
```
Narsil, line 27: Could you assert it triggers an exception with unknown token ?

bindings/python/tests/implementations/test_sentencepiece.py
```diff
@@ -14,10 +16,40 @@ def test_train_from_iterator(self):
...
+    def test_train_with_unk_token(self, tmpdir):
+        p = tmpdir.mkdir("tmpdir").join("file.txt")
+        p.write("A first sentence\nAnother sentence\nAnd a last one")
+
+        tokenizer = SentencePieceUnigramTokenizer()
+        tokenizer.train(
+            files=str(p), show_progress=False, special_tokens=["<unk>"], unk_token="<unk>"
+        )
+        output = tokenizer.encode("A sentence 🤗")
+        assert output.ids[-1] == 0
```
Narsil, line 38: Could you assert the full list of `ids` and `tokens` (like the other test ?)

Narsil (approved):
LGTM (small nits)

bindings/python/tests/implementations/test_sentencepiece.py
SaulLu: Regarding the `ids` assert, this is what I initially wanted to do but the algorithm doesn't always calculate exactly the same vocabulary: when I run this test several times one id in the id list changes (between `7` and `8`). 

Concretely, the output of `print(output.ids, output.tokens)` is sometimes:
```
[12, 1, 4, 5, 2, 5, 7, 3, 1, 0] ['▁A', '▁', 's', 'en', 't', 'en', 'c', 'e', '▁', '🤗']
```
and other times:
```
[12, 1, 4, 5, 2, 5, 8, 3, 1, 0] ['▁A', '▁', 's', 'en', 't', 'en', 'c', 'e', '▁', '🤗']
```

Would it be okay if I replaced `assert output.ids[-1] == 0` with `assert (output.ids == [12, 1, 4, 5, 2, 5, 7, 3, 1, 0] or output.ids == [12, 1, 4, 5, 2, 5, 8, 3, 1, 0])? :slightly_smiling_face:

bindings/python/tests/implementations/test_sentencepiece.py
SaulLu: Great idea! I'll do it :slightly_smiling_face:

bindings/python/tests/implementations/test_sentencepiece.py
SaulLu: Regarding the assert on the token list, the following assert fails: 
``` python
assert output.tokens == ["▁A", "▁", "s", "en", "t", "en", "c", "e", "_", "<unk>"]
```
with the error:
```
>       assert output.tokens == ["▁A", "▁", "s", "en", "t", "en", "c", "e", "_", "<unk>"]
E       AssertionError: assert ['▁A', '▁', '...t', 'en', ...] == ['▁A', '▁', '...t', 'en', ...]
E         At index 8 diff: '▁' != '_'
E         Full diff:
E         - ['▁A', '▁', 's', 'en', 't', 'en', 'c', 'e', '_', '<unk>']
E         ?                                              ^    ^^^^^
E         + ['▁A', '▁', 's', 'en', 't', 'en', 'c', 'e', '▁', '🤗']
E         ?                                              ^    ^
```
Is this indeed the expected behavior?

Narsil: So if the tests are not deterministic definitely let's not add them.

However it's suprising that it's not determinstic though..... It shouldn't be ....

▁ != _  
first is a special unicode character (rarely used that's why google chose it)
Second is regular underscore.

Narsil: <unk> instead of 🤗 is indeed expected

SaulLu: Thanks a lot for your answer. I did not make any change for the assert on the `ids` and included the change to put the assert on the `tokens`. Can you check to see if this corresponds to what you expected ? :blush:

Narsil: That's better, and OK I understand better why IDs are not deterministic, they essentially have the same score, so no particular reason that ids should be in a particular order.

Narsil (approved):
LGTM

bindings/python/tests/implementations/test_sentencepiece.py
SaulLu: Thanks for your feedback and for investigating the change in ids. It makes sense!

PR #768, opened 2021-07-27
Add `unk_token` argument in UnigramTrainer documentation

SaulLu:
# What does this PR do?

This PR proposes to add the `unk_token` argument to the python documentation of `UnigramTrainer`. This addition is especially motivated by [this issue](https://github.com/huggingface/transformers/issues/12756) on the :hugs: transformers repo.

## Additional details

To add this part to the documentation, I manually modified the file `bindings/python/src/trainers.rs` and then run the command `python stub.py` to automatically modify the `bindings/python/py_src/tokenizers/trainers/__init__.pyi` file.

I have tested the change locally, here is a screen of the modified part of the documentation
image

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. I would love to hear your opinion @n1t0 and @Narsil. :smile:

PR #770, opened 2021-08-04
Expand documentation of UnigramTrainer

sgugger:
Currently the documentation of the [UnigramTrainer](https://huggingface.co/docs/tokenizers/python/latest/api/reference.html#tokenizers.trainers.UnigramTrainer) does not contain all the options you can set, and the auto-complete in an IDE does not show them either.

This PR fixes that.

### DIFF
--- a/bindings/python/src/trainers.rs
+++ b/bindings/python/src/trainers.rs
@@ -684,7 +684,7 @@ impl PyWordLevelTrainer {
 ///         The number of iterations of the EM algorithm to perform before
 ///         pruning the vocabulary.
 #[pyclass(extends=PyTrainer, module = "tokenizers.trainers", name=UnigramTrainer)]
-#[text_signature = "(self, vocab_size=8000, show_progress=True, special_tokens= [])"]
+#[text_signature = "(self, vocab_size=8000, show_progress=True, special_tokens=[], shrinking_factor=0.75, unk_token=None, max_piece_length=16, n_sub_iterations=2)"]
 pub struct PyUnigramTrainer {}
 #[pymethods]
 impl PyUnigramTrainer {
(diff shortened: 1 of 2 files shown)


Narsil:
@sgugger I think we should update `bindings/python/src/trainers.rs` instead.

The `.pyi` files are generated automatically with `python stub.py` (so we have consistent doc across Rust and Python).

sgugger:
Oh I didn't know that, added the same there.

bindings/python/src/trainers.rs
```diff
@@ -669,6 +669,20 @@ impl PyWordLevelTrainer {
...
+///         The token used for out-of-vocabulary tokens.
+///
+///     max_piece_length (:obj:`int`):
+///         The maximum length of a given token.
+///
+///     n_sub_iterations (:obj:`int`):
+///         The number of iterations of the EM algorithm to perform before
+///         pruning the vocabulary.
 #[pyclass(extends=PyTrainer, module = "tokenizers.trainers", name=UnigramTrainer)]
 #[text_signature = "(self, vocab_size=8000, show_progress=True, special_tokens= [])"]
```
Narsil, line 687: I think this needs editing too. (There should be a test to detect that, let's see if it kicks in correctly.)

Narsil:
@sgugger fyi: https://github.com/huggingface/tokenizers/pull/768/files

sgugger:
I had not seen this PR from @SaulLu. I think this PR is more comprehensive as it adds all the arguments the `UnigramTrainer` can accept.

Narsil:
https://github.com/huggingface/tokenizers/pull/773 (The rust version update)

n1t0 (approved):
LGTM, Thank you for taking care of this @sgugger!

PR #837, opened 2021-12-02
Add truncate overflow with stride test case

McPatate:
Adding a test case to truncate.

The goal is to make refactoring easier and later implement a `truncate_right` method (or `truncate(left: bool)`).


Narsil (approved):
LGTM, CI complaints are clippy only :)

PR #846, opened 2021-12-14
Clippy fixes.

Narsil:
Fixing all new clippy warnings.

### DIFF
--- a/bindings/node/native/src/tasks/tokenizer.rs
+++ b/bindings/node/native/src/tasks/tokenizer.rs
@@ -11,7 +11,7 @@ pub enum EncodeTask<'s> {
 }
 
 pub enum EncodeOutput {
-    Single(Encoding),
+    Single(Box<Encoding>),
     Batch(Vec<Encoding>),
 }
 
@@ -35,7 +35,7 @@ impl Task for EncodeTask<'static> {
                         *add_special_tokens,
                     )
                     .map_err(|e| format!("{}", e))
-                    .map(EncodeOutput::Single)
+                    .map(|item| EncodeOutput::Single(Box::new(item)))
             }
             EncodeTask::Batch(worker, input, add_special_tokens) => {
                 let mut input: Option<Vec<EncodeInput>> =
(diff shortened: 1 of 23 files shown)


bindings/node/native/src/tasks/tokenizer.rs
```diff
@@ -11,7 +11,7 @@ pub enum EncodeTask<'s> {
 }
 
 pub enum EncodeOutput {
-    Single(Encoding),
+    Single(Box<Encoding>),
```
McPatate, line 14: Is `Box` needed because Javascript? 😅

bindings/node/package.json
```diff
@@ -29,7 +29,7 @@
     "eslint-plugin-prettier": "^3.4.1",
     "eslint-plugin-simple-import-sort": "^5.0.3",
     "jest": "^26.6.3",
-    "neon-cli": "^0.3.3",
+    "neon-cli": "^0.9.1",
```
McPatate, line 32: Big jump! 😁

McPatate (approved):
LGTM

bindings/node/native/src/tasks/tokenizer.rs
Narsil: No, clippy yells becuse that variant is 240 bytes long (pretty long for an enum). Setting it as a box just uses a pointer instead.

bindings/node/package.json
Narsil: `npm audit --fix` did it !

Narsil:
I will merge this so we can rebase other PRs.

@n1t0 we can still discuss some choices here.

- 3.6 MacOS as a target was dropped by GA (so we cannot run it)
- Had to modify the testing because MacOS Python 3.8+ uses `spawn` and not `fork` anymore. (Just modifying the tests seems reasonable)
- Had to bump `setup-python@v2` because otherwise the lib location was not found by rust.
- Had to `npm audit --fix` (security warning for `tmpl`) 
- `pyarrow < 3.0` was invalid since now datasets requires `>= 3.0` and the original reason for downgrading seems gone.

PR #850, opened 2021-12-17
Feature gate http-deps

MarcusGrass:
Make pulling in reqwest and cached-path optional, reducing the dependency footprint from 235 to 117 with a 20% release build time reduction (in very primitive timing on my specific machine). 

Useful for providing some additional safety if running on production servers where dynamically downloading external files are prohibited.

### DIFF
--- a/tokenizers/src/lib.rs
+++ b/tokenizers/src/lib.rs
@@ -26,11 +26,13 @@
 //! use tokenizers::tokenizer::{Result, Tokenizer};
 //!
 //! fn main() -> Result<()> {
-//!     let tokenizer = Tokenizer::from_pretrained("bert-base-cased", None)?;
+//!     #[cfg(feature = "http")]
+//!     {
+//!         let tokenizer = Tokenizer::from_pretrained("bert-base-cased", None)?;
 //!
-//!     let encoding = tokenizer.encode("Hey there!", false)?;
-//!     println!("{:?}", encoding.get_tokens());
-//!
+//!         let encoding = tokenizer.encode("Hey there!", false)?;
+//!         println!("{:?}", encoding.get_tokens());
+//!     }
 //!     Ok(())
 //! }
 //! ```
@@ -145,4 +147,5 @@ pub use tokenizer::*;
 pub use utils::parallelism;
 
 // Re-export for from_pretrained
+#[cfg(feature = "http")]
 pub use utils::from_pretrained::FromPretrainedParameters;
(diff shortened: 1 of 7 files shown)


tokenizers/src/lib.rs
```diff
@@ -26,11 +26,13 @@
...
-//!
-//!     let encoding = tokenizer.encode("Hey there!", false)?;
-//!     println!("{:?}", encoding.get_tokens());
+//!     #[cfg(feature = "http")]
+//!     {
+//!         let tokenizer = Tokenizer::from_pretrained("bert-base-cased", None)?;
 //!
+//!         let encoding = tokenizer.encode("Hey there!", false)?;
+//!         println!("{:?}", encoding.get_tokens());
+//!     }
```
Narsil, line 35: Do you mind hiding these as they are mostly necessary for the test but shouldn't be displayed in the doc (since it's a default feature)

```suggestion
//!     # #[cfg(feature = "http")]
//!     # {
//!      let tokenizer = Tokenizer::from_pretrained("bert-base-cased", None)?;
//!
//!      let encoding = tokenizer.encode("Hey there!", false)?;
//!      println!("{:?}", encoding.get_tokens());
//!     # }
```

tokenizers/src/utils/from_pretrained.rs
```diff
@@ -1,7 +1,6 @@
 use crate::Result;
 use cached_path::CacheBuilder;
 use itertools::Itertools;
-use reqwest::{blocking::Client, header};
```
Narsil, line 4: Do you mind reverting this, this is not necessary for this PR.

tokenizers/src/utils/from_pretrained.rs
```diff
@@ -96,6 +95,7 @@ impl Default for FromPretrainedParameters {
     }
 }
 
+#[cfg(feature = "http")]
```
Narsil, line 98: This isn't necessary since the whole file is already behind a feature flag.

Narsil (changes requested):
Thanks for this PR, very valid concern and PR seems good.

A few nits before merging please.

Cheers !

tokenizers/src/lib.rs
Narsil: You will need to modify the README.md file too (`cargo readme > README.md` should be enouh)

MarcusGrass: Wow I spent so long trying to figure out how to do that, nice, fixed!

Narsil (approved):
LGTM

PR #858, opened 2021-12-24
Feature: Handle invalid truncate direction

McPatate:
As requested in #841.

### DIFF
--- a/bindings/node/lib/bindings/raw-encoding.test.ts
+++ b/bindings/node/lib/bindings/raw-encoding.test.ts
@@ -105,9 +105,8 @@ describe("RawEncoding", () => {
       expect(encoding.truncate(10, undefined)).toBeUndefined();
     });
     it("should throw an Error on invalid direction", () => {
-      const invalid_dir = "not_valid";
-      const t = () => encoding.truncate(10, 3, invalid_dir);
-      expect(t).toThrow(`Invalid truncation direction value : ${invalid_dir}`);
+      const t = () => encoding.truncate(10, 3, "not_valid");
+      expect(t).toThrow(`Invalid truncation direction value : not_valid`);
     });
   });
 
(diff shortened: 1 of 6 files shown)


bindings/node/lib/bindings/raw-encoding.test.ts
```diff
@@ -104,6 +104,11 @@ describe("RawEncoding", () => {
     it("accepts `undefined` as second parameter", () => {
       expect(encoding.truncate(10, undefined)).toBeUndefined();
     });
+    it("should throw an Error on invalid direction", () => {
+      const invalid_dir = "not_valid";
+      const t = () => encoding.truncate(10, 3, invalid_dir);
+      expect(t).toThrow(`Invalid truncation direction value : ${invalid_dir}`);
+    });
```
Narsil, line 111:
```suggestion
    it("should throw an Error on invalid direction", () => {
      const t = () => encoding.truncate(10, 3, "not_valid");
      expect(t).toThrow(`Invalid truncation direction value : not_valid`);
    });
```
Can we keep everything static in the assertion ?

Test logic should be minimal and that includes templating IMO.

bindings/python/tests/bindings/test_encoding.py
```diff
@@ -106,3 +106,16 @@ def test_char_to_word(self, encodings):
...
+        single.truncate(2, 1, "right")
+        assert single.tokens == ["[CLS]", "i"]
+        assert single.overflowing[0].tokens == ["i", "love"]
+
+    def test_invalid_truncate_direction(self, encodings):
+        single, _ = encodings
+        invalid_direction = "not_a_direction_:("
+        with pytest.raises(ValueError) as excinfo:
+            single.truncate(2, 1, invalid_direction)
+        assert f"Invalid truncation direction value : {invalid_direction}" in str(excinfo.value)
```
Narsil, line 121:
```suggestion
        with pytest.raises(ValueError) as excinfo:
            single.truncate(2, 1, "not_a_direction")
        assert "Invalid truncation direction value : not_a_direction" == str(excinfo.value)
```

Narsil (approved):
looks ok.

Small nits to keep tests even simpler imho.

PR #859, opened 2021-12-25
Improve truncation_side

NielsRogge:
This PR:

- adds the "direction" parameter to `enable_truncation`, as this is required for it to be used in Transformers. 
- renames `TruncateDirection` to `TruncationDirection` everywhere (to be consistent with `PaddingDirection`, sorry bit of nitpicking here)

To do:

- [ ] add tests to [test_tokenizer.py](https://github.com/huggingface/tokenizers/blob/master/bindings/python/tests/bindings/test_tokenizer.py) for testing padding/truncation on the right/left

### DIFF
--- a/bindings/python/src/tokenizer.rs
+++ b/bindings/python/src/tokenizer.rs
@@ -731,6 +731,7 @@ impl PyTokenizer {
             dict.set_item("max_length", params.max_length)?;
             dict.set_item("stride", params.stride)?;
             dict.set_item("strategy", params.strategy.as_ref())?;
+            dict.set_item("direction", params.direction.as_ref())?;
 
             Ok(Some(dict))
         })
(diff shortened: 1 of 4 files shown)


bindings/python/src/tokenizer.rs
```diff
@@ -687,6 +687,19 @@ impl PyTokenizer {
...
                         }?
                     }
+                    "direction" => {
+                        let value: &str = value.extract()?;
+                        params.direction = match value {
+                            "left" => Ok(TruncationDirection::Left),
+                            "right" => Ok(TruncationDirection::Right),
+                            other => Err(PyError(format!(
+                                "Unknown `direction`: `{}`. Use \
+                                 one of `left` or `right`",
```
McPatate, line 697: I think we should keep the wording coherent between all our invalid direction errors. I like the fact you added the correct values to the err msg.

Could you update the other direction error messages too ?

McPatate:
`TruncationDirection` has been renamed in #858. See if you can rebase (once it's merged ?).

Narsil:
I am on it.

Narsil:
Probably to be closed, as it's superseeded by https://github.com/huggingface/tokenizers/pull/860 ( Started with your commits so credit is where it's due, but I had to edit other things to make it work)

PR #870, opened 2022-01-05
Fix wordlevel encode `<unk>`

### LINKED ISSUE #746: WordLevel does not output [UNK] token
1. I trained a WordLevel tokenizer from scratch on the first few lines of the "Zen of Python".
2. I then encoded a sentence that included two words that did not appear in the training data.
3. When I output the tokens, the original words instead of the special [UNK] token appear. Why?

Unknown words: "and", "🤗"
Encode: "sparse is complex and 🤗 beautiful"
Token ids: [1, 16, 6, 9, 0, 0, 11, 2]
Tokens: ['[CLS]', 'sparse', 'is', 'complex', 'and', '🤗', 'beautiful', '[SEP]']
Decode: 'sparse is complex beautiful'

I am using tokenizers 0.10.3 on Python 3.8.5 in Jupyter notebook on MacOS 10.15.7

Here is my code:

INPUT
```
from tokenizers.processors import TemplateProcessing
from tokenizers.trainers import WordLevelTrainer

# Tokenizer-Model
bert_tokenizer = Tokenizer(WordLevel(unk_token="[UNK]"))

# Pre-tokenizer
bert_tokenizer.pre_tokenizer = Whitespace()

# Post-processor
bert_tokenizer.post_processor = TemplateProcessing(
    single="[CLS] $A [SEP]",
    pair="[CLS] $A [SEP] $B:1 [SEP]:1",
    special_tokens=[
        ("[CLS]", 1),
        ("[SEP]", 2),
    ],
)

# Define Trainer
trainer = WordLevelTrainer(
     min_frequency=0, show_progress=True, special_tokens=["[UNK]", "[CLS]", "[SEP]", "[PAD]", "[MASK]"]
)

# Create Data
data = [
    "beautiful is better than ugly."
    "explicit is better than implicit."
    "simple is better than complex."
    "complex is better than complicated."
    "flat is better than nested."
    "sparse is better than dense."
    "readability counts."
]

# Train Tokenizer
bert_tokenizer.train_from_iterator(data, trainer)

# Encode a text: Include 2 unknown words: "and" "🤗" (i.e., words not in training data)
test = bert_tokenizer.encode("sparse is complex and 🤗 beautiful")

# Show token ids: Returns 0 for unknown words
print(test.ids)

# Show Tokens: Instead of the expected special "[UNK]" token for unknown words, returns actual words!
print(test.tokens)

# Decode into sentence: Correct: unknown words are missing
bert_tokenizer.decode(test.ids) 
```

OUTPUT:

[1, 16, 6, 9, 0, 0, 11, 2]
['[CLS]', 'sparse', 'is', 'complex', 'and', '🤗', 'beautiful', '[SEP]']
'sparse is complex beautiful'

mishig25:
Closes #746

#746 is a bug? If not, please feel free to ignore this PR

let's say
```
# Tokenizer-Model
tokenizer = Tokenizer(WordLevel(unk_token="<unk>"))
with vocab = {"<unk>": 0, "a": 1, "b": 1}

# Pre-tokenizer
tokenizer.pre_tokenizer = Whitespace()
```

, then the expected behaviour for `tkns = tokenizer.encode("a b c")` should be 
```
tkns.tokens == ['a', 'b', '<unk>']
```

however, what actually gets produced is a buggy result
```
tkns.tokens == ['a', 'b', 'c']
```

and this PR fixes this bug

### DIFF
--- a/tokenizers/src/models/wordlevel/mod.rs
+++ b/tokenizers/src/models/wordlevel/mod.rs
@@ -177,7 +177,10 @@ impl Model for WordLevel {
                 .get(&*token)
                 .or_else(|| self.vocab.get(&*self.unk_token))
                 .ok_or(Error::MissingUnkToken)?,
-            value: token.to_owned(),
+            value: self
+                .vocab
+                .get(&*token)
+                .map_or(self.unk_token.clone(), |_| token.to_owned()),
             offsets: (0, token.len()),
         }])
     }
@@ -218,5 +221,28 @@ impl Model for WordLevel {
 
     fn get_trainer(&self) -> Self::Trainer {
         WordLevelTrainer::default()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_unk() {
+        let vocab: Vocab = [("<unk>".into(), 0), ("a".into(), 1), ("b".into(), 2)]
+            .iter()
+            .cloned()
+            .collect();
+        let wordlevel = WordLevelBuilder::default()
+            .vocab(vocab)
+            .unk_token("<unk>".to_string())
+            .build()
+            .unwrap();
+        let tokens = wordlevel.tokenize("c").unwrap();
+        assert_eq!(tokens, vec![Token::new(0u32, "<unk>".into(), (0, 1)),]);
+
+        let tokens = wordlevel.tokenize("a").unwrap();
+        assert_eq!(tokens, vec![Token::new(1u32, "a".into(), (0, 1)),]);
     }
 }
--- a/tokenizers/src/models/wordlevel/mod.rs
+++ b/tokenizers/src/models/wordlevel/mod.rs
@@ -229,7 +229,7 @@ mod tests {
     use super::*;
 
     #[test]
-    fn test_unk() {
+    fn test_tokenize_unk() {
         let vocab: Vocab = [("<unk>".into(), 0), ("a".into(), 1), ("b".into(), 2)]
             .iter()
             .cloned()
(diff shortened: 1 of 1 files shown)


tokenizers/src/models/wordlevel/mod.rs
```diff
@@ -177,7 +177,10 @@ impl Model for WordLevel {
                 .get(&*token)
                 .or_else(|| self.vocab.get(&*self.unk_token))
                 .ok_or(Error::MissingUnkToken)?,
-            value: token.to_owned(),
+            value: self
+                .vocab
+                .get(&*token)
+                .map_or(self.unk_token.clone(), |_| token.to_owned()),
```
Narsil, line 183: Other option which feel slightly cleaner IMO:

```rust
        if let Some(&id) = self.vocab.get(token) {
            Ok(vec![Token {
                id,
                value: token.to_owned(),
                offsets: (0, token.len()),
            }])
        } else if let Some(&unk_id) = self.vocab.get(&self.unk_token) {
            Ok(vec![Token {
                id: unk_id,
                value: self.unk_token.to_owned(),
                offsets: (0, token.len()),
            }])
        } else {
            Err(Box::new(Error::MissingUnkToken))
        }
```
It's personal taste, probably gets compiled the exact same way.

Narsil (approved):
LGTM.
Small nit if you want.

I also ran the full `transformers` test suite for tokenizer on this PR and everything passes (just something to try when modifying these peculiar behaviors, `transformers` might be relying on it in some odd way so we need to check that the test suite at least passes.

tokenizers/src/models/wordlevel/mod.rs
mishig25: indeed, it looks cleaner
in commit https://github.com/huggingface/tokenizers/pull/870/commits/a566cde962a47e331eeb7d8cff472c03784e1432,
1. made the suggested change
2. added a unit cases that tests the error

Narsil: Perfect, I'll go ahead an merge.

PR #877, opened 2022-01-11
add python3.10 release

jc-louis:
Add CI steps for a python 3.10 build

### DIFF
--- a/.github/workflows/python-release-conda.yml
+++ b/.github/workflows/python-release-conda.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, macos-latest]
-        python: [3.7, 3.8, 3.9]
+        python: ["3.7", "3.8", "3.9", "3.10"]
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -91,7 +91,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.9
+          python-version: "3.10"
           architecture: "x64"
 
       - name: Install
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python: [3.7, 3.8, 3.9]
+        python: ["3.7", "3.8", "3.9", "3.10"]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v1
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -91,7 +91,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@v2
         with:
-          python-version: "3.10"
+          python-version: 3.9
           architecture: "x64"
 
       - name: Install
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -30,6 +30,7 @@ setup(
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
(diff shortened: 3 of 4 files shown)


bindings/python/setup.py
```diff
@@ -30,6 +30,8 @@
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
```
jc-louis, line 33: already supports python3.9 but missing this [classifier](https://pypi.org/classifiers/)

.github/workflows/python-release-conda.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, macos-latest]
-        python: [3.7, 3.8, 3.9]
+        python: ["3.7", "3.8", "3.9", "3.10"]
```
jc-louis, line 17: we need to use quotes for `3.10` otherwise yaml will convert it to `3.1`

.github/workflows/python.yml
```diff
@@ -91,7 +91,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.9
+          python-version: "3.10"
```
jc-louis, line 94: not sure we want to keep running tests on 3.10 only, but it seems like the matrix (with multiple pythoon version) steps are only used for builds tests and not actual tests

Narsil: No for tests, 1 version is usually enough. (There is nothing very new to python world in this lib)

Narsil:
Seems like two tests need to be updated since the error message was slightly changed in 3.10:

```
commit 7d37b2cab045bb6372c1e9275642cc541b1be7ab (HEAD -> jc-louis/master)
Author: Nicolas Patry <patry.nicolas@protonmail.com>
Date:   Tue Jan 11 15:41:36 2022 +0100

    Update tests for Python 3.10

diff --git a/bindings/python/tests/bindings/test_normalizers.py b/bindings/python/tests/bindings/test_normalizers.py
index 6712472..a64b776 100644
--- a/bindings/python/tests/bindings/test_normalizers.py
+++ b/bindings/python/tests/bindings/test_normalizers.py
@@ -148,7 +148,7 @@ class TestCustomNormalizer:

         assert isinstance(bad, Normalizer)
         assert isinstance(good, Normalizer)
-        with pytest.raises(Exception, match="TypeError: normalize()"):
+        with pytest.raises(Exception, match="TypeError:.*normalize()"):
             bad.normalize_str("Hey there!")
         assert good.normalize_str("Hey there!") == "Hey you!"
         with pytest.raises(
diff --git a/bindings/python/tests/bindings/test_pre_tokenizers.py b/bindings/python/tests/bindings/test_pre_tokenizers.py
index e73b6ba..ead8fca 100644
--- a/bindings/python/tests/bindings/test_pre_tokenizers.py
+++ b/bindings/python/tests/bindings/test_pre_tokenizers.py
@@ -213,7 +213,7 @@ class TestCustomPreTokenizer:

         assert isinstance(bad, PreTokenizer)
         assert isinstance(good, PreTokenizer)
-        with pytest.raises(Exception, match="TypeError: pre_tokenize()"):
+        with pytest.raises(Exception, match="TypeError:.*pre_tokenize()"):
             bad.pre_tokenize_str("Hey there!")
         assert good.pre_tokenize_str("Hey there!") == [
             ("Hey there!", (0, 10)),
```

jc-louis:
@Narsil I think I should rather revert my last commit and keep testing on 3.9 only? 

Given that 3.10 is still "bleeding edge" (and most contributors might not be using it) it makes more sense to keep the tests as-is with 3.9?

Narsil:
As you wish, the tests can be updated regardless to get forward compatibility.

Narsil (approved):
LGTM

PR #884, opened 2022-01-17
Fixing bad deserialization following inclusion of a default for `Punctuation`.

Narsil:
Deserialization is a bit basic, and ignore `type` field.

IIRC, it's because we didn't always add it, and so old `tokenizers` might still have `pre_tokenizers` without any `type` field in them being serialized.

However https://github.com/huggingface/tokenizers/pull/882 introduced a pretty bad bug, since it allowed `Punctuation` to work without any data. Since it's based **Before** `WhitespaceSplit`, then all `WhitespaceSplit` would be deserialized as `Punctuation`.

In order to avoid that, there are 2 options:

- Change the order in the `PretokenizerWrapper` enum, but that seems quite brittle
- Force the presence of `type` in *some* pretokenizers.

I went for the second version to avoid further hacks like ordering of the enum, which is brittle and we're bound to forget about it later.

In order to make sure we don't break old tokenizers that would be using `Punctuation`, I am making a full sweep on the hub to see how things get loaded

### DIFF
--- a/tokenizers/src/pre_tokenizers/punctuation.rs
+++ b/tokenizers/src/pre_tokenizers/punctuation.rs
@@ -1,4 +1,4 @@
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::tokenizer::{PreTokenizedString, PreTokenizer, Result, SplitDelimiterBehavior};
 use unicode_categories::UnicodeCategories;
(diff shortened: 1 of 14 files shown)


tokenizers/src/pre_tokenizers/sequence.rs
```diff
@@ -3,6 +3,7 @@ use crate::tokenizer::{PreTokenizedString, PreTokenizer, Result};
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(test, derive(PartialEq))]
```
n1t0, line 6: Why not add it directly to `derive` in all cases ?

tokenizers/src/pre_tokenizers/punctuation.rs
```diff
@@ -7,13 +7,35 @@ fn is_punc(x: char) -> bool {
     char::is_ascii_punctuation(&x) || x.is_punctuation()
 }
 
-#[derive(Serialize, Deserialize, Copy, Clone, Debug)]
-#[serde(tag = "type")]
```
n1t0, line 11: Did you mean to remove this?

tokenizers/src/pre_tokenizers/sequence.rs
Narsil: Because it wasn't and I imagined it was intended.

tokenizers/src/pre_tokenizers/punctuation.rs
Narsil: No I didn't.

tokenizers/src/pre_tokenizers/sequence.rs
n1t0: Nope, I usually added them only when needed, feel free to include it directly!

PR #887, opened 2022-01-19
[FAILING] Attempt (failing) to remove custom `Deserialize`.

Narsil:


Narsil:
Was just to showcase an attempt to use pure `serde` which here shows that it would fail on `PyPreTokenizerWrapper` in Python which is another layer of enum which doesn't have a tag

PR #895, opened 2022-01-28
fix python3.10 build

jc-louis:
since https://github.com/huggingface/tokenizers/pull/877 we have python3.10 wheels  for macOS but not linux

### DIFF
--- a/bindings/python/build-wheels.sh
+++ b/bindings/python/build-wheels.sh
@@ -4,7 +4,7 @@ set -ex
 curl https://sh.rustup.rs -sSf | sh -s -- -y
 export PATH="$HOME/.cargo/bin:$PATH"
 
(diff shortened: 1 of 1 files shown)


Narsil:
This will happen on the next release.

ram02z:
Is it possible that you can have a minor release? I want to use `tokenizers` but I am getting a rust compiler error since there are no wheels for python 3.10

Narsil:
Hi, I understand the impatience, and it will come, but I can't promise a date.

ram02z:
> Hi, I understand the impatience, and it will come, but I can't promise a date.
> For reference torch still has no support : https://github.com/pytorch/pytorch/issues/67669

Thank you. BTW torch has support on their nightly releases.

Narsil:
It should be ok now ! Thanks for your patience.

PR #901, opened 2022-02-03
tokenizer.save has the wrong arguments compared to documentation

thomasw21:
Unless I'm wrong, `tokenizer.save` should accept a `path:str` as argument.

### DIFF
--- a/bindings/node/native/Cargo.lock
+++ b/bindings/node/native/Cargo.lock
@@ -1136,6 +1136,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "paste"
+version = "1.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


Narsil:
Actually, those files are generated. I took the liberty of modifying this and using `python stub.py` to regenerate those files.

Because we need to use signatures we can't use type hints though. (Well it has drawbacks at the very least)

thomasw21:
Thanks! Sorry did not know that that file was generated. Concerning some of the changes I don't know why there's additional dependency changes. Could you elaborate on those? Otherwise looks good

Narsil:
Well, they are not really related, but otherwise the quality would fail since the dependencies got changed upstream.

I need to find a way to make that less noisy (it didn't use to be)

PR #904, opened 2022-02-11
Load tokenizer via String

brandonmkunkel:
Being able to load a tokenizer via a string can allow enable loading tokenizers in use cases such as loading a string asset from an App Asset bundle or streaming over network. In these cases we do not have a file path to resolve and/or prefer not to save a file to a temporary location.

I added a test for this but couldn't get the test to pass in my local environment because I cannot find the "data/roberta.json" file.

### DIFF
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -406,9 +406,8 @@ impl Tokenizer {
         let tokenizer = serde_json::from_str(&content)?;
         Ok(tokenizer)
     }
-    pub fn from_string(content: String) -> Result<Self> {
-        let tokenizer = serde_json::from_str(&content)?;
-        Ok(tokenizer)
+    pub fn from_string(content: &str) -> Result<Self> {
+        serde_json::from_str(content)
     }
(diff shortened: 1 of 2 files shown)


Narsil (commented):
I left a nit.

Overall I think this is OK, but this should probable be also re-exported in `Python` bindings no ?

(I can take care of this if you want)

brandonmkunkel:
Thanks for the feedback @Narsil . Still learning about the Rust memory model, makes sense to use unowned strings to expand its usage. I accepted your change and updated the unit test to match your suggested string type

brandonmkunkel:
Interestingly enough, I was looking at the Python bindings and actually found a deserialization method that accepts a string in the Python bindings here (from main branch):

https://github.com/huggingface/tokenizers/blob/448054f3c7f7ae99d273e6a3b80d550841ec7d87/bindings/python/src/tokenizer.rs#L506

Seems the Python bindings already had that capability?

Narsil:
Definitely seems so :)

brandonmkunkel:
Yup I see that now, thanks for your time on this PR!  I'll close the PR

Issue #63, opened 2020-01-14
JS / WebAssembly binding planned ?


mikbry:
I see your Node.js binding using Neon. But have you considered WebAssembly ? There are some tools to compile Rust code easily. So you will get a browser compatiblity and node v13 with a low impact on speed.

n1t0:
Thank you for suggesting this. As far as I know, WebAssembly does not support multithreading so I don't know how this would integrate. We clearly don't want to sacrifice the speed on every other platform for the sake of having WebAssembly, but if we can work around this, it would indeed be really valuable to have such bindings!

mikbry:
Thank @n1t0 and great job ! I started playing and understanding how it works. I will take a look at WebAssembly binding, if you are ok.

VogtAI:
Any updates on this? I'm also very interested in running the models in the browser.

mikbry:
I made an experiment several months ago and got it running without multithreading.

Here is my branch but totally out of sync:
https://github.com/mikbry/tokenizers/tree/feat_wasm

But now it needs to be updated to master and multithreading using webworkers.  

I need to find time to work on it again.

josephrocca:
@Narsil Thanks for your interest in this! I'd definitely be willing to spend a week or two on this if you think it's feasible to get a JS/web version working within that sort of time frame. If it's likely to take longer than that, I'm still happy to have a shot at solving a chunk of the problem - especially if it could be useful to the OSS community in its own right (e.g. creating a wasm build of a required cpp module).

I've got a lot of JS experience, but only a bit of experience with Rust + wasm-pack. I'd probably need you to lay out the overall plan for me and ideally give me a heads up on which aspects might be harder than others. It would also be great if you could give me a few concrete first steps to get me rolling¹. Is that okay? No worries if you're currently too busy!

¹ My guess, just so you're aware of my mental models (which are probably quite bad), is that the best approach would, roughly, be to copy the node bindings, and then port the code to JS module syntax (`import ...` instead of `require("...")`), and then port the node native bindings code to wasm-pack bindings code, and then swap in ports of the cpp modules?

Narsil:
> I've got a lot of JS experience, but only a bit of experience with Rust + wasm-pack. I'd probably need you to lay out the overall plan for me and ideally give me a heads up on which aspects might be harder than others. It would also be great if you could give me a few concrete first steps to get me rolling¹. Is that okay? No worries if you're currently too busy!

No it's good of you to ask !

[...]

PR #908, opened 2022-02-16
Abstract regex and add `fancy_regex` backend

josephrocca:
This is the first part of one potential approach to getting Wasm and pure rust builds of tokenizers.

Relevant issue: https://github.com/huggingface/tokenizers/issues/63

Here's the command to test the `fancy_regex` backend:
```
cargo test --no-default-features --features progressbar,http,regex-fancy
```
It's a bit strange that there's not a nicer way to do that (IIUC), but it seems like this is a [known problem](https://github.com/rust-lang/cargo/issues/3126).

You can also test both backends at the same time (return values checked against one another for consistency):
```
cargo test --features regex-all-test
```

The abstraction layer code was created by the nlprule and syntect project devs, with only minor edits from myself:
* https://github.com/trishume/syntect/blob/master/src/parsing/regex.rs
* https://github.com/bminixhofer/nlprule/blob/main/nlprule/src/utils/regex.rs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 
 .vim
 .env
+.venv
 target
 .idea
 Cargo.lock
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -36,7 +36,6 @@ harness = false
 [dependencies]
 lazy_static = "1.4"
 rand = "0.7"
-onig = { version = "6.0", default-features = false }
 regex = "1.3"
 regex-syntax = "0.6"
 rayon = "1.3"
@@ -59,11 +58,18 @@ cached-path = { version = "0.5", optional = true }
 aho-corasick = "0.7"
 paste = "1.0.6"
 proc_macros = { path = "./src/utils/proc_macros" }
+once_cell = "1.8"
+cfg-if = "1"
+onig = { version = "6.0", default-features = false, optional = true }
+fancy-regex = { version = "0.7", optional = true }
 
 [features]
-default = ["progressbar", "http"]
+default = ["progressbar", "http", "regex-onig"]
 progressbar = ["indicatif"]
 http = ["reqwest", "cached-path"]
+regex-fancy = ["fancy-regex"]
+regex-onig = ["onig"]
+regex-all-test = ["regex-onig", "regex-fancy"]
 
 [dev-dependencies]
 criterion = "0.3"
--- a/tokenizers/src/utils/regex.rs
+++ b/tokenizers/src/utils/regex.rs
@@ -1,0 +1,497 @@
+// All code and comments below this first line are copied from here (with some edits): https://github.com/bminixhofer/nlprule/blob/main/nlprule/src/utils/regex.rs
+
+
+
+//! An abstraction on top of Regular Expressions to add support for Serialization and
+//! to modularize the Regex backend.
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -3,6 +3,7 @@
 
 .vim
 .env
+.venv
```
Narsil, line 6: No. That's a user thing it does not belong here.

Please use a global `.gitignore` for you (then you also only define it once on your machine).

tokenizers/Cargo.toml
```diff
@@ -59,11 +58,18 @@ cached-path = { version = "0.5", optional = true }
 aho-corasick = "0.7"
 paste = "1.0.6"
 proc_macros = { path = "./src/utils/proc_macros" }
+once_cell = "1.8"
+cfg-if = "1"
```
Narsil, line 62: I don't think that's needed actually.

You can probably write this as pure rust compilation `#cfg`.

tokenizers/src/utils/regex.rs
```diff
@@ -0,0 +1,497 @@
...
+impl Hash for Regex {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.regex_str.hash(state);
+    }
+}
+
+impl Regex {
+    /// Create a new regex from the pattern string.
+    ///
+    /// Note that the regex compilation happens on first use, which is why this method does not
```
Narsil, line 59: I really think this library wants to compile the regex on load, not on-use, that changes this to a Result.

It also probably enables your to pass `&str`  since the Regex compilation will create the object without making the copy and doesn't need to keep a reference to the string, (so no lifetime issues here it seems).

tokenizers/Cargo.toml
```diff
@@ -59,11 +58,18 @@ cached-path = { version = "0.5", optional = true }
 aho-corasick = "0.7"
 paste = "1.0.6"
 proc_macros = { path = "./src/utils/proc_macros" }
+once_cell = "1.8"
```
Narsil, line 61: I don't think we need that either.

Even keeping the idea of `fancy-regex`, a simple enum is OK.
And even then, we **don't** want an enum. We want something that uses the correct regex engine ( `onig` or `onig-rs`) directly at compilation time.

tokenizers/src/utils/regex.rs
```diff
@@ -0,0 +1,497 @@
+// All code and comments below this first line are copied from here (with some edits): https://github.com/bminixhofer/nlprule/blob/main/nlprule/src/utils/regex.rs
```
Narsil, line 1: Yes, overall refrain from doing things this way.

Copy pasting code from another repo is likely to be the wrong solution.
They have a different context, different goals, you can't just reuse and hope it fits the current projects needs.

Taking an inspiration from other sources is GREAT, and you should definitely continuing looking at other repos for ideas on how to handle problems.

But you cannot blindly apply things, you need to understand the current project scopes, philosophy, concerns.

Rule of thumb:
- Less changes is better
- Imitate the style of previous code is better

Narsil (commented):
Hi since this is your first contribution, and even if I don't approve of just using a *different* regex engine, I did review it so you can have in mind what this projects concerns are.

Cheers !

josephrocca:
@Narsil Thanks for the detailed review! It was helpful. I'll close this per your comment [here](https://github.com/huggingface/tokenizers/issues/63#issuecomment-1041309647).

PR #913, opened 2022-02-18
Propose another regex for ByteLevel based on whitespace

thomasw21:


bindings/python/src/pre_tokenizers.rs
```diff
@@ -229,7 +230,7 @@ macro_rules! setter {
 ///         Whether to add a space to the first word if there isn't already one. This
 ///         lets us treat `hello` exactly like `say hello`.
 #[pyclass(extends=PyPreTokenizer, module = "tokenizers.pre_tokenizers", name=ByteLevel)]
-#[text_signature = "(self, add_prefix_space=True)"]
+#[text_signature = "(self, add_prefix_space=True, trim_offsets=True, regex_type=\"original\")"]
```
Narsil, line 233: I think we should remove `trim_offsets`  since we're not actually parsing it nor feeding to ByteLevel object here (and it's not used by `pre_tokenizer` so it's actually better to not mention it here.

tokenizers/src/pre_tokenizers/byte_level.rs
```diff
@@ -87,14 +99,27 @@ impl ByteLevel {
...
         self
     }
+
+    #[must_use]
+    pub fn regex_type(mut self, v: RegexType) -> Self {
+        self.regex_type = v;
+        self
+    }
+
+    pub fn regex(&self) -> &Regex {
```
Narsil, line 109: That shouldn't be `pub` IMO.

Narsil (commented):
Seems ok to me:

- Still needs Serialization tests (+ backward deserialization tests)
- Node support (I can work on that)

thomasw21:
Yeah just trying to merge on `bigscience-fork` right now. I understand those are necessary if we want to merge to master.

PR #916, opened 2022-02-22
in serialization.rs, the supplementary tokens are now added "in batch…

### LINKED ISSUE #914: Loading of Tokenizer is really slow when there are lots of additional tokens
I noticed this issue in the context of this [other issue](https://github.com/huggingface/transformers/issues/15733) reported on the 🤗 Transformers Github.

When you load a fast tokenizer from a `tokenizer.json` which contains lots of tokens (125936 in my case), it takes hours to loading.

That happens for both the slow and fast tokenizer - given that, in this respect, they behave in the very same way.

Indeed, [here](https://github.com/huggingface/tokenizers/blob/2fecdc10dd2792a2f033a27e510ac828c744d619/tokenizers/src/tokenizer/serialization.rs#L155) you can see that the code loads the tokens one at time - because it checks, after having added each token, that everything is ok.

However, adding a token at the time is really inefficient: in fact, that implies that the trie of the tokenizer is rebuilt each time, and this is the root cause of taking literally hours to load the tokenizer with my 125936 tokens.

I have two points:
1. I need a "quick&dirty" fix in order to have my tokenizer loaded and complete my task: the fix is trivial - just load all the tokens in one shot and - temporary - avoid the "sanity check" of the tokenizer status; however, I am totally new to Rust, and I cannot inderstand how I am supposed to compile everything and, let's say, "hot patch" my 🤗 Transformers installation with the new library; I run cargo build, I found the `tokenizers.d` and `libtokenizers.rlib` (I assume that these are the libs that eventually are somehow embedded in the transformers ), but I cannot find them in my Transformers installation: could you point me at some documenation that let me understand how the tokenizers libs are eventually bundled with the Transformers package?
2. I would like to propose a patch for this code, but - as you can infer from above notes - I didn't found any documentation that explains how to properly build the project and run the tests: could you please point me at the proper documentation?

PiercarloSlavazza:
…" to the tokenizer vocabulary, so that the tokenizer trie is built just once (Fix #914)

Building the trie is a very expensive operation: previously this operation was carried out for every and each token, so that, for large vocabularies, the overall loading time of the vocabulary resulted unacceptable.

### DIFF
--- a/tokenizers/examples/serialization.rs
+++ b/tokenizers/examples/serialization.rs
@@ -2,13 +2,15 @@ use tokenizers::models::wordpiece::WordPiece;
 use tokenizers::{AddedToken, Tokenizer};
 
 fn main() {
+    let start = std::time::Instant::now();
     let mut tokenizer = Tokenizer::new(WordPiece::default());
 
-    let tokens: Vec<_> = (0..4000)
+    let tokens: Vec<_> = (0..120_000)
         .map(|i| AddedToken::from(format!("[SPECIAL_{}]", i), false))
         .collect();
     tokenizer.add_tokens(&tokens);
     tokenizer.save("_tok.json", false).unwrap();
+    println!("Save took {:?}", start.elapsed());
     let start = std::time::Instant::now();
     let _tok = Tokenizer::from_file("_tok.json").unwrap();
     println!("Took {:?}", start.elapsed());
--- a/tokenizers/src/tokenizer/serialization.rs
+++ b/tokenizers/src/tokenizer/serialization.rs
@@ -150,34 +150,17 @@ where
             .build()
             .map_err(|e| V::Error::custom(e.to_string()))?;
 
+        let mut regular_tokens = vec![];
+        let mut special_tokens = vec![];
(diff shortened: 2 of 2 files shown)


Narsil:
Hi @PiercarloSlavazza , I managed to get a reproduicible script:
```
cargo run --example serialization --release
```

with `120k` added tokens.

I took the liberty of making minor changes to your code, 

- Prevent iterating 3 times over the amount of tokens
- Remove all `.clone()`. It most likely was optimized away before, but it's even better if we don't have to count on that.

Does this code work ? In your issue you seem to encounter another problem with extra `\n` in your special tokens, do you mind sharing something to reproduce (it shouldn't be done by code in this lib, but there could be a bug).

Narsil:
Hi @PiercarloSlavazza ,

I'll merge this to include in the release. (0.11.6)

PiercarloSlavazza:
Sorry for the late reply - the patch seems good, thank you 👍🏻

Narsil:
No worries, it's not that late, it's just that I am trying to squeeze the release before `transformers` 4.17.0 which will happen next week, and I would prefer to have a bit of time to make sure all the tests are running and all master users report any other undesired effects

Issue #918, opened 2022-02-23
Bug: Mask token is not tokenized correctly since v0.11.3


saattrupdan:
Since (Python) version 0.11.3, `tokenizers` are not dealing with mask tokens as single tokens, and instead split them up into characters.

## Minimal example
```python
from tokenizers import tokenizers, models, AddedToken, trainers

# Initialize tokenizer
model = models.BPE(unk_token='<unk>')
tokenizer = tokenizers.Tokenizer(model)

# Add special tokens
special_tokens = [
    AddedToken('<mask>', single_word=True, normalized=False),
    AddedToken('<unk>', single_word=True, normalized=False)
]
tokenizer.add_special_tokens(special_tokens)

# Add a white space pre-tokenizer
tokenizer.pre_tokenizer = pre_tokenizers.Whitespace()

# Initialise trainer
trainer = trainers.BpeTrainer(vocab_size=1000, special_tokens=special_tokens)

# Load corpus
corpus = ['Dette er en test']

# Train on dummy corpus
tokenizer.train_from_iterator(iterator=corpus, trainer=trainer)

# Test <mask> encoding
print(tokenizer.encode('Dette er <mask> test').tokens)
print(tokenizer.encode('<mask> test').tokens)
print(tokenizer.encode('<mask> ').tokens)
print(tokenizer.encode('Dette er <mask>').tokens)
print(tokenizer.encode('Dette er <mask> ').tokens)
print(tokenizer.encode('<mask>').tokens)
```

Correct output with `tokenizers==0.11.2`:
```
['Dette', 'er', '<mask>', 'test']
['<mask>', 'test']
['<mask>']
['Dette', 'er', '<mask>']
['Dette', 'er', '<mask>']
['<mask>']
```

Wrong output with `tokenizers` versions 0.11.3, 0.11.4 and the current latest version, 0.11.5:
```
['Dette', 'er', '<unk>', '<unk>', '<unk>', 's', '<unk>', '<unk>', 'test']
['<unk>', '<unk>', '<unk>', 's', '<unk>', '<unk>', 'test']
['<unk>', '<unk>', '<unk>', 's', '<unk>', '<unk>']
['Dette', 'er', '<mask>']
['Dette', 'er', '<unk>', '<unk>', '<unk>', 's', '<unk>', '<unk>']
['<mask>']
```

As can be seen in the above tests, the bug does not happen when the document ends with the mask token (but even a whitespace after is enough to trigger the bug).

[...]

PR #919, opened 2022-02-23
Fixing `single_word` AddedToken.

Narsil:
Fixes https://github.com/huggingface/tokenizers/issues/918

### DIFF
--- a/tokenizers/src/tokenizer/added_vocabulary.rs
+++ b/tokenizers/src/tokenizer/added_vocabulary.rs
@@ -111,6 +111,9 @@ fn space_leftmost_at_end(sentence: &str) -> usize {
         match_.start()
     } else {
         // This should never happen since the Regex should match all the time
+        // It didn't match any space, so we can safely return sentence.len() to be the leftmot space and the
+        // end
+        warn!("Warning in leftmost_at_end code path, we expect the regex to match always and it didn't, Please report this.");
         sentence.len()
     }
 }
@@ -119,6 +122,9 @@ fn space_rightmost_at_start(sentence: &str) -> usize {
         match_.end()
     } else {
         // This should never happen since the Regex should match all the time
+        // It didn't match any space, so we can safely return 0 to be the rightmost space and the
+        // start
+        warn!("Warning in leftmost_at_end code path, we expect the regex to match always and it didn't, Please report this.");
         0
     }
 }
--- a/tokenizers/src/tokenizer/added_vocabulary.rs
+++ b/tokenizers/src/tokenizer/added_vocabulary.rs
@@ -304,7 +304,7 @@ impl AddedVocabulary {
             if added_token.single_word {
                 let start_space = start == 0 || sentence.as_bytes().get(start - 1) == Some(&b' ');
                 let stop_space =
-                    stop == sentence.len() || sentence.as_bytes().get(stop + 1) == Some(&b' ');
+                    stop == sentence.len() || sentence.as_bytes().get(stop) == Some(&b' ');
 
                 if !stop_space || !start_space {
                     // Discard not single word
@@ -697,5 +697,44 @@ mod tests {
         let vocab = AddedVocabulary::new();
         let matches = vocab.find_matches("", &vocab.split_trie);
         assert_eq!(matches, vec![(None, (0, 0))]);
+    }
+
+    #[test]
+    fn test_single_word_is_correct() {
+        // Is able to extract both normal and special tokens, with various options (lstrip, rstrip,
+        // single_word, normalized)
+        let model = ModelMock::new(&[]);
+        let mut vocab = AddedVocabulary::new();
+        let normalizer = Lowercase;
+
+        vocab.add_tokens(
+            &[AddedToken::from("<mask>", false).single_word(true)],
+            &model,
+            Some(&normalizer),
+        );
+        // Left, in the middle, non single world left, non single word right, end of sentence valid
+        let result = vocab.extract_and_normalize(
+            Some(&normalizer),
+            "<mask> My name <mask> A<mask> <mask>ony <mask>",
+        );
+        assert_eq!(
+            result
+                .get_splits(OffsetReferential::Original, OffsetType::Byte)
+                .into_iter()
+                .map(|(s, _, tokens)| (
+                    s,
+                    tokens
+                        .as_ref()
+                        .map(|t| t.iter().map(|t| t.id).collect::<Vec<_>>())
+                ))
+                .collect::<Vec<_>>(),
+            vec![
+                ("<mask>", Some(vec![0])),
+                (" my name ", None),
+                ("<mask>", Some(vec![0])),
+                (" a<mask> <mask>ony ", None),
+                ("<mask>", Some(vec![0]))
+            ]
+        );
     }
 }
--- a/tokenizers/src/tokenizer/added_vocabulary.rs
+++ b/tokenizers/src/tokenizer/added_vocabulary.rs
@@ -2,6 +2,7 @@ use super::{
     normalizer::Range, Model, NormalizedString, Normalizer, Offsets, PreTokenizedString, Token,
 };
 use aho_corasick::{AhoCorasick, AhoCorasickBuilder, MatchKind};
+use regex::Regex;
 use serde::{ser::SerializeSeq, Deserialize, Serialize, Serializer};
 use std::collections::{HashMap, HashSet};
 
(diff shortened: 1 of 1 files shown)


mishig25 (approved):

tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -698,4 +698,43 @@ mod tests {
...
+
+        vocab.add_tokens(
+            &[AddedToken::from("<mask>", false).single_word(true)],
+            &model,
+            Some(&normalizer),
+        );
+        // Left, in the middle, non single world left, non single word right, end of sentence valid
+        let result = vocab.extract_and_normalize(
+            Some(&normalizer),
+            "<mask> My name <mask> A<mask> <mask>ony <mask>",
```
mishig25, line 718: nice test case!

tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -304,7 +304,7 @@ impl AddedVocabulary {
             if added_token.single_word {
                 let start_space = start == 0 || sentence.as_bytes().get(start - 1) == Some(&b' ');
```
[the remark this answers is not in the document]
Narsil: Start and stop are offsets in the strings.

"Hi"

offset 0 -> to the left of H
offset 1 -> between H and i
offset 2 -> right of i

char 0 -> H
char 1 -> i

If the capture is H let's say, then, start=0, offset=1
Then, start == 0 and char 0 = H, so the previous char would indeed be start - 1
And next char  is char 1, so next char == stop_offset indeed (and that's also why it's stop == sentence.len() where it would usually be stop == sentence.len() - 1

McPatate (dismissed):

Narsil:
Sorry, re requesting your reviews since the breaking change was actually much bigger than that.

The old code used regexp with `\b` to match word boundaries and `\s*` to match spaces.
The rust `regex` crate uses by default the full unicode definition of those things.

\w is a word, and \b is word boundary (\w followed by \W or end of string roughly).
I chose to use `\w` in the current form because using `\b` felt less clear to me, but we could change that.

`\s` should be clear in this state.

tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -90,6 +91,37 @@ impl std::cmp::Eq for AddedToken {}
...
+    }
+}
+fn space_rightmost_at_start(sentence: &str) -> usize {
+    if let Some(match_) = RIGHTMOST_SPACE_AT_START.find(sentence) {
+        match_.end()
+    } else {
+        // This should never happen since the Regex should match all the time
+        0
+    }
+}
```
McPatate, line 124: Wouldn't it be better to return a `Result` here ?

tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -302,32 +334,26 @@ impl AddedVocabulary {
             let id = split_re.1[aho_id];
             let added_token = &self.added_tokens_map_r.get(&id).unwrap();
             if added_token.single_word {
-                let start_space = start == 0 || sentence.as_bytes().get(start - 1) == Some(&b' ');
-                let stop_space =
-                    stop == sentence.len() || sentence.as_bytes().get(stop + 1) == Some(&b' ');
+                let start_space = start == 0 || !ends_with_word(&sentence[..start]);
```
McPatate, line 337: That is indeed cleaner 👌🏻

McPatate (approved):
LGTM

tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -90,6 +91,37 @@ impl std::cmp::Eq for AddedToken {}
...
+    STARTS_WITH_WORD.is_match(sentence)
+}
+
+fn space_leftmost_at_end(sentence: &str) -> usize {
+    if let Some(match_) = LEFTMOST_SPACE_AT_END.find(sentence) {
+        match_.start()
+    } else {
+        // This should never happen since the Regex should match all the time
+        sentence.len()
+    }
```
mishig25, line 115: if it is `never happen`, can we just call `LEFTMOST_SPACE_AT_END.find(sentence).expect("LEFTMOST_SPACE_AT_END should always match but didnt match int this sentence: {sentence}").`

tokenizers/src/tokenizer/added_vocabulary.rs
Narsil: Well if the regexp doesn't match, it's strictly equivalent to matching nothing, right ?

I do that to avoid adding and `unwrap` which could potential `panic!` which makes users pretty unhappy.

I don't think the pathway should be taken, but if it ever is, then I think `0` is a valid default which prevents panicking and is still, correct. Wdyt ?

tokenizers/src/tokenizer/added_vocabulary.rs
Narsil: `.unwrap()` is something I tend to avoid within logic code at all if possible, since having the program panic, is never a great user experience.

My answer : https://github.com/huggingface/tokenizers/pull/919#discussion_r813862919 applies too. I think returning `Option` would be preferrable to panicking, but the result would be the same IMO, if the regex doesn't match, I just don't capture anything.

tokenizers/src/tokenizer/added_vocabulary.rs
McPatate: I understand adding `Result`s all the way up to the public API is cumbersome but to me is the cleaner approach : notify the user there was an error / undefined / unexpected behavior and let him handle it.

Returning 0 is hiding the fact that you've run into an unexpected case/behavior (aka silent error). Adding comments to clarify code is a smell imo, make it comprehensible in the first place, you almost always can !

But then again, it's more work.

Narsil: > I understand adding Results all the way up to the public API is cumbersome but to me is the cleaner approach : notify the user there was an error / undefined / unexpected behavior and let him handle it.

In that case it's not true. Finding 0 match is different than an error.
I could very well change the regex to be `\s+` and then the `else` branch would be expected and working as intended.

Is adding a `warn!("AddedToken with `single_word` seems to have an issue, Please report this")` in that code path OK ?
It it warning the user, but still prevent catastrophing failure ?

We don't expect the code path to be taken, but it's not preventing the algorithm from working.

mishig25 (approved):

tokenizers/src/tokenizer/added_vocabulary.rs
mishig25: oh I see 🙂

tokenizers/src/tokenizer/added_vocabulary.rs
McPatate: Am I correct in assuming that if we go through the `else` clause, it means the user has inputted a malformatted `.lstrip` or `.rstrip` token ?

Narsil: No, the else clause will trigger, when `REGEXP.find()` returns `None`.

This is normal when the REGEX fails to find any match in the submitted string.
This particular brand of REGEX should match all the time (since it should match the empty string being `^\s*` and `\s*$`).
So I don't expect it to not match. But if it doesn't match, it's roughly the same as saying no space where found.

McPatate: But that's what I don't understand, shouldn't there be a space ?

Narsil: Oh no no necessarily.

`AddedTokens("<mask>", lstrip=True)` means you want it to match "Something <mask>  else" -> ['Something", " <mask>", " else"]. Meaning you are actually capturing the left spaces in addition to your token (effectively tripping it from being seen for your model).

But "Something<mask>else" will also capture ["Something", "<mask>", "else"]. no space have been deleted.
If you want this to NOT capture you need to activate `single_word=True` too.

But both options are orthogonal and don´t  necessarily need one another (although I think in practice they are probably reasonned about in conjunction)

Narsil:
I'll merge this so there's time for a release today.

tokenizers/src/tokenizer/added_vocabulary.rs
McPatate: Ok that makes sense. One thing tho : I would remove the comment, imho it's misleading !

Narsil: @McPatate I merged since I am starting the necessary work to release this before transformers's own release on thursday, but we can continue the discussion to make this clear.

PR #920, opened 2022-02-24
Attempt to make Unigram trainer parallel.

Narsil:


chris-ha458:
I feel like #976 has resolved both #921 and #920.

They all touch on the same processes of the training (e-m part) and I am under the impression that @mishig25 did a good job in #976 of referring to both the previous PRs and incorporating to the now merged #976.
Besides, with such a major rewrite of this part of the codes, even if there are some aspects that might still be useful from these PRs they should rather be rewritten in my opinion.

As such, I feel like both #920 and #921 can be safely closed without further action.

Does anybody feel like they would be useful to be kept open? Are there any ideas that these PRs explore that havent been handled by #976?
Are there any outstanding adjustments that these

PR #921, opened 2022-02-24
Attempt to make unigram faster 2.

thomasw21:


Narsil:
Nice ! Still a few lints :)

tokenizers/src/models/unigram/trainer.rs
```diff
@@ -293,16 +293,49 @@ impl UnigramTrainer {
...
+        use rayon::current_num_threads;
+        use rayon::iter::IndexedParallelIterator;
+        use rayon::iter::IntoParallelRefIterator;
+        use rayon::iter::ParallelIterator;
+
+        let chunk_size = std::cmp::max(sentences.len() / current_num_threads(), 1);
+        let collected: Vec<(f64, Vec<f64>, Vec<Vec<usize>>)> = sentences
+            .par_iter()
+            .enumerate()
+            .chunks(chunk_size)
```
Narsil, line 306: Hm, Just read that.

I think you really want `par_chunks` and not `par_iter() + chunks()` since I don't expect it to do the same thing.
Doesn't this reallocate a lot ?

You can recover the ids by using this kind of logic:
https://github.com/google/sentencepiece/blob/master/src/unigram_model_trainer.cc#L337

Not sure the cleanest way to handle this in `rayon` ecosystem but `sentences.iter().enumerate().par_chunks` seems natural and like it should work.

thomasw21: Tried 28f98c1f6c0a732ae2a867a09b5b2f300e8f3d00 let me know what you think?

tokenizers/src/models/unigram/trainer.rs
```diff
@@ -425,22 +454,22 @@ impl UnigramTrainer {
...
+                || (0.0, 0, vec![0.0; model.len()]),
+                |(objs, ntokens, expected), (lobjs, lntokens, lexpected)| {
+                    (
+                        objs + lobjs,
+                        ntokens + lntokens,
+                        expected
+                            .iter()
+                            .zip(lexpected)
+                            .map(|(global_el, local_el)| global_el + local_el)
+                            .collect(),
```
Narsil, line 467: `iter_mut()` + `for_each` makes sure there are not reallocations.

`.collect()` is like `.clone()`, avoid if possible.

It might be optimized away but I would rather not count on it.

thomasw21: Hum but you need to set that variable to `mut` no?

Narsil: Probably, but if the reduce is not itself parallel, then it's ok to have `global` be mutable.

PR #923, opened 2022-02-25
Use Self where possible, some minor refactoring

adamnemecek:


Narsil:
@adamnemecek thanks for this. This seems a nice little lint. Does any software detect those ?

Narsil:
Ahaa the test seem to point that it wasn't software but a full search+replace. Was it ?

adamnemecek:
I'm not aware. It was done with a text editor. I have pushed an updated version that fixed the failing test. However there were some failing tests even before I made my changes.

Narsil:
It seems to be in the nursery, so it'll come sooner or later to `clippy`: https://rust-lang.github.io/rust-clippy/master/#use_self

Your editor probably just includes those too.

Thanks for the fixes !

tokenizers/src/pre_tokenizers/unicode_scripts/scripts.rs
```diff
@@ -144,1934 +144,1940 @@ pub enum Script {
...
-        0x17000..=0x187EC => Script::Tangut,
-        0x18800..=0x18AF2 => Script::Tangut,
-        _ => Script::Any,
+impl Script {
+    pub fn new(c: char) -> Self {
+        match c as u32 {
+            0x0000..=0x001F => Self::Common,
+            0x0020 => Self::Common,
+            0x0021..=0x0023 => Self::Common,
+            0x0024 => Self::Common,
```
Narsil: This is an arguably good change, but it does NOT belong in this PR.

tokenizers/src/tokenizer/encoding.rs
```diff
@@ -110,11 +110,7 @@ impl Encoding {
 
     /// Return the number of sequences combined in this Encoding
     pub fn n_sequences(&self) -> usize {
-        if self.sequence_ranges.is_empty() {
-            1
-        } else {
-            self.sequence_ranges.len()
-        }
+        self.sequence_ranges.len().max(1)
```
Narsil, line 113: Same, I would rather not include this in this PR.

The `Whatever` -> `Self` is a pretty big change that touches most of the code base.
We should keep everything that might have other side effects OUT so that it's easier to find the problem later.

On this particular one, I am not particularly fond of the final result, `std::cmp::max(1, self.sequences_ranges.len())` feels much more natural (should we make a change).

Narsil (commented):
Thank you for those changes.

Do you mind creating another PR for the two seemingly unrelated changes?

I think they're ok but I would prefer to not mix those changes if possible (I can do the split if you want).

tokenizers/src/tokenizer/encoding.rs
adamnemecek: Done.

Narsil (approved):
LGTM

Issue #932, opened 2022-03-01
Prebuilding `tokenizers` for Mac M1.


Narsil:
We're at this point waiting for GH actions to support M1 runners.

Any help to do it differently is appreciated, but for now we will probably resolve to manual builds on a M1 computer we have as cross compilation is rendered hard because of both the CPP dependencies we have, and the PyO3 compatibility requirements (which would require us to cross compile Python too).

McPatate:
Hey, I've added a few binaries for `0.11.6`. Feedback is welcome to see if everything works as expected !

itchingpixels:
@McPatate : It works, thank you! would it be possible to automatically add it to all subsequent releases?

McPatate:
We're waiting for the M1 workers to add it to our CI.

Will build manually for the time being as it's less hassle.

moinnadeem:
Any luck doing this automatically? I think the latest release also requires Rust to build tokenizers, since prebuilt wheels for the M1 aren't available.

PR #936, opened 2022-03-01
Feat/m1 manual build

McPatate:
#932

### DIFF
--- a/bindings/python/.cargo/config.toml
+++ b/bindings/python/.cargo/config.toml
@@ -1,0 +1,14 @@
+[target.x86_64-apple-darwin]
+rustflags = [
+  "-C", "link-arg=-undefined",
+  "-C", "link-arg=dynamic_lookup",
+  "-C", "link-arg=-mmacosx-version-min=10.11",
+]
+
+[target.aarch64-apple-darwin]
+rustflags = [
+  "-C", "link-arg=-undefined",
+  "-C", "link-arg=dynamic_lookup",
+  "-C", "link-arg=-mmacosx-version-min=10.11",
+]
+
--- a/bindings/python/tokenizers_manual_m1_build.sh
+++ b/bindings/python/tokenizers_manual_m1_build.sh
@@ -1,0 +1,8 @@
+#! /bin/bash
+
+MACOSX_DEPLOYMENT_TARGET=10.11 SDKROOT="/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk" CFLAGS="-I/usr/include/openssl -I/usr/local/opt/readline/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include" CPPFLAGS="-I/usr/local/opt/zlib/include" LDFLAGS="-L/usr/lib -L/usr/local/opt/readline/lib" pyenv install <wtv_version>
+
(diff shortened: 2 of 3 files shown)


bindings/python/tokenizers_manual_m1_build.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+
+MACOSX_DEPLOYMENT_TARGET=10.11 SDKROOT="/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk" CFLAGS="-I/usr/include/openssl -I/usr/local/opt/readline/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include" CPPFLAGS="-I/usr/local/opt/zlib/include" LDFLAGS="-L/usr/lib -L/usr/local/opt/readline/lib" pyenv install <wtv_version>
+
+~/.pyenv/versions/<wtv_version>/bin/pip install setuptools wheel setuptools-rust==0.11.3 --ignore-installed --force-reinstall
+
+MACOSX_DEPLOYMENT_TARGET=10.11 ~/.pyenv/versions/<wtv_version>/bin/python setup.py bdist_wheel
+
```
Narsil, line 8:
```suggestion
for VARIABLE in "3.7.12" "3.8.12" "3.9.10" "3.10.2"
do
    MACOSX_DEPLOYMENT_TARGET=10.11 SDKROOT="/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk" CFLAGS="-I/usr/include/openssl -I/usr/local/opt/readline/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include" CPPFLAGS="-I/usr/local/opt/zlib/include" LDFLAGS="-L/usr/lib -L/usr/local/opt/readline/lib" pyenv install $VARIABLE

    ~/.pyenv/versions/$VARIABLE/bin/pip install setuptools wheel setuptools-rust==0.11.3 --ignore-installed --force-reinstall

    MACOSX_DEPLOYMENT_TARGET=10.11 ~/.pyenv/versions/$VARIABLE/bin/python setup.py bdist_wheel
done
```

Is that OK ? (Fix the library versions if you want)

bindings/python/.cargo/config.toml
```diff
@@ -0,0 +1,14 @@
+[target.x86_64-apple-darwin]
+rustflags = [
+  "-C", "link-arg=-undefined",
+  "-C", "link-arg=dynamic_lookup",
+  "-C", "link-arg=-mmacosx-version-min=10.11",
```
Narsil, line 5: Do we still have a link where this is explained ? I can only find the `MACOSX_DEPLOYMENT_TARGET` with my google fu.

(It's not super important)

Narsil (commented):
Thank you very much !

Narsil (approved):

bindings/python/.cargo/config.toml
McPatate: It comes from this thread :
https://users.rust-lang.org/t/compile-rust-binary-for-older-versions-of-mac-osx/38695

McPatate: I think we should leave that here for now. I'll test on next release without that flag to see if it still works.

PR #938, opened 2022-03-04
Changing `Decoder` trait to be more composable.

### LINKED ISSUE #872: Add an `Sequence` object to the decoders
I wonder if it would be useful to have a sequence object for the decoders too.

It seems to me for example that if we build a tokenizer with a BPE model that defines a `end_of_word_suffix`, we will need to use the `BPEDecoder` decoder to replace the`end_of_word_suffix` and if we also used a ByteLevel pre-tokenization we will need the `ByteLevel` decoder to  realign the codes.

At the moment, it seems to me that we don't have a solution to choose a suitable decoder for such a tokenizer.

What do you think? :smile:

Narsil:
Fix #872

### DIFF
--- a/bindings/node/lib/bindings/decoders.test.ts
+++ b/bindings/node/lib/bindings/decoders.test.ts
@@ -12,7 +12,7 @@ describe("wordPieceDecoder", () => {
   it("can decode arrays of strings", () => {
     expect(
       wordPieceDecoder().decode(["Hel", "##lo", "there", "my", "fr", "##iend"])
-    ).toEqual("Hello there my friend");
+    ).toEqual(["Hel", "lo", " there", " my", " fr", "iend"]);
   });
 });
 
@@ -39,6 +39,6 @@ describe("ctcDecoder", () => {
   it("encodes correctly", () => {
     expect(
       ctcDecoder().decode(["<pad>", "h", "h", "e", "e", "l", "l", "<pad>", "l", "l", "o"])
-    ).toEqual("hello");
+    ).toEqual(["h", "e", "l", "l", "o"]);
   });
 });
--- a/bindings/python/src/decoders.rs
+++ b/bindings/python/src/decoders.rs
@@ -342,7 +342,7 @@ impl Decoder for CustomDecoder {
             let decoded = self
                 .inner
                 .call_method(py, "decode", (tokens,), None)?
-                .extract::<_>(py)?;
+                .extract(py)?;
             Ok(decoded)
         })
     }
--- a/bindings/python/src/decoders.rs
+++ b/bindings/python/src/decoders.rs
@@ -51,7 +51,7 @@ impl PyDecoder {
 }
 
 impl Decoder for PyDecoder {
-    fn decode(&self, tokens: Vec<String>) -> tk::Result<String> {
+    fn decode(&self, tokens: Vec<String>) -> tk::Result<Vec<String>> {
         self.decoder.decode(tokens)
     }
 }
@@ -98,7 +98,7 @@ impl PyDecoder {
     /// Returns:
     ///     :obj:`str`: The decoded string
     #[text_signature = "(self, tokens)"]
-    fn decode(&self, tokens: Vec<String>) -> PyResult<String> {
+    fn decode(&self, tokens: Vec<String>) -> PyResult<Vec<String>> {
         ToPyResult(self.decoder.decode(tokens)).into()
     }
 }
@@ -337,12 +337,12 @@ impl CustomDecoder {
 }
 
 impl Decoder for CustomDecoder {
-    fn decode(&self, tokens: Vec<String>) -> tk::Result<String> {
+    fn decode(&self, tokens: Vec<String>) -> tk::Result<Vec<String>> {
         Python::with_gil(|py| {
             let decoded = self
                 .inner
                 .call_method(py, "decode", (tokens,), None)?
-                .extract::<String>(py)?;
+                .extract::<_>(py)?;
             Ok(decoded)
         })
     }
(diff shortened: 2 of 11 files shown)


tokenizers/src/decoders/ctc.rs
```diff
@@ -97,7 +108,13 @@ mod tests {
...
         assert_eq!(
             ctc_decoder.decode(id_to_string_result).unwrap(),
-            "HIS INSTANCT PANIC WAS FOLLOWED BY A SMALL SHARP BLOW HIGH ON HIS CHEST ".to_string()
+            vec![
+                "H", "I", "S", " ", "I", "N", "S", "T", "A", "N", "C", "T", " ", "P", "A", "N",
```
McPatate, line 112: Instanct ?

tokenizers/src/decoders/ctc.rs
```diff
@@ -97,7 +108,13 @@ mod tests {
...
+                "I", "C", " ", "W", "A", "S", " ", "F", "O", "L", "L", "O", "W", "E", "D", " ",
+                "B", "Y", " ", "A", " ", "S", "M", "A", "L", "L", " ", "S", "H", "A", "R", "P",
+                " ", "B", "L", "O", "W", " ", "H", "I", "G", "H", " ", "O", "N", " ", "H", "I",
+                "S", " ", "C", "H", "E", "S", "T", " "
```
McPatate, line 116: Is the space at the end normal ?

tokenizers/src/decoders/wordpiece.rs
```diff
@@ -44,12 +44,21 @@ pub fn cleanup(dirty_input: String) -> String {
...
+            .iter_mut()
+            .enumerate()
+            .map(|(i, token)| {
+                if token.starts_with(&self.prefix) {
+                    *token = token.replacen(&self.prefix, "", 1);
+                } else if i != 0 {
+                    *token = format!(" {}", token);
+                }
+                if self.cleanup {
+                    *token = cleanup(token.clone());
```
McPatate, line 58: Isn't better API-wise to pass in a reference, `&token` and call a `.to_owned()` or `.clone()` inside the `cleanup()` method ?
Feels cleaner, the responsibility of the clone is imo `cleanup`'s.

McPatate (approved):
Few questions for comprehension.

Looks good overall !

bindings/node/lib/bindings/decoders.test.ts
```diff
@@ -39,6 +39,6 @@ describe("ctcDecoder", () => {
   it("encodes correctly", () => {
     expect(
       ctcDecoder().decode(["<pad>", "h", "h", "e", "e", "l", "l", "<pad>", "l", "l", "o"])
-    ).toEqual("hello");
+    ).toEqual(["h", "e", "l", "l", "o"]);
```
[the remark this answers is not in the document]
Narsil: It dedpuplicates earlier, `<pad>` acts as a separator to  enable actually multiple letters to be outputted.

bindings/python/src/decoders.rs
```diff
@@ -337,12 +337,12 @@ impl CustomDecoder {
...
 
 impl Decoder for CustomDecoder {
-    fn decode(&self, tokens: Vec<String>) -> tk::Result<String> {
+    fn decode(&self, tokens: Vec<String>) -> tk::Result<Vec<String>> {
         Python::with_gil(|py| {
             let decoded = self
                 .inner
                 .call_method(py, "decode", (tokens,), None)?
-                .extract::<String>(py)?;
+                .extract::<_>(py)?;
```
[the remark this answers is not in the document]
Narsil: Well it's evidently `Vec<String>` from the return type.

I am not sure why I just didn't remove the turbofish altogether.
the other methods also just use pure `.into()` and `.extract` without turbofish.

tokenizers/src/decoders/ctc.rs
Narsil: It's the original test, I am pretty sure it just comes from the output of some wav2vec2 which actually had issues.

tokenizers/src/decoders/ctc.rs
Narsil: Yes, it was in the original test too. 

In `transformers` it's removed at the end, but the model really outputs it.

The token is actually called `word_delimiter` we just happen to use a space on space separated languages :)

tokenizers/src/decoders/wordpiece.rs
Narsil: I think it's almost always preferable to not clone on a user's behalf and let the user decide how best to pass owned instances (if you really require ownership).

It seems maybe this function doesn't actually require ownership, I'll change to reflect that.

tokenizers/src/decoders/ctc.rs
McPatate: Aah legacy :D

tokenizers/src/decoders/wordpiece.rs
McPatate: I see what you mean. W/out the clone, you will have to create a new instance of a `String`, which is somehow similar to cloning ^^

SaulLu:
I'm sorry, I didn't have time to do a proper review. Nevertheless, I would be very interested to understand a bit better the changes that have been made :slightly_smiling_face: 

In particular, I see that the test target has changed (e.g.`.toEqual("hello");` to `.toEqual(["h", "e", "l", "l", "o"]);`): does this mean that the decode output will be different now?

Narsil:
No not the final thing, basically the output of the tokenizer will be `"".join(decoder.decode(tokens))` if you will.

Only breaking changes are if you are using decoders directly.

SaulLu:
Thank you very much for your answer! It's really great!

To put all the pieces together, am I right to understand that this last step is done here :smile: : https://github.com/huggingface/tokenizers/blob/2620531e9a360a4d3a010581fd28b7b1f0a5f3bb/tokenizers/src/tokenizer/mod.rs#L781

Narsil:
Indeed !  You see Rust is not that hard ^_^

tokenizers/src/decoders/wordpiece.rs
McPatate: If that is the rust way, so be it 😄

PR #940, opened 2022-03-05
Adding `Sequence` Decoder.

Narsil:
Fixes https://github.com/huggingface/tokenizers/issues/872

Goes after https://github.com/huggingface/tokenizers/pull/938

### DIFF
--- a/bindings/python/tests/bindings/test_decoders.py
+++ b/bindings/python/tests/bindings/test_decoders.py
@@ -2,7 +2,7 @@ import pytest
 import pickle
 import json
 
-from tokenizers.decoders import Decoder, ByteLevel, WordPiece, Metaspace, BPEDecoder, CTC
+from tokenizers.decoders import Decoder, ByteLevel, WordPiece, Metaspace, BPEDecoder, CTC, Sequence
 
 
 class TestByteLevel:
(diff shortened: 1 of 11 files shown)


bindings/python/tests/bindings/test_decoders.py
```diff
@@ -169,3 +169,27 @@ def test_can_modify(self):
...
+        assert isinstance(Sequence([]), Decoder)
+        assert isinstance(Sequence([]), Sequence)
+        serialized = pickle.dumps(Sequence([]))
+        assert isinstance(pickle.loads(serialized), Sequence)
+
+    def test_decoding(self):
+        decoder = Sequence([CTC(), Metaspace()])
+        initial = ["▁", "▁", "H", "H", "i", "i", "▁", "y", "o", "u"]
+        expected = [
+            "",
```
McPatate, line 187: Shouldn't empty strings be discarded ?

McPatate (approved):

bindings/python/tests/bindings/test_decoders.py
Narsil: It might, but I wouldn't do it pre-emptively. Simpler code is usually more correct, so I'll keep this as-is, but if it starts causing issues we can change this easily

PR #942, opened 2022-03-07
Python - add initial_alphabet to spm unigram trainer

dctelus:


bindings/python/py_src/tokenizers/implementations/sentencepiece_unigram.py
```diff
@@ -84,6 +84,7 @@ def train_from_iterator(
         vocab_size: int = 8000,
         show_progress: bool = True,
         special_tokens: List[Union[str, AddedToken]] = [],
+        initial_alphabet: List[str] = [],
```
Narsil, line 87: Python should never have mutable default arguments: https://docs.python-guide.org/writing/gotchas/

Using `Optional` with a None default is the correct way to handle this I think.

Narsil (commented):
Thank you for this, aside from the mutable default it's good.

bindings/python/py_src/tokenizers/implementations/sentencepiece_unigram.py
Narsil: And... `special_tokens` should be modified to (just one line above).

dctelus:
Nice to know, I didn't know!

In the rust impl, this is how it's extracted:
```
                    "initial_alphabet" => {
                        let alphabet: Vec<String> = val.extract()?;
                        builder = builder.initial_alphabet(
                            alphabet
                                .into_iter()
                                .filter_map(|s| s.chars().next())
                                .collect(),
                        );
                    }
```

I'm afraid if I pass in a `None` it will error out (and I am too unfamiliar with rust to attempt to fix it). I could do `initial_alphabet=[] if initial_alphabet is None else initial_alphabet`, but that seems overly verbose (and it's only passed to rust so I assume it's copied and can't be modified?). Let me know if that's what you want.

Narsil:
> Nice to know, I didn't know!

Yes, it's a very common gotcha, and actually because of Rust current implem it's sort of safe, but it might not always be the case.

> I'm afraid if I pass in a None it will error out (and I am too unfamiliar with rust to attempt to fix it). I could do initial_alphabet=[] if initial_alphabet is None else initial_alphabet, but that seems overly verbose (and it's only passed to rust so I assume it's copied and can't be modified?). Let me know if that's what you want.

Sorry I should have been more explicit (it's just GH suggestions don't really handle well multi block suggestions...)

```python

def train_from_iterator(
     ....
     initial_alphabet: Optional[List[str]] = None
    ....
):
     if initial_alphabet is None:
        initial_alphabet = []

     .... Rest is unchanged

     trainer(...
         initial_alphabet=initial_alphabet
         ...
      )
```
the gotcha is not triggered because now the list gets instantiated when the function is called (as opposed as when the file is read) to it's no longer a global reference variable and everything works as you assume it should)

Narsil (approved):
LGTM. Thank you very much !

Issue #947, opened 2022-03-09
Introspect python objects


cakiki:
I am currently doing something like this to get info about created normalizers and pretokenizers:

`eval(pre_tokenizer.__getstate__().decode('utf8'))` which returns the following dict: 

```
{'type': 'Sequence', 
'pretokenizers': [
                            {'type': 'WhitespaceSplit'}, 
                            {'type': 'Punctuation', 'behavior': 'Isolated'}
]
}
```

Is there a better way of doing this? Could this method ever fail?

Narsil:
Hi @cakiki ,

This shouldn't fail in the sense that's it's what is being used for pickling/unpickling. The actual values in there might change though.

And this is definitely not the easiest user experience, having something more pythonic like `repr(pre_tokenizer)` could be interesting but does not exist atm.

cakiki:
Thank you for the explanation @Narsil ! I'll go ahead and close the issue.

Issue #525, opened 2020-11-13
Ability to re-train a Tokenizer with relevant parameters


n1t0:
### Current state

When we want to train a Tokenizer, we need to give a `Trainer` initialized with a set of custom parameters:
```python
tokenizer = Tokenizer.from_file("byte-level-bpe.tokenizer.json")

# We need to provide the relevant parameters, to avoid using the general defaults
trainer = BpeTrainer(vocab_size=30000, special_tokens=[...], initial_alphabet=[...], ...)
tokenizer.train(files=[...], trainer=trainer)
```

### Goal

Add the ability to re-train a `Tokenizer` using the same custom parameters that were used for training the first time.
This would allow users to re-train some pre-trained tokenizers provided by the community with their own dataset. We'd be able to do this:
```python
tokenizer = Tokenizer.from_file("byte-level-bpe.tokenizer.json")
tokenizer.train(files=[ ... ])
```
and expect to get a `Tokenizer` very similar to the one we originally loaded (same special_tokens, vocab_size, ...), with a brand new vocabulary.

### How

One of the ways to achieve this is to make the `Trainer` save its training params on the `Model` during training, thus allowing `Model::get_trainer` to return a `Trainer` instantiated as expected. All of this should be added to the serialization process.

### Considerations

If some tokens were added with `add_tokens` or `add_special_tokens`, re-training is not currently supported because `AddedVocabulary` adds tokens on top of an existing vocabulary, expecting it to never change (cf #523)
Also depends on #527

Virviil:
Just want to check that i'm facing the same problem:

```
use tokenizers::{
    models::bpe::{BpeTrainer, BPE},
    pre_tokenizers::whitespace::Whitespace,
    AddedToken, DecoderWrapper, Model, NormalizerWrapper, PostProcessorWrapper,
    PreTokenizerWrapper, TokenizerImpl,
};

fn main() -> Result<(), tokenizers::Error> {
    let mut tokenizer: TokenizerImpl<
        BPE,
        NormalizerWrapper,
        PreTokenizerWrapper,
        PostProcessorWrapper,
        DecoderWrapper,
    > = TokenizerImpl::new(
        BPE::builder()
            .unk_token("[UNK]".to_string())
            .build()
            .unwrap(),
    );

    let mut trainer = BpeTrainer::builder()
        .special_tokens(vec![
            AddedToken::from("[UNK]", true),
            AddedToken::from("[CLS]", true),
[...]

PR #949, opened 2022-03-10
Adding pickling support for trainers

### LINKED ISSUE #941: Support for pickling trainers
Right now, it is impossible to pickle or deepcopy trainer objects. E.g.:

```python
# tokenizers.__version__ is '0.11.6'
import pickle
from copy import deepcopy
from tokenizers.trainers import WordPieceTrainer

trainer = WordPieceTrainer(
    vocab_size=30522, special_tokens=["[UNK]", "[CLS]", "[SEP]", "[PAD]", "[MASK]"]
)
pickle.dumps(trainer)  # raises TypeError
deepcopy(trainer)  # raises TypeError
```

I know that pickling has some disadvantages but sometimes it cannot be avoided, e.g. because an existing library uses it.

Since the trainer seems to be implemented directly in Rust, I don't know what the reason for the errors is.

Narsil:
Fixes #941

### DIFF
--- a/bindings/python/py_src/tokenizers/trainers/__init__.pyi
+++ b/bindings/python/py_src/tokenizers/trainers/__init__.pyi
@@ -7,19 +7,6 @@ class Trainer:
     Trainer will return an instance of this class when instantiated.
     """
 
-    def __init__(
-        self,
-        vocab_size=30000,
-        min_frequency=0,
-        show_progress=True,
-        special_tokens=[],
-        limit_alphabet=None,
-        initial_alphabet=[],
-        continuing_subword_prefix=None,
-        end_of_word_suffix=None,
-    ):
-        pass
-
 class BpeTrainer(Trainer):
     """
     Trainer capable of training a BPE model
--- a/bindings/python/src/trainers.rs
+++ b/bindings/python/src/trainers.rs
@@ -50,7 +50,7 @@ impl PyTrainer {
     fn __getstate__(&self, py: Python) -> PyResult<PyObject> {
         let data = serde_json::to_string(&self.trainer).map_err(|e| {
             exceptions::PyException::new_err(format!(
-                "Error while attempting to pickle PreBpeTrainer: {}",
+                "Error while attempting to pickle PyTrainer: {}",
                 e
             ))
         })?;
@@ -62,7 +62,7 @@ impl PyTrainer {
             Ok(s) => {
                 let unpickled = serde_json::from_slice(s.as_bytes()).map_err(|e| {
                     exceptions::PyException::new_err(format!(
-                        "Error while attempting to unpickle PreTokenizer: {}",
+                        "Error while attempting to unpickle PyTrainer: {}",
                         e
                     ))
                 })?;
--- a/bindings/python/src/trainers.rs
+++ b/bindings/python/src/trainers.rs
@@ -1,28 +1,32 @@
 use std::sync::{Arc, RwLock};
 
-use pyo3::exceptions;
-use pyo3::prelude::*;
-use pyo3::types::*;
-use tk::models::TrainerWrapper;
-use tk::Trainer;
-use tokenizers as tk;
-
 use crate::models::PyModel;
 use crate::tokenizer::PyAddedToken;
 use crate::utils::PyChar;
+use pyo3::exceptions;
+use pyo3::prelude::*;
+use pyo3::types::*;
+use serde::{Deserialize, Serialize};
+use tk::models::TrainerWrapper;
+use tk::Trainer;
+use tokenizers as tk;
 
 /// Base class for all trainers
 ///
 /// This class is not supposed to be instantiated directly. Instead, any implementation of a
 /// Trainer will return an instance of this class when instantiated.
 #[pyclass(name=Trainer, module = "tokenizers.trainers", name=Trainer)]
-#[derive(Clone)]
-#[text_signature = "(self, vocab_size=30000, min_frequency=0,show_progress=True, special_tokens=[],limit_alphabet=None, initial_alphabet = [], continuing_subword_prefix=None, end_of_word_suffix=None)"]
+#[derive(Clone, Deserialize, Serialize)]
 pub struct PyTrainer {
+    #[serde(flatten)]
     pub trainer: Arc<RwLock<TrainerWrapper>>,
 }
 
 impl PyTrainer {
+    #[cfg(test)]
+    pub(crate) fn new(trainer: Arc<RwLock<TrainerWrapper>>) -> Self {
+        PyTrainer { trainer }
+    }
     pub(crate) fn get_as_subtype(&self) -> PyResult<PyObject> {
         let base = self.clone();
         let gil = Python::acquire_gil();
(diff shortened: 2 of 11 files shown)


bindings/python/src/trainers.rs
```diff
@@ -41,6 +45,34 @@ impl PyTrainer {
         })
     }
 }
+#[pymethods]
+impl PyTrainer {
```
Narsil, line 49: The main implementation of pickling.


__getstate__ and __setstate__ are enough to pickle objects in python.

bindings/python/src/trainers.rs
```diff
@@ -41,6 +45,34 @@ impl PyTrainer {
...
+    fn __getstate__(&self, py: Python) -> PyResult<PyObject> {
+        let data = serde_json::to_string(&self.trainer).map_err(|e| {
```
McPatate, line 51: `serde_json` ? Shouldn't you be using [serde-pickle](https://docs.rs/serde-pickle/latest/serde_pickle/) ?

bindings/python/src/trainers.rs
```diff
@@ -1,28 +1,32 @@
...
-#[text_signature = "(self, vocab_size=30000, min_frequency=0,show_progress=True, special_tokens=[],limit_alphabet=None, initial_alphabet = [], continuing_subword_prefix=None, end_of_word_suffix=None)"]
+#[derive(Clone, Deserialize, Serialize)]
 pub struct PyTrainer {
+    #[serde(flatten)]
     pub trainer: Arc<RwLock<TrainerWrapper>>,
 }
 
 impl PyTrainer {
+    #[cfg(test)]
+    pub(crate) fn new(trainer: Arc<RwLock<TrainerWrapper>>) -> Self {
```
McPatate, line 27: Why do we need `Arc<>` here ?

bindings/python/src/trainers.rs
Narsil: Well we're in control of the bytes here, so we don't need to use pickle.

It's more consistent with the rest of the lib too (Moving would be another PR imo).

bindings/python/src/trainers.rs
Narsil: Because it's python a python object holding a ref to a rust object.

Since Python has the GIL we don't know if we're on the same thread or not, hence `Arc<RwLock<>>` To become `Send+Sync`. This is already what the `PyTrainer` owns as an object.

bindings/python/src/trainers.rs
McPatate: Would I be able to unpickle the output from python code without tokenizers imported ? I don't see any uses of having proper pickling, it's just to understand what's happening here :)

Narsil: Also I feel the need to mention that the json pickle thing is already being relied on : https://github.com/huggingface/tokenizers/issues/947 :).

(I think we should make the things more printable in python itself without relying on this, but hey.)

tokenizers/src/models/mod.rs
```diff
@@ -121,6 +121,7 @@ impl Model for ModelWrapper {
     }
 }
 
+#[derive(Serialize, Deserialize)]
```
mishig25, line 124: just to confirm, `TrainerWrapper` serde is using [external tagging](https://serde.rs/enum-representations.html#externally-tagged) (not internal) so that the serializations look like:
```json
{"BpeTrainer": { ... }
```

rather than
```json
{"type": "BpeTrainer", ... }
```
right?

bindings/python/py_src/tokenizers/trainers/__init__.pyi
```diff
@@ -7,19 +7,6 @@ class Trainer:
...
-        min_frequency=0,
-        show_progress=True,
-        special_tokens=[],
-        limit_alphabet=None,
-        initial_alphabet=[],
-        continuing_subword_prefix=None,
-        end_of_word_suffix=None,
-    ):
-        pass
-
```
mishig25, line 22: these lines were removed because of[ this change](https://github.com/huggingface/tokenizers/pull/949#discussion_r823826088) ?

bindings/python/src/trainers.rs
```diff
@@ -1,28 +1,32 @@
...
-use crate::tokenizer::PyAddedToken;
-use crate::utils::PyChar;
-
 /// Base class for all trainers
 ///
 /// This class is not supposed to be instantiated directly. Instead, any implementation of a
 /// Trainer will return an instance of this class when instantiated.
 #[pyclass(name=Trainer, module = "tokenizers.trainers", name=Trainer)]
-#[derive(Clone)]
-#[text_signature = "(self, vocab_size=30000, min_frequency=0,show_progress=True, special_tokens=[],limit_alphabet=None, initial_alphabet = [], continuing_subword_prefix=None, end_of_word_suffix=None)"]
```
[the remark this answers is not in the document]
mishig25: if `PyTrainer` is not created on its own, does it make sense to turn:
`class Trainer:` -> `class Trainer(ABC):` ? or it doesn't matter?

mishig25:
Is this PR somewhat related to https://github.com/huggingface/tokenizers/issues/525 ?

bindings/python/src/trainers.rs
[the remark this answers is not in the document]
Narsil: It doesn't matter very much I think.

Besides I am not sure how PyO3 would work with ABC.

tokenizers/src/models/mod.rs
Narsil: Very good remark.

I wasn't even sure, so I added a Python test to make sure we commit to the serialized format (or at least know when we break).

And yes it is.

bindings/python/tests/bindings/test_trainers.py
```diff
@@ -57,6 +58,21 @@ def test_can_modify(self):
         trainer.end_of_word_suffix = None
         assert trainer.continuing_subword_prefix == None
 
+    def test_can_pickle(self):
+        assert (
+            trainers.BpeTrainer(min_frequency=12).__getstate__()
+            == b"""{"BpeTrainer":{"min_frequency":12,"vocab_size":30000,"show_progress":true,"special_tokens":[],"limit_alphabet":null,"initial_alphabet":[],"continuing_subword_prefix":null,"end_of_word_suffix":null,"words":{}}}"""
```
Narsil, line 64: @mishig25 New test to make sure we're consistent in this serialization.

Narsil:
> Is this PR somewhat related to #525 ?

Not really unfortunately.

The main issue for that one, is information flow between a tokenizer and the trainer arguments. (The tokenizer doesn't know all training args, `vocab_size` might not be kept, things like `initial alphabet` are forgotten, and `special_tokens` might cause issues when they are added after training)

bindings/python/src/trainers.rs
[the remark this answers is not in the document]
mishig25: got it 🙂

mishig25 (approved):
lgtm! from my side

I can't judge much the `python binding` side (will learn/read more about `pyo3`)

PR #951, opened 2022-03-16
Use `thiserror` crate for Errors

mishig25:
# TLDR: use [thiserror crate](https://github.com/dtolnay/thiserror) for Errors

Pros: derive macro for [std::error::Error](https://doc.rust-lang.org/std/error/trait.Error.html) & [std::fmt::Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) traits while still being readable

Cons: a new dependency

@Narsil , I'll let you decide whether this PR is necessary. If not, I'll just close it 😊

### DIFF
--- a/tokenizers/src/models/bpe/mod.rs
+++ b/tokenizers/src/models/bpe/mod.rs
@@ -13,10 +13,18 @@ type Pair = (u32, u32);
 pub enum Error {
     /// An error encountered while reading files mainly.
     #[error("IoError: {0}")]
-    Io(#[from] #[source] std::io::Error),
+    Io(
+        #[from]
+        #[source]
+        std::io::Error,
+    ),
     /// An error forwarded from Serde, while parsing JSON
     #[error("JsonError: {0}")]
-    JsonError(#[from] #[source] serde_json::Error),
+    JsonError(
+        #[from]
+        #[source]
+        serde_json::Error,
+    ),
     /// When the vocab.json file is in the wrong format
     #[error("Bad vocabulary json file")]
     BadVocabulary,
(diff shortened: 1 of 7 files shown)


tokenizers/src/models/bpe/mod.rs
```diff
@@ -9,65 +9,40 @@ mod word;
...
 /// Errors that can be encountered while using or constructing a `BPE` model.
-#[derive(Debug)]
+#[derive(thiserror::Error, Debug)]
 pub enum Error {
     /// An error encountered while reading files mainly.
-    Io(std::io::Error),
+    #[error("IoError: {0}")]
+    Io(
+        #[from]
+        #[source]
```
Narsil, line 18: > The #[from] attribute always implies that the same field is #[source], so you don't ever need to specify both attributes.

https://github.com/dtolnay/thiserror

Do we really need `#[source]` here ?

Narsil (commented):
This looks much cleaner to me.

I love `thiserror`.

I will approve if `#[source]` is necessary in some way.

tokenizers/src/models/bpe/mod.rs
Narsil: I think the macro derives that automatically though, doesn't it ?

mishig25: according to this rule, it will not (unless I'm misreading it)
> The Error trait's source() method is implemented to return whichever field has a #[source] attribute or is named source, if any. This is for identifying the underlying lower level error that caused your error.

mishig25: https://github.com/huggingface/tokenizers/pull/951/commits/3856dbce6b74c4be9472b90cdfc5b313cbf4a951

PR #952, opened 2022-03-16
Fixing the vocab size of the trained Unigram model

### LINKED ISSUE #950: Different behaviors in adjusting vocab sizes when training WordPiece and Unigram
```python
    tokenizer = Tokenizer(Unigram())

    ...

    trainer = UnigramTrainer(
        vocab_size=30000,
        special_tokens=["[CLS]", "[SEP]", "<unk>", "<pad>", "[MASK]"], 
        unk_token="<unk>"
    )
    tokenizer.train_from_iterator(texts, trainer=trainer)

    ...

    tokenizer.save('foo/bar.json')
```

the resulting file contains 30,005 tokens 

```python
    tokenizer = BertWordPieceTokenizer()

    ...

    tokenizer.train_from_iterator(texts, vocab_size=30000)

    ...

    tokenizer.save('foo/bar.json')
```

the resulting file contains 30,000 tokens

HelloRusk:
fix #950

### DIFF
--- a/tokenizers/src/models/unigram/trainer.rs
+++ b/tokenizers/src/models/unigram/trainer.rs
@@ -150,9 +150,11 @@ impl UnigramTrainer {
             inserted.insert(token.to_string());
             pieces.push((token.to_string(), if score.is_nan() { 0.0 } else { *score }));
 
-            let vocab_size_without_special_tokens = 
-                if need_add_unk { self.vocab_size as usize - self.special_tokens.len() - 1 } 
-                else { self.vocab_size as usize - self.special_tokens.len() };
+            let vocab_size_without_special_tokens = if need_add_unk {
+                self.vocab_size as usize - self.special_tokens.len() - 1
+            } else {
+                self.vocab_size as usize - self.special_tokens.len()
+            };
             if pieces.len() == vocab_size_without_special_tokens {
                 break;
             }
--- a/tokenizers/src/models/unigram/trainer.rs
+++ b/tokenizers/src/models/unigram/trainer.rs
@@ -126,19 +126,7 @@ impl UnigramTrainer {
                 min_score_penalty += min_score_penalty_delta;
             }
         }
-        for (token, score) in model.iter() {
-            if inserted.contains::<str>(token) {
-                continue;
-            }
-            inserted.insert(token.to_string());
-            pieces.push((token.to_string(), if score.is_nan() { 0.0 } else { *score }));
-            if pieces.len() == self.vocab_size as usize - self.special_tokens.len() {
-                break;
-            }
-        }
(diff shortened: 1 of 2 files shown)


Narsil:
(So we know when we're breaking later)

HelloRusk:
@Narsil 

I have a question about where we should run `pytest` command. On `bindings/python/` or `bindings/python/tests`?

I ask this because some of the test files have its path like `tests/data/xxx.txt` while others have `data/xxx.txt`.
  
https://github.com/huggingface/tokenizers/blob/4b6055d4fb6ac8223c57ca087ad37693f13ea196/bindings/python/tests/bindings/test_trainers.py#L161

https://github.com/huggingface/tokenizers/blob/4b6055d4fb6ac8223c57ca087ad37693f13ea196/bindings/python/tests/bindings/test_trainers.py#L286

If we want to test on `bindings/python/`, all the paths are to be `tests/data/xxx.txt`.

Narsil:
You can run `make test` in the bindings to make it run properly. It will download the necessary files too.

And `data/` is for the static data needed for tests (and ignored in git). `tests/data` is where we save temporary files being created by the tests themselves (that can safely be removed afterwards)

HelloRusk:
Oh, I didn't notice the Makefile, thanks!

Narsil:
Hey, can you change this test instead : https://github.com/huggingface/tokenizers/blob/master/bindings/python/tests/bindings/test_trainers.py#L191  It's specifically used for testing training with special tokens, it just lacks vocab_size checking.

If we can keep the simple test about the simple case it's better.
If we hadn't already a test targeting training with special tokens we would have created one.

HelloRusk:
I discovered that if `unk_token` is not included in `special_tokens`, we have to reduce the vocab size by 1 token, so the new codes take care of both cases

tokenizers/src/models/unigram/trainer.rs
```diff
@@ -154,6 +142,26 @@ impl UnigramTrainer {
...
+                continue;
+            }
+            inserted.insert(token.to_string());
+            pieces.push((token.to_string(), if score.is_nan() { 0.0 } else { *score }));
+
+            let vocab_size_without_special_tokens = if need_add_unk {
+                self.vocab_size as usize - self.special_tokens.len() - 1
+            } else {
+                self.vocab_size as usize - self.special_tokens.len()
+            };
```
Narsil, line 157: Sorry I should have seen that earlier, do you mind extracting this from the loop (it can be defined on top.
This will likely be optimized away, but it still help readability too.

I can take care of it if you want.

Narsil (commented):
Last Nit, otherwise looks very good !

tokenizers/src/models/unigram/trainer.rs
HelloRusk: I should have found it out...

PR #954, opened 2022-03-18
Fixing issue with ConvBert not being able to save because of of holes in the vocab.

Narsil:
Original issue:

```python

tokenizer = AutoTokenizer.from_pretrained("YituTech/conv-bert-base")
tokenizer.save_pretrained("test-tok")
# or alternatively the line that causes the issue:
# tokenizer.backend_tokenizer.save("tokenizer.json")
```

Core of the issue is that the original `vocab.txt` contains duplicates (https://huggingface.co/YituTech/conv-bert-base/raw/main/vocab.txt look towards the end), meaning this code https://github.com/huggingface/transformers/blob/master/src/transformers/models/bert/tokenization_bert.py#L97 would created holes in the vocab (there's not check that `token` doesn't already exist).

The crust of the proposed fix, is to mimick slow tokenizers behavior, which is yelling that something bad is ocurring but continue saving the vocab without crashing.

### DIFF
--- a/tokenizers/src/models/mod.rs
+++ b/tokenizers/src/models/mod.rs
@@ -33,7 +33,17 @@ impl<'a> Serialize for OrderedVocabIter<'a> {
     where
         S: Serializer,
     {
-        let iter = (0u32..(self.vocab_r.len() as u32)).map(|i| (&self.vocab_r[&i], i));
+        // There could be holes so max + 1 is more correct than vocab_r.len()
+        let max = self.vocab_r.iter().map(|(key, _)| key).max().unwrap_or(&0) + 1;
+        let iter = (0..max).filter_map(|i| {
+            if let Some(token) = self.vocab_r.get(&i){
+                Some((token, i))
+            }else{
+                warn!("The OrderedVocab you are attempting to save contains a hole for index {}, your vocabulary could be corrupted !", i);
+                println!("The OrderedVocab you are attempting to save contains a hole for index {}, your vocabulary could be corrupted !", i);
+                None
+            }
+        });
         serializer.collect_map(iter)
     }
 }
@@ -193,5 +203,16 @@ mod tests {
 
         let result = trainer.train(&mut model);
         assert!(result.is_err());
+    }
+
+    #[test]
+    fn incomplete_ordered_vocab() {
+        let vocab_r: HashMap<u32, String> =
+            HashMap::from([(0, "Hi".to_string()), (2, "There".to_string())]);
+
+        let ordered = OrderedVocabIter::new(&vocab_r);
+
+        let serialized = serde_json::to_string(&ordered).unwrap();
+        assert_eq!(serialized, "{\"Hi\":0,\"There\":2}");
     }
 }
--- a/bindings/node/native/Cargo.lock
+++ b/bindings/node/native/Cargo.lock
@@ -1743,6 +1743,7 @@ dependencies = [
  "serde",
  "serde_json",
  "spm_precompiled",
+ "thiserror",
  "unicode-normalization-alignments",
  "unicode-segmentation",
  "unicode_categories",
--- a/bindings/python/Cargo.lock
+++ b/bindings/python/Cargo.lock
@@ -1719,18 +1719,18 @@ dependencies = [
 
 [[package]]
 name = "thiserror"
-version = "1.0.29"
+version = "1.0.30"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "602eca064b2d83369e2b2f34b09c70b605402801927c65c11071ac911d299b88"
+checksum = "854babe52e4df1653706b98fcfc05843010039b406875930a70e4d9644e5c417"
 dependencies = [
  "thiserror-impl",
 ]
 
 [[package]]
 name = "thiserror-impl"
-version = "1.0.29"
+version = "1.0.30"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bad553cc2c78e8de258400763a647e80e6d1b31ee237275d756f6836d204494c"
+checksum = "aa32fd3f627f367fe16f893e2597ae3c05020f8bba2666a4e6ea73d377e5714b"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -1788,6 +1788,7 @@ dependencies = [
  "serde",
  "serde_json",
  "spm_precompiled",
+ "thiserror",
  "unicode-normalization-alignments",
  "unicode-segmentation",
  "unicode_categories",
--- a/tokenizers/src/models/wordlevel/serialization.rs
+++ b/tokenizers/src/models/wordlevel/serialization.rs
@@ -83,7 +83,7 @@ impl<'de> Visitor<'de> for WordLevelVisitor {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use crate::models::wordlevel::{Vocab, WordLevel, WordLevelBuilder};
 
     #[test]
     fn serde() {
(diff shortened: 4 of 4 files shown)


tokenizers/src/models/mod.rs
```diff
@@ -33,7 +33,17 @@ impl<'a> Serialize for OrderedVocabIter<'a> {
...
     {
-        let iter = (0u32..(self.vocab_r.len() as u32)).map(|i| (&self.vocab_r[&i], i));
+        // There could be holes so max + 1 is more correct than vocab_r.len()
+        let max = self.vocab_r.iter().map(|(key, _)| key).max().unwrap_or(&0) + 1;
+        let iter = (0..max).filter_map(|i| {
+            if let Some(token) = self.vocab_r.get(&i){
+                Some((token, i))
+            }else{
+                warn!("The OrderedVocab you are attempting to save contains a hole for index {}, your vocabulary could be corrupted !", i);
+                println!("The OrderedVocab you are attempting to save contains a hole for index {}, your vocabulary could be corrupted !", i);
```
mishig25, line 43: why are `warn!` & `println!` both necessary? Because of bindings?

mishig25:
Does it also mean [load_vocab](https://github.com/huggingface/transformers/blob/master/src/transformers/models/bert/tokenization_bert.py#L97-L105) function needs to be fixed ?

tokenizers/src/models/mod.rs
Narsil: `warn!` won't print anything if the logger is not set (the default in rust).

Since the extreme level of concern this behavior should cause I think I want to really yell even if the `warn!` `debug!` thing is not setup.

Highly debattable.

Narsil:
> Does it also mean [load_vocab](https://github.com/huggingface/transformers/blob/master/src/transformers/models/bert/tokenization_bert.py#L97-L105) function needs to be fixed ?

I don't think it should, `bert` is so much legacy that changing anything here is just asking for trouble.

The old code does yell (like this PR will ) that something very fishy is at play though when attempting to save.

McPatate (approved):
Kudos for the small PR with testing !

tokenizers/src/models/mod.rs
```diff
@@ -33,7 +33,17 @@ impl<'a> Serialize for OrderedVocabIter<'a> {
     where
         S: Serializer,
     {
-        let iter = (0u32..(self.vocab_r.len() as u32)).map(|i| (&self.vocab_r[&i], i));
+        // There could be holes so max + 1 is more correct than vocab_r.len()
+        let max = self.vocab_r.iter().map(|(key, _)| key).max().unwrap_or(&0) + 1;
```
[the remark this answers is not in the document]
Narsil: Yes, but the following line too, so we're not worsening anything here (adding a constant)

Narsil:
> If the default value is just error, maybe add a note in the documentation saying the user has to activate log level to warn if he wants it. IMO, default log level should be info :)

No the default is showing nothing, you need to set `TOKENIZERS_LOG` to see anything in Python ( in rust, you need to use your favorite logging tool).

> I would let this to be the control of the user, if he deactivates warnings, he knows what he's in for.

Yes, my issue is that it is not on by default, meaning this warning will get lost 99% of the time, that's why the println.
I could remove the `warn!` altogether, but having holes in a vocabulary should warrant a LOT of attention. panicking was actually alright IMO, but if you look at the original issue, it's not the behavior of `transformers` so we sort have to align.
Having some sort of flag could help but would be tedious to maintaining. Yelling hard is OK imo, we can always revisit later when people complain (but given the seriousness of having holes I think fixing the actual tokenizer is the only way to "solve" this atm)

McPatate:
Ah, I didn't know logging was opt-in 😕 

In that case, yes, I think it's ok to leave the `println!` and the `warn!`, even if redundant, it's good to have the associated criticality. Better bark twice that nonce !

mbrunecky:
I am sorry entering this discussion with my limited understanding, but this issue prevents me from using bert-conv-base saved model. Under Spacy (3.4), switching from the default roberta to bert-conv-base is deceptively easy.
Except that now, when saving the model (checkpoint) during training, I get over thousand of those 'contains a hole' messages. 
And of course, when loaded, such model is completely useless - despite NOT reporting any error/warning. The only 'hint' that something is wrong is instead of p/r/f scores near 97%, the same evaluation data gives p/r/f scores near ZERO.

Perhaps a workaround is using some 'other' tokenizer (if I can figure out how to do it under Spacy), But I would thing that the 'fix' in case of 'holes' is either switching to some 'slow' save (that can be loaded) OR a hard failure - rather than pretending that the model 'may' work.

mbrunecky:
With help of [polm](https://github.com/polm) I traced the problem to the 'vocab.txt' checked-in https://huggingface.co/YituTech/conv-bert-base  containing (many) duplicates. There is 188 'tokens' replicated, many of them of them up to 34 times, for a total of 1008 'extra' (replicated) entries. Replicated entries are not loaded by the huggingface code fixed here. ConvBERT replicated entries are (probably entirely) on Chinese tokens.

My 'solution' (workaround) is patching the 'vocab.txt' making the 'extra' (repeated) token entries unique by adding a suffix such as '00', '01'...  resulting in '銆00'. etc.  Such tokens are highly unlikely in any text, especially in my case (English). It takes care of the 'contains hole' problem, and patched convert-bert-base now works with Spacy (3.4) as advertised. Early results suggest that it provides a bit better result or my NER domain, and the evaluation speed appears faster (not sure about training - need more testing).

I uploaded my patch as: 
[vocab_patch.txt](https://github.com/huggingface/tokenizers/files/9699743/vocab_patch.txt)
Simply replace vocab.txt content (in cloned conv-bert-base) with my patch.

Narsil:
Hi @mbrunecky ,

Do you have a minimal reproducible example of your issue ?
I have limited knowledge of `spacy` so I am not sure about what you are referring to.

>  But I would thing that the 'fix' in case of 'holes' is either switching to some 'slow' save (that can be loaded) OR a hard failure - rather than pretending that the model 'may' work.

Hard error used to be the behavior, the problem is that real tokenizer (like convBert) did exist on hf.co meaning there was no way to save this tokenizer anymore. Because we try to be backward compatible as much as possible, we decided to go for warnings instead of hard error (so that existing tokenizers with holes/duplicates) could still be manipulated.

If you can provide a reproducible example we can look as what happens, notably the p/r/f scores you are hinting which seem to be the underlying issue.

mbrunecky:
Hi [Narsil](https://github.com/Narsil).
Thank you for looking at the issue. Since I found a workaround, I did not investigate the cause of the bad p/r/f scores - I simply _assumed_ that missing vocab entries result in 'shifting' token indexes in the ordered vocab and hence miss-representing the vocab (I did not see any code compensating for that). 
With my limited time, I am mainly trying to decide IF the ConvBERT is worth using for some of my data - if the gains out weight the risk of using the code which may have little or no support.
And trying to go back to find out what created those gaps in the first place is probably impossible - the underlying code has changed. My testing shows that training the model now does not create any 'new' gaps, so the original problem is either gone or limited to Chinese characters.

I posted my 'workaround' (fixed vocab.txt) because it makes the posted conv-bert-base (and resulting models) usable, and it does not require any code changes. Regarding the difference between a warning and hard error, it is mostly a preference: in either case you can not use the model as-is because even if the resulting model worked, getting 1008 warnings on each model training checkpoint/save is prohibitive (how do you find any 'real' errors?).

Spacy does not have much to do with the issue. Initially I thought the problem may be caused/fixed by a tokenizer - but it is not. It is simply a broken vocab,txt. I would run into the same problem using HuggingFace environment. I just use Spacy for multiple reasons - such as transparently using other than transformer models.

PR #650, opened 2021-03-10
Upgrade pyo3 to 0.15

messense:
This would require dropping Python 3.5 support.

PR #956, opened 2022-03-21
Upgrade pyo3 to 0.16

### LINKED ISSUE #934: Upgrade pyo3 version
To 0.16 currently (we're on `0.12`)

h-vetinari:
Rebased #650 by @messense

Closes #934

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -63,6 +63,13 @@ jobs:
           toolchain: stable
           components: rustfmt, clippy
 
+      - name: Install Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+          architecture: "x64"
+
+
       - name: Cache Cargo Registry
         uses: actions/cache@v1
         with:
@@ -87,12 +94,6 @@ jobs:
         with:
           command: clippy
           args: --manifest-path ./bindings/python/Cargo.toml --all-targets --all-features -- -D warnings
-
-      - name: Install Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.9
-          architecture: "x64"
 
       - name: Install
         working-directory: ./bindings/python
--- a/bindings/python/Cargo.toml
+++ b/bindings/python/Cargo.toml
@@ -14,8 +14,8 @@ serde = { version = "1.0", features = [ "rc", "derive" ]}
 serde_json = "1.0"
 libc = "0.2"
 env_logger = "0.7.1"
-pyo3 = "0.12"
-numpy = "0.12"
+pyo3 = "0.15.0"
+numpy = "0.15.0"
 ndarray = "0.13"
 onig = { version = "6.0", default-features = false }
 itertools = "0.9"
@@ -29,5 +29,3 @@ tempfile = "3.1"
 
 [features]
 default = ["pyo3/extension-module"]
-
-
--- a/bindings/python/Cargo.toml
+++ b/bindings/python/Cargo.toml
@@ -26,6 +26,7 @@ path = "../../tokenizers"
 
 [dev-dependencies]
 tempfile = "3.1"
+pyo3 = { version = "0.16.2", features = ["auto-initialize"] }
 
 [features]
 default = ["pyo3/extension-module"]
--- a/bindings/python/Cargo.toml
+++ b/bindings/python/Cargo.toml
@@ -14,8 +14,8 @@ serde = { version = "1.0", features = [ "rc", "derive" ]}
 serde_json = "1.0"
 libc = "0.2"
 env_logger = "0.7.1"
-pyo3 = "0.15.0"
-numpy = "0.15.0"
+pyo3 = "0.16.2"
+numpy = "0.16.2"
 ndarray = "0.13"
 onig = { version = "6.0", default-features = false }
 itertools = "0.9"
--- a/bindings/python/src/decoders.rs
+++ b/bindings/python/src/decoders.rs
@@ -21,7 +21,7 @@ use super::error::ToPyResult;
 ///
 /// This class is not supposed to be instantiated directly. Instead, any implementation of
 /// a Decoder will return an instance of this class when instantiated.
-#[pyclass(dict, module = "tokenizers.decoders", name = "Decoder")]
+#[pyclass(dict, module = "tokenizers.decoders", name = "Decoder", subclass)]
 #[derive(Clone, Deserialize, Serialize)]
 pub struct PyDecoder {
     #[serde(flatten)]
--- a/bindings/python/src/decoders.rs
+++ b/bindings/python/src/decoders.rs
@@ -420,7 +420,7 @@ mod test {
         let py_meta = py_dec.get_as_subtype().unwrap();
         let gil = Python::acquire_gil();
         assert_eq!(
-            "tokenizers.decoders.Metaspace",
+            "Metaspace",
             py_meta.as_ref(gil.python()).get_type().name().unwrap()
         );
     }
--- a/bindings/python/src/decoders.rs
+++ b/bindings/python/src/decoders.rs
@@ -21,7 +21,7 @@ use super::error::ToPyResult;
 ///
 /// This class is not supposed to be instantiated directly. Instead, any implementation of
 /// a Decoder will return an instance of this class when instantiated.
-#[pyclass(dict, module = "tokenizers.decoders", name=Decoder)]
+#[pyclass(dict, module = "tokenizers.decoders", name = "Decoder")]
 #[derive(Clone, Deserialize, Serialize)]
 pub struct PyDecoder {
     #[serde(flatten)]
@@ -97,7 +97,7 @@ impl PyDecoder {
     ///
     /// Returns:
     ///     :obj:`str`: The decoded string
-    #[text_signature = "(self, tokens)"]
+    #[pyo3(text_signature = "(self, tokens)")]
     fn decode(&self, tokens: Vec<String>) -> PyResult<String> {
         ToPyResult(self.decoder.decode(tokens)).into()
     }
@@ -141,8 +141,8 @@ macro_rules! setter {
 ///
 /// This decoder is to be used in tandem with the :class:`~tokenizers.pre_tokenizers.ByteLevel`
 /// :class:`~tokenizers.pre_tokenizers.PreTokenizer`.
-#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name=ByteLevel)]
-#[text_signature = "(self)"]
+#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name = "ByteLevel")]
+#[pyo3(text_signature = "(self)")]
 pub struct PyByteLevelDec {}
 #[pymethods]
 impl PyByteLevelDec {
@@ -161,8 +161,8 @@ impl PyByteLevelDec {
 ///     cleanup (:obj:`bool`, `optional`, defaults to :obj:`True`):
 ///         Whether to cleanup some tokenization artifacts. Mainly spaces before punctuation,
 ///         and some abbreviated english forms.
-#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name=WordPiece)]
-#[text_signature = "(self, prefix=\"##\", cleanup=True)"]
+#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name = "WordPiece")]
+#[pyo3(text_signature = "(self, prefix=\"##\", cleanup=True)")]
 pub struct PyWordPieceDec {}
 #[pymethods]
 impl PyWordPieceDec {
@@ -203,8 +203,8 @@ impl PyWordPieceDec {
 ///     add_prefix_space (:obj:`bool`, `optional`, defaults to :obj:`True`):
 ///         Whether to add a space to the first word if there isn't already one. This
 ///         lets us treat `hello` exactly like `say hello`.
-#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name=Metaspace)]
-#[text_signature = "(self, replacement = \"▁\", add_prefix_space = True)"]
+#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name = "Metaspace")]
+#[pyo3(text_signature = "(self, replacement = \"▁\", add_prefix_space = True)")]
 pub struct PyMetaspaceDec {}
 #[pymethods]
 impl PyMetaspaceDec {
@@ -244,8 +244,8 @@ impl PyMetaspaceDec {
 ///     suffix (:obj:`str`, `optional`, defaults to :obj:`</w>`):
 ///         The suffix that was used to caracterize an end-of-word. This suffix will
 ///         be replaced by whitespaces during the decoding
-#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name=BPEDecoder)]
-#[text_signature = "(self, suffix=\"</w>\")"]
+#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name = "BPEDecoder")]
+#[pyo3(text_signature = "(self, suffix=\"</w>\")")]
 pub struct PyBPEDecoder {}
 #[pymethods]
 impl PyBPEDecoder {
@@ -276,8 +276,8 @@ impl PyBPEDecoder {
 ///     cleanup (:obj:`bool`, `optional`, defaults to :obj:`True`):
 ///         Whether to cleanup some tokenization artifacts.
 ///         Mainly spaces before punctuation, and some abbreviated english forms.
-#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name=CTC)]
-#[text_signature = "(self, pad_token=\"<pad>\", word_delimiter_token=\"|\", cleanup=True)"]
+#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name = "CTC")]
+#[pyo3(text_signature = "(self, pad_token=\"<pad>\", word_delimiter_token=\"|\", cleanup=True)")]
 pub struct PyCTCDecoder {}
 #[pymethods]
 impl PyCTCDecoder {
@@ -421,7 +421,7 @@ mod test {
         let gil = Python::acquire_gil();
         assert_eq!(
             "tokenizers.decoders.Metaspace",
-            py_meta.as_ref(gil.python()).get_type().name()
+            py_meta.as_ref(gil.python()).get_type().name().unwrap()
         );
     }
 
--- a/bindings/python/src/encoding.rs
+++ b/bindings/python/src/encoding.rs
@@ -9,7 +9,7 @@ use tokenizers as tk;
 use crate::error::{deprecation_warning, PyError};
 
 /// The :class:`~tokenizers.Encoding` represents the output of a :class:`~tokenizers.Tokenizer`.
-#[pyclass(dict, module = "tokenizers", name=Encoding)]
+#[pyclass(dict, module = "tokenizers", name = "Encoding")]
 #[repr(transparent)]
 pub struct PyEncoding {
     pub encoding: tk::tokenizer::Encoding,
@@ -86,7 +86,7 @@ impl PyEncoding {
     ///     :class:`~tokenizers.Encoding`: The resulting Encoding
     #[staticmethod]
     #[args(growing_offsets = true)]
-    #[text_signature = "(encodings, growing_offsets=True)"]
+    #[pyo3(text_signature = "(encodings, growing_offsets=True)")]
     fn merge(encodings: Vec<PyRef<PyEncoding>>, growing_offsets: bool) -> PyEncoding {
         tk::tokenizer::Encoding::merge(
             encodings.into_iter().map(|e| e.encoding.clone()),
@@ -108,7 +108,7 @@ impl PyEncoding {
     ///
     /// Set the given sequence index for the whole range of tokens contained in this
     /// :class:`~tokenizers.Encoding`.
-    #[text_signature = "(self, sequence_id)"]
+    #[pyo3(text_signature = "(self, sequence_id)")]
     fn set_sequence_id(&mut self, sequence_id: usize) {
         self.encoding.set_sequence_id(sequence_id);
     }
@@ -270,7 +270,7 @@ impl PyEncoding {
     /// Returns:
     ///     :obj:`Tuple[int, int]`: The range of tokens: :obj:`(first, last + 1)`
     #[args(sequence_index = 0)]
-    #[text_signature = "(self, word_index, sequence_index=0)"]
+    #[pyo3(text_signature = "(self, word_index, sequence_index=0)")]
     fn word_to_tokens(&self, word_index: u32, sequence_index: usize) -> Option<(usize, usize)> {
         self.encoding.word_to_tokens(word_index, sequence_index)
     }
@@ -286,7 +286,7 @@ impl PyEncoding {
     /// Returns:
     ///     :obj:`Tuple[int, int]`: The range of characters (span) :obj:`(first, last + 1)`
     #[args(sequence_index = 0)]
-    #[text_signature = "(self, word_index, sequence_index=0)"]
+    #[pyo3(text_signature = "(self, word_index, sequence_index=0)")]
     fn word_to_chars(&self, word_index: u32, sequence_index: usize) -> Option<Offsets> {
         self.encoding.word_to_chars(word_index, sequence_index)
     }
@@ -302,7 +302,7 @@ impl PyEncoding {
     ///
     /// Returns:
     ///     :obj:`int`: The sequence id of the given token
-    #[text_signature = "(self, token_index)"]
+    #[pyo3(text_signature = "(self, token_index)")]
     fn token_to_sequence(&self, token_index: usize) -> Option<usize> {
         self.encoding.token_to_sequence(token_index)
     }
@@ -319,7 +319,7 @@ impl PyEncoding {
     ///
     /// Returns:
     ///     :obj:`Tuple[int, int]`: The token offsets :obj:`(first, last + 1)`
-    #[text_signature = "(self, token_index)"]
+    #[pyo3(text_signature = "(self, token_index)")]
     fn token_to_chars(&self, token_index: usize) -> Option<Offsets> {
         let (_, offsets) = self.encoding.token_to_chars(token_index)?;
         Some(offsets)
@@ -337,7 +337,7 @@ impl PyEncoding {
     ///
     /// Returns:
     ///     :obj:`int`: The index of the word in the relevant input sequence.
-    #[text_signature = "(self, token_index)"]
+    #[pyo3(text_signature = "(self, token_index)")]
     fn token_to_word(&self, token_index: usize) -> Option<u32> {
         let (_, word_idx) = self.encoding.token_to_word(token_index)?;
         Some(word_idx)
@@ -354,7 +354,7 @@ impl PyEncoding {
     /// Returns:
     ///     :obj:`int`: The index of the token that contains this char in the encoded sequence
     #[args(sequence_index = 0)]
-    #[text_signature = "(self, char_pos, sequence_index=0)"]
+    #[pyo3(text_signature = "(self, char_pos, sequence_index=0)")]
     fn char_to_token(&self, char_pos: usize, sequence_index: usize) -> Option<usize> {
         self.encoding.char_to_token(char_pos, sequence_index)
     }
@@ -370,7 +370,7 @@ impl PyEncoding {
     /// Returns:
     ///     :obj:`int`: The index of the word that contains this char in the input sequence
     #[args(sequence_index = 0)]
-    #[text_signature = "(self, char_pos, sequence_index=0)"]
+    #[pyo3(text_signature = "(self, char_pos, sequence_index=0)")]
     fn char_to_word(&self, char_pos: usize, sequence_index: usize) -> Option<u32> {
         self.encoding.char_to_word(char_pos, sequence_index)
     }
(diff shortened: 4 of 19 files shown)


h-vetinari:
Looking at the abundance of CI errors, I'm not sure I'm going to be the best person to shepherd this to completion. I don't know rust that well, and pyo3/tokenizers even less. With some guidance I might get there, but I'm not going to be autonomous.

messense:
@h-vetinari Here is the pyo3 migration guide: https://pyo3.rs/v0.15.1/migration.html

Narsil:
Hi here.

First, I think we should move onto `0.16` directly since it's the latest version of `pyo3` (as long as we're making an update here, we might as well get the latest one). Unless there's some unintended breaking changes that could prevent this from happening.

Narsil:
For the linting, you can normally use `make style` within `bindings/python` to fix the format.
We also use `clippy` (`cargo clippy`) in the formatting

messense:
FYI, move onto pyo3 0.16 requires dropping Python 3.6 support.

Narsil:
> FYI, move onto pyo3 0.16 requires dropping Python 3.6 support.

It will refuse to compile ? If that's the case it it's not great.
`tokenizers` stopped building for `3.6` because there's not GH runner anymore, but if people are able to build still that would be better indeed. I don't see any killer feature for `0.16` in the changelog.

messense:
> It will refuse to compile ? If that's the case it it's not great.

I think so. 

Python 3.6 reached EOL in 23 Dec 2021 so users should upgrade if they care about security.

h-vetinari:
If someone wants to push into this PR, I'd be thrilled to receive support (and make people collaborators on my fork if necessary).

I mainly attempted this because pyo3>=0.15 is a hard [requirement](https://github.com/PyO3/pyo3/pull/1847#discussion_r727122252) to support python 3.10 in conda-forge, and several NLP packages are blocked on not having tokenizers for 3.10.

h-vetinari:
However, a minimal backport of #650 to `tags/python-v0.11.6` also didn't work.

It fails with something like:
```
import: 'tokenizers'
TypeError: type 'tokenizers.models.Model' is not an acceptable base type
thread '<unnamed>' panicked at 'An error occurred while initializing class BPE', /home/conda/feedstock_root/build_artifacts/tokenizers_1647850774618/_build_env/.cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-0.15.0/src/type_object.rs:102:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
pyo3_runtime.PanicException: An error occurred while initializing class BPE
thread '<unnamed>' panicked at 'Python API call failed'
```

h-vetinari:
PS. I don't care much about pyo3 0.15 or 0.16, though IMO python 3.6 support should really not be a determining factor in anything anymore. A lot of projects following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) (among them numpy, scipy, pandas, etc.) also dropped support for python 3.7 in their latest releases already.

Narsil:
> PS. I don't care much about pyo3 0.15 or 0.16, though IMO python 3.6 support should really not be a determining factor in anything anymore. For perspective - a lot of projects following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) (among them numpy, scipy, pandas, etc.) also dropped support for python 3.7 in their latest releases already.

Fair enough ! Let's try 0.16 then.

messense:
Hey @h-vetinari, I've sent you a PR to upgrade to 0.16: https://github.com/h-vetinari/tokenizers/pull/1

h-vetinari:
> Hey @h-vetinari, I've sent you a PR to upgrade to 0.16: [h-vetinari#1](https://github.com/h-vetinari/tokenizers/pull/1)

Thanks a lot! 🙃 

I sent you an invite to collaborate on my fork, then you can push into this PR directly

messense:
> TypeError: PreTokenizedEncodeInput must be Union[PreTokenizedInputSequence, Tuple[PreTokenizedInputSequence, PreTokenizedInputSequence]]

I'm not sure what changed in rust-numpy or pyo3 that makes this test case fail: https://github.com/huggingface/tokenizers/runs/5625562727?check_suite_focus=true

@adamreichold Any idea?

adamreichold:
> @adamreichold Any idea?

I am sorry but I have a hard time following the layers here. My first impression is that the test in question does not even reach the Rust code yet but fails already in the Python around it? That said, I think the best candidate for surfacing typing issues at runtime is that before 0.16, rust-numpy (incorrectly) did not check element type and dimension when downcasting to arrays, i.e. https://github.com/PyO3/rust-numpy/pull/265 (I did not find any mention of downcasts with `PyArray` though. Actually I did not find `PyArray` mentioned at all?)

messense:
> My first impression is that the test in question does not even reach the Rust code yet but fails already in the Python around it?

It's rejected here in Rust code https://github.com/huggingface/tokenizers/blob/1bb9884f459376508059b11b2ac6dc1585f44050/bindings/python/src/tokenizer.rs#L420-L423

messense:
And I suspect it has something to do with these code in `PyArrayUnicode` or `PyArrayStr`

https://github.com/huggingface/tokenizers/blob/1bb9884f459376508059b11b2ac6dc1585f44050/bindings/python/src/tokenizer.rs#L259-L340

adamreichold:
> And I suspect it has something to do with these code in `PyArrayUnicode` or `PyArrayStr`
> 
> https://github.com/huggingface/tokenizers/blob/1bb9884f459376508059b11b2ac6dc1585f44050/bindings/python/src/tokenizer.rs#L259-L340

This would indeed point to the downcast fixes and hence did probably only work by accident before. I think adding an `.unwrap();` instead of `?` in https://github.com/huggingface/tokenizers/blob/1bb9884f459376508059b11b2ac6dc1585f44050/bindings/python/src/tokenizer.rs#L262 might shed some light on this.

messense:
Fails with `PyDowncastError`

> pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: PyDowncastError { from: array(['My', 'name', 'is', 'John'], dtype='<U7'), to: "PyArray<T, D>" }

adamreichold:
> Fails with `PyDowncastError`
> 
> > pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: PyDowncastError { from: array(['My', 'name', 'is', 'John'], dtype='<U7'), to: "PyArray<T, D>" }

This should not have worked in the first place as `<U7` would an array of 7-byte-large elements of `Py_UCS4` stored in little-endian byte-order. So considering this a single byte `u8` elements should yield at least incorrect strides.
 
> I guess tokenizers also wants [PyO3/rust-numpy#141](https://github.com/PyO3/rust-numpy/issues/141)

This would be the best solution, but the existing code does not really use the API provided by `rust-numpy` anyway (which is why this worked in the past), so I think an immediate fix would be to just use the [`PyArray_Check`](https://docs.rs/numpy/latest/numpy/npyffi/array/fn.PyArray_Check.html) directly (which checks only if it is an array but does not consider element type and dimensionality):

So instead of

```rust
fn extract(ob: &PyAny) -> PyResult<Self> {
        let array = ob.downcast::<PyArray1<u8>>()?;
        let arr = array.as_array_ptr();
...
```

one could do

```rust
fn extract(ob: &PyAny) -> PyResult<Self> {
        if npyffi::PyArray_Check(ob.py(), ob.as_ptr()) == 0 {
            return Err(exceptions::PyTypeError::new_err(
                "Expected an np.array",
            ));
        }
        let arr = ob.as_ptr() as *mut npyffi::PyArrayObject;
...
```

and the rest of the code should continue to work as-is. (I have not actually tried to compile this so there are certainly errors in there.)

adamreichold:
I am sorry that I did not try this out myself, but from reading the code, I think the part

```rust
let shape =
    unsafe { slice::from_raw_parts((*arr).dimensions as *mut usize, (*arr).nd as usize) };
let n_elem = shape[0];
```

should probably check `nd` for correctness and also verify contiguousness as it then accesses the data as a slice (Even a one-dimensional array could have non-unit strides.), i.e.

```rust
if (*arr).nd != 1 { /* return dimensionality error */ }
let n_elem = *(*arr).dimenions;

if (*arr).flags & (npyffi::NPY_ARRAY_C_CONTIGUOUS | npyffi::NPY_ARRAY_F_CONTIGUOUS) == 0 { /* return non-contiguours error */ }
```

bindings/python/src/tokenizer.rs
```diff
@@ -258,8 +259,24 @@ impl<'s> From<TextInputSequence<'s>> for tk::InputSequence<'s> {
...
+            return Err(exceptions::PyTypeError::new_err(
+                "Expected a 1 dimensional np.array",
+            ));
+        }
+        if unsafe { (*arr).flags }
+            & (npyffi::NPY_ARRAY_C_CONTIGUOUS | npyffi::NPY_ARRAY_F_CONTIGUOUS)
+            == 0
+        {
+            return Err(exceptions::PyTypeError::new_err(
+                "Expected a continuous np.array",
```
adamreichold, line 276: `s/continuous/contiguous` (and in line 344)

bindings/python/Cargo.toml
```diff
@@ -13,12 +13,12 @@ rayon = "1.3"
 serde = { version = "1.0", features = [ "rc", "derive" ]}
 serde_json = "1.0"
 libc = "0.2"
-env_logger = "0.7.1"
-pyo3 = "0.12"
-numpy = "0.12"
-ndarray = "0.13"
+env_logger = "0.9.0"
+pyo3 = "0.16.2"
```
adamreichold, line 17: I think the remaining test failures could be resolved by adding

```toml
resolver = "2" # or edition = "2021"

[dev-dependencies]
pyo3 = { version = "0.16", features = ["auto-initialize"]
```

This used to be part of the default features but is not any more since 0.14.

(If Rust 1.51 [which introduced the `resolver = "2"`](https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html#details) option, then the feature can just be added to the normal `[dependencies]` entry.)

Narsil:
Hi @messense ,

I don't have time today to do a full review (I tried  to make the tests run during the day so you could see what was happening).

This is a becoming a very big PR, which I don't think is a good thing about a PR.

Do you mind adding comments yourself on the PRs of what is going on, and why changes are important ? It would help me tremendously review faster (otherwise I will just ask questions :))

The `unsafe` calls are basically big NO in `tokenizers`.

adamreichold:
> The unsafe calls are basically big NO in tokenizers.

If this refers to the calls related to `npyfii`, this is not materially "more unsafe" than it already was as the old (and incorrect) version of `downcast::<PyArray1<u8>>` was doing the exact same thing. The only difference is the manual access to the dimensions, but I would say that this is balanced out by fixing the missing check for contiguous arrays which is missing on `main` (or alternatively the accesses would need to consider the array's stride to be fully general).

Hopefully, we will be able to implement https://github.com/PyO3/rust-numpy/issues/141 eventually and the whole business can be done using safe code.

bindings/python/src/tokenizer.rs
```diff
@@ -310,10 +326,27 @@ impl From<PyArrayUnicode> for tk::InputSequence<'_> {
...
+            return Err(exceptions::PyTypeError::new_err(
+                "Expected a contiguous np.array",
+            ));
+        }
+        let n_elem = unsafe { *(*arr).dimensions } as usize;
+
         let (type_num, data) = unsafe { ((*(*arr).descr).type_num, (*arr).data) };
-        let n_elem = array.shape()[0];
 
         if type_num != 17 {
```
adamreichold, line 351: I just noticed that this case is not about a Unicode array, but actually an array containing `PyObject` which we do support since 0.16, so this whole method be able to become safe by using `downcast::<PyArray1<PyObject>>()` and then `array.readonly().as_array().iter()` which would even remove the requirement of a contiguous array.

bindings/python/src/tokenizer.rs
```diff
@@ -270,7 +286,6 @@ impl FromPyObject<'_> for PyArrayUnicode {
                 (*arr).data,
             )
         };
-        let n_elem = array.shape()[0];
 
         // type_num == 19 => Unicode
         if type_num != 19 {
```
adamreichold, line 291: I think this could be more readable by using the [`npyffi::types::NPY_TYPES::NPY_UNICODE`](https://docs.rs/numpy/latest/numpy/npyffi/types/enum.NPY_TYPES.html#variant.NPY_UNICODE) constant.

.github/workflows/python.yml
```diff
@@ -63,6 +63,13 @@ jobs:
           toolchain: stable
           components: rustfmt, clippy
 
+      - name: Install Python
```
messense, line 66: Moved this up because Ubuntu 18.04 comes with Python 3.6 and now we require Python 3.7.

bindings/python/Cargo.toml
```diff
@@ -13,21 +13,20 @@ rayon = "1.3"
...
+numpy = "0.16.2"
+ndarray = "0.15"
 onig = { version = "6.0", default-features = false }
-itertools = "0.9"
+itertools = "0.10"
```
messense, line 21: Since we're updating dependencies, upgrade `env_logger` , `ndarray` and `itertools` along the way.

bindings/python/Cargo.toml
```diff
@@ -13,21 +13,20 @@ rayon = "1.3"
...
 
 [dependencies.tokenizers]
 version = "*"
 path = "../../tokenizers"
 
 [dev-dependencies]
 tempfile = "3.1"
+pyo3 = { version = "0.16.2", features = ["auto-initialize"] }
```
messense, line 29: `auto-initialize` feature is required for `cargo test` to work.

bindings/python/src/decoders.rs
```diff
@@ -21,7 +21,7 @@ use super::error::ToPyResult;
 ///
 /// This class is not supposed to be instantiated directly. Instead, any implementation of
 /// a Decoder will return an instance of this class when instantiated.
-#[pyclass(dict, module = "tokenizers.decoders", name=Decoder)]
+#[pyclass(dict, module = "tokenizers.decoders", name = "Decoder", subclass)]
```
messense, line 24: 1. The `name` change is a breaking change from pyo3.
2. Added `subclass` because since pyo3 0.13 classes are not inheritable by default unless `subclass` attribute is set: https://pyo3.rs/v0.16.2/migration.html#runtime-changes-to-support-the-cpython-limited-api

bindings/python/src/encoding.rs
```diff
@@ -73,6 +54,18 @@ impl PyEncoding {
         }
     }
 
+    fn __repr__(&self) -> PyResult<String> {
```
messense, line 57: `#[pyproto]` is deprecated in pyo3 0.16: https://pyo3.rs/main/migration.html#pyproto-has-been-deprecated , `#[pymethods]` is the replacement.

bindings/python/src/lib.rs
```diff
@@ -126,7 +126,7 @@ fn normalizers(_py: Python, m: &PyModule) -> PyResult<()> {
 /// Tokenizers Module
 #[pymodule]
 fn tokenizers(_py: Python, m: &PyModule) -> PyResult<()> {
-    env_logger::init_from_env("TOKENIZERS_LOG");
+    let _ = env_logger::try_init_from_env("TOKENIZERS_LOG");
```
messense, line 129: Ignore error if logger is already initialized.

bindings/python/src/trainers.rs
```diff
@@ -864,8 +868,8 @@ mod tests {
         let py_bpe = py_trainer.get_as_subtype().unwrap();
         let gil = Python::acquire_gil();
         assert_eq!(
-            "tokenizers.trainers.BpeTrainer",
-            py_bpe.as_ref(gil.python()).get_type().name()
+            "BpeTrainer",
```
messense, line 871: This is odd, I'm not sure why.

cc @adamreichold @davidhewitt

messense (commented):
This PR is better reviewed commit by commit.

bindings/python/Cargo.toml
adamreichold: I think adding it here without also adding `resolver = "2"` in a leaf crate (this will build a cdylib at the end of the day) means that the features will be unified, so one could just as well enable the features on the dependency itself. (But I think adding `resolver = "2"` would be nicer in any case, but yeah another change in an already big PR.)

messense: CI was running with `cargo test --no-default-features` so the `extension-module`  feature is disabled thus only `auto-initialize` is enabled.

bindings/python/src/trainers.rs
adamreichold: I think is due to https://github.com/PyO3/pyo3/pull/1152/files, i.e. support for building against the stable CPython API, i.e. the last bullet point in https://pyo3.rs/v0.16.2/migration.html#runtime-changes-to-support-the-cpython-limited-api

bindings/python/Cargo.toml
adamreichold: Yes, certainly. I just mean that having it as a separate `dev-dependency` instead of enabling the feature on the original `dependency` does only help if `resolver = "2"` is enable. Otherwise, the activated features will be merge over dependencies and development dependencies.

messense: Right, I can add `resolver = "2"` if the newer Rust version requirement (1.51.0) is acceptable.

bindings/python/src/trainers.rs
messense: But that class does have `#[pyclass(module = "tokenizers.trainers")]`
https://github.com/huggingface/tokenizers/blob/cd730594e9d6b6349e4b95bc522525d7e5a178a6/bindings/python/src/trainers.rs#L18

adamreichold: Maybe something is messed by repeated specifying `name=Trainer`?

bindings/python/src/decoders.rs
```diff
@@ -161,8 +161,8 @@ impl PyByteLevelDec {
 ///     cleanup (:obj:`bool`, `optional`, defaults to :obj:`True`):
 ///         Whether to cleanup some tokenization artifacts. Mainly spaces before punctuation,
 ///         and some abbreviated english forms.
-#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name=WordPiece)]
-#[text_signature = "(self, prefix=\"##\", cleanup=True)"]
+#[pyclass(extends=PyDecoder, module = "tokenizers.decoders", name = "WordPiece")]
+#[pyo3(text_signature = "(self, prefix=\"##\", cleanup=True)")]
```
davidhewitt, line 165: @messense FYI in https://github.com/PyO3/pyo3/pull/2234 (I guess PyO3 0.16.3) the text signature can also be part of `#[pyclass()]` arguments.

bindings/python/src/tokenizer.rs
```diff
@@ -259,8 +259,24 @@ impl<'s> From<TextInputSequence<'s>> for tk::InputSequence<'s> {
...
+        }
+        if unsafe { (*arr).flags }
+            & (npyffi::NPY_ARRAY_C_CONTIGUOUS | npyffi::NPY_ARRAY_F_CONTIGUOUS)
+            == 0
+        {
+            return Err(exceptions::PyTypeError::new_err(
+                "Expected a contiguous np.array",
+            ));
+        }
+        let n_elem = unsafe { *(*arr).dimensions } as usize;
```
Narsil, line 279: I feel quite uncomfortable with adding new `unsafe` code.

It was there previously just next line, but I would rather keep those unsafe limited, and ideally having a safety note associated with them.

I think this is used for sending data with `numpy` arrays which is ease of use not the main entry point.

I will investigate a bit more this part on my end to understand what's happening too.

do you know why the approach used in `PyArrayStr` does not work ?

Narsil (commented):
Thank you so much for this PR, I wanted to do this for a long time but never got down to it,

This saves me a ton of time.

Still have lots of questions, and I fell uneasy about adding new `unsafe`, even If I understand there was some use of unsafe beforehand (but I would like to take the opportunity to understand it.

.github/workflows/python.yml
messense: No it wasn't, see this build: https://github.com/huggingface/tokenizers/runs/5624656581?check_suite_focus=true

bindings/python/src/normalizers.rs
```diff
@@ -630,8 +628,8 @@ mod test {
         let py_nfc = py_norm.get_as_subtype().unwrap();
         let gil = Python::acquire_gil();
         assert_eq!(
-            "tokenizers.normalizers.NFC",
-            py_nfc.as_ref(gil.python()).get_type().name()
+            "NFC",
+            py_nfc.as_ref(gil.python()).get_type().name().unwrap()
```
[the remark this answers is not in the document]
messense: It was switched to return a `Result`: https://docs.rs/pyo3/latest/pyo3/types/struct.PyType.html#method.name

I'm not sure about the value change though.

bindings/python/src/tokenizer.rs
messense: As @adamreichold said before, before 0.16 rust-numpy (incorrectly) did not check element type and dimension when downcasting to arrays, thus it was working by accident.

bindings/python/src/decoders.rs
messense: pyo3 0.16.3 isn't released yet anyway.

bindings/python/src/tokenizer.rs
adamreichold: > It was there previously just next line, but I would rather keep those unsafe limited, and ideally having a safety note associated with them.

Those unsafe blocks are required to check the preconditions of the existing code. `unsafe` is also non-local w.r.t function bodies, it can only be encapsulated at the module level. So there is little difference whether a function contains one or N unsafe blocks, it needs to be verified completely in both cases. (As the existing issue of assuming contiguous array without checking for that highlights.)

> do you know why the approach used in PyArrayStr does not work ?

This is just not implemented in `rust-numpy` yet, c.f. https://github.com/PyO3/rust-numpy/issues/141

.github/workflows/python.yml
Narsil: Ahhh clippy is triggering it and it doesn't work after the upgrade !

So clippy was running against 3.6 Python interpreter, which is fine, since clippy wasn't running any python code anyway.

messense: Actually clippy will trigger running build scripts, which in turn runs python code from pyo3-ffi: https://github.com/PyO3/pyo3/blob/main/pyo3-ffi/build.rs

bindings/python/Cargo.toml
messense: Reverted to only upgrade pyo3 and numpy.

bindings/python/src/tokenizer.rs
Narsil: > non-local w.r.t  + here is little difference whether a function contains one or N unsafe blocks,

Yes unsafe is tricky and non local, but just because it's non local does not mean that adding new locations doesn't have implications. 

I didn't write the first implementation, so I do not necessarily understand the implications of those unsafe calls, I need to dive anyway to understand what was there previously. Crossing the ffi boundary is unsafe and definitely needed if `rust-numpy` does not implement some functions for us. I would just much prefer understanding a bit what's going on and adding SAFETY notes so we can remove those unsafe calls later when `rust-numpy` adds more support and takes on those SAFETY problems itself (if it does).

Narsil: I dived a bit into this.

Current understanding:
- TextInputSequence seems to work fine without including any unsafe, even on `numpy` array types, while working on real numpy arrays, not sure how that casting works.
- for testing we should be careful that `numpy.array` of strings can be either  `<UXX` when it uses pure bytes and aligns to the large string in the array I guess ?, and it uses `dtype('o')` when the sizes of lists cannot create a  tensor (which is non obvious when dtypes are not explicity mentionned in the numpy array creation), those cases are handled differently in rust.
- Current proposed unsafe does only ffi numpy boundary crossing (accessing the raw numpy array to recreate everything).
- Casting from `PyArray<_>` is not possible because the equivalent of `PyObject` (Which I guess is dtype('O'), right) does not exist yet for `<UXX` (And the unsafe code would basically be the `rust-numpy` implementation iiuc.)

I still need to investigate how `TextInputSequence` can bypass numpy casting altogether, and use purely `<Vec<&PyAny>>`. I tried doing the same for `PreTokenizedInputSequence` but somehow, I seemed to receive only the first part of the array, which is confusing (I must have done it wrong).

Does `TextInputSequence` create a whole lot of copies while succeeding to avoid casting to `PyArray` ?
Fixing/modifying this exceeds the scope of this PR, so let's not include this (unless we find a very simple solution) here.

Ideally, we could remove the `ffi` boundary, but so far it's OK.

1/ As for the current code, I see a lot of pointer dereferencing. Usually I think checking for the null pointer should be done ,I am guessing `PyArray_Check`,  we could probably add that for the first unsafe comment.

2/ Would it make sense to deference numpy array metadata only once, and then do all the error checking (Like we do for `let (type_num, elsize, alignment, data) = unsafe {` ) ? I think it makes more sense, but I am not sure what's considered better here.

3/ For the last part where we're actually recreating the strings, why do we need `PyUnicode_FromUnicode` (which is deprecated) and then casting it to `PyString` before converting it over to rust String. I think I saw a while back that `Python strings` !=  `Rust strings` but I cannot find it anymore, is that the case ? Otherwise since we have the byte array, can't we purely capture them as CString to String, or are they not necessarly `\0` terminated in the numpy byte format ?

Finally as a possible safety comments, I would see 3 sections.
```rust
/// SAFETY Making sure the pointer is a valid numpy array requires calling numpy C code

/// SAFETY Getting all the metadata about the numpy array to check its sanity

/// SAFETY Looking at the raw numpy data to create new owned Rust strings via copies (so it's safe afterwards).


Is my understanding correct ? Did I miss something ?

Comment (messense) [reply bindings/python/src/tokenizer.rs]: > 3/ For the last part where we're actually recreating the strings, why do we need PyUnicode_FromUnicode (which is deprecated) and then casting it to PyString before converting it over to rust String. 

[numpy unicode](https://numpy.org/doc/stable/reference/c-api/dtype.html#c.NPY_UNICODE) is `UCS-4` while Rust string is UTF-8.

File: bindings/python/src/normalizers.rs (hunk 14 above)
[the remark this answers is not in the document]
Comment (Narsil) [reply bindings/python/src/normalizers.rs:632]: Ok, Didn't know `__qualname__` TIL.

File: bindings/python/src/tokenizer.rs (hunk 13 above)
Comment (Narsil) [reply bindings/python/src/tokenizer.rs]: Ok, diving was really important, as now much of your previous discussions make a lot more sense now !

Comment (adamreichold) [reply bindings/python/src/tokenizer.rs]: > Casting from PyArray<_> is not possible because the equivalent of PyObject (Which I guess is dtype('O'), right) does not exist yet for <UXX (And the unsafe code would basically be the rust-numpy implementation iiuc.)

Almost, the main problem is that we need to define which `T` in `PyArray<T>` safely maps to `<Un`. This might possibly be `[pyo3::ffi::Py_UCS4; n]`, but this might require const generics.

File: bindings/python/src/tokenizer.rs [hunk 15]
```diff
@@ -259,38 +259,54 @@ impl<'s> From<TextInputSequence<'s>> for tk::InputSequence<'s> {
...
+        let (type_num, elsize, alignment, data, nd, flags, n_elem) = unsafe {
             let desc = (*arr).descr;
             (
                 (*desc).type_num,
                 (*desc).elsize as usize,
                 (*desc).alignment as usize,
                 (*arr).data,
+                (*arr).nd,
+                (*arr).flags,
+                *(*arr).dimensions as usize,
```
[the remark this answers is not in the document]
messense: Good catch! Fixed.

Narsil:
Thank you so much for this !

This is a very cool and valuable PR.

In terms of merging, what I project to do is to first do a release without this change `0.12` which does contain some slight backward breaking changes (for the decoder) and the ability to drop the regex in `ByteLevel` (those are already on master). These are relatively significant changes and are necessary for HF's bigscience project https://bigscience.huggingface.co/.

I will probably wait a week or so afterwards to make sure those changes have no unintended consequences and we have a safe base for our Bigscience project.

Then I will merge this PR, and release `0.13` probably shortly after with all due tests.

McPatate (approved):
👌🏻

Narsil:
Following the trend of other HF repos we're moving the the `main` branch instead of master.

```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

Is all that is needed on your end.

h-vetinari:
Sorry, messed up the rebase. 🤦 

Fixing it

bindings/python/src/decoders.rs
h-vetinari: > pyo3 0.16.3 isn't released yet anyway.

Not sure what the timeline is now for tokenizers 0.13.0 (just saw 0.12.1 getting prepared), but if it's worth doing the text signature bits before that, I wanted to add that pyo3 0.16.3 has now been [released](https://github.com/PyO3/pyo3/releases/tag/v0.16.3) a few days ago.

h-vetinari:
Hey @Narsil 👋 

I'm happy to keep rebasing this PR, but just wanted to check where things stand currently with the plans for 0.13.0 🙃

h-vetinari:
Thanks for merging this @Narsil! :)

Any timeline for 0.13? 🙃

Narsil:
Unfortunately no definite timeline. As you might have guessed, handling `tokenizers` is only a little part of what I do at HF and releases do take quite a bit of attention.

h-vetinari:
A very gentle ping for a tokenizer release with the updated pyo3 :)

h-vetinari:
Another month, another ping... 🙃

Narsil:
Hey @h-vetinari After quite a long time (sorry, but there's definitely a lot to do and this is basically done on spare time from me.).

I wanted to release 0.13.0 today, but afaik I cannot because `manylinux2010` wheel is built with a static interpreter and I don't really know how to fix that issue.

How can we run the manylinux build and make it work.

I had 2 ideas:
- Finding some quay image with shared python interpreter (couldn't find one even in the recommended crates to do distribution)
- Removing `auto-initialize` just for those manylinux builds and place the interpreter inside for them, but it seems it needs a bit more work : https://github.com/huggingface/tokenizers/pull/1064

Narsil:
Yes it claims that it's using a static interpreter (which it is), but the feature `extension-module` should be used, which should (afaik) disable the warning, and compile properly.

PR #958, opened 2022-03-21
Update outdated dependencies and feature-gate CLI

MarcusGrass:
Making the library a bit more up to date with common dependencies.
Unsure whether special testing is required for the cli.

Added a feature-gate to cli but kept it as default. This allows someone only using the library who doesn't wish to use the binary at all to opt out using `default-features = false`

Also put dirs behind the feature gate http, as it's only used hogether with cached_path in the from_pretrained-flow

### DIFF
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -24,6 +24,7 @@ bench = false
 name = "cli"
 path = "src/cli.rs"
 bench = false
+required-features = ["cli"]
 
 [[bench]]
 name = "bpe_benchmark"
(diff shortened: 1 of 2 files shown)


Narsil:
Hi Thanks for this PR.

Do you mind splitting the feature gates from the dependencies update in separate PRs ?

It's much easier for me to add the cli-feature gate as I understand the implications more easily.

For dependency updates, we need to run a full chain of tests (in `transformers` to make sure nothing is broken there). 
This typically is done before any release too, but for dependency updates I would also do it before merging on master as it's easier then to figure out what dependency might be breaking things.

MarcusGrass:
> Hi Thanks for this PR.
> 
> Do you mind splitting the feature gates from the dependencies update in separate PRs ?
> 
> It's much easier for me to add the cli-feature gate as I understand the implications more easily.
> 
> For dependency updates, we need to run a full chain of tests (in `transformers` to make sure nothing is broken there). This typically is done before any release too, but for dependency updates I would also do it before merging on master as it's easier then to figure out what dependency might be breaking things.

No problem, I did leave the dirs feature gating in since it's only used in a single place in the codebase, which is already behind an existing feature gate. That PR is here https://github.com/huggingface/tokenizers/pull/960

Narsil:
Thanks you for changing the PR !

PR #960, opened 2022-03-22
Feature gate CLI and clap dependency

MarcusGrass:
Move CLI behind a feature so that users dependent on the lib don't pull in the clap dependency tree.
Split from https://github.com/huggingface/tokenizers/pull/958


Narsil:
Thanks you for this !

Issue #911, opened 2022-02-17
Tokenizers for Node 16?


etiennelunetta:
Currently using node 16 and I'm not able to use the latest npm release. Is it possible to rebuild for v15+?

Narsil:
Hi, this will require an update of `neon` which is the library we use for node bindings.

Unfortunately, `neon` introduces a lot of breaking changes (for the better it seems) since we made the bindings and so this is going to be a relatively big endeavor to rewrite.

Any help on that department will be appreciated.

Until then, we unfortunately can't build for Node 16.

Narsil:
When we redo the bindings, we can also think about manylinux support: https://github.com/huggingface/tokenizers/issues/972

markhughes:
But I can confirm it won't build for Node v16 in the current state. 

There are a lot of breaking changes: https://github.com/neon-bindings/neon/blob/0.10.0/MIGRATION_GUIDE_0.10.md

Would it be helpful if I wrote a script to highlight the changes and we can all start tackling them? 💪

Narsil:
Hi @markhughes 

I think given the age of neon, it might even be more practical to start over from scratch.
Maybe take the current version as a starting point but that's all.

Help on that front would be highly appreciated ! 

Don't hesitate to share work early if you want to tackle it !

rattrayalex:
For others blocked on this, [this analysis](https://github.com/transitive-bullshit/compare-tokenizers) finds that [`@dqbd/tiktoken`](https://www.npmjs.com/package/@dqbd/tiktoken) – a wasm port of OpenAI's official Rust tiktoken – works well.

MatousAc:
I'm also stuck on this. I'm using SvelteKit and thus can't use Tokenizers.
I'm still a beginner, so I would appreciate help:
I've got a [`tokenizer.json`](https://raw.githubusercontent.com/MatousAc/aqg/main/pbe/src/assets/OptimumONNX/tokenizer.json) and am about to start making my own function to convert the words into numbers (which is what it looks like the tokenizer is doing to me when I run it in python). Is there a way to use [`@dqbd/tiktoken`](https://www.npmjs.com/package/@dqbd/tiktoken) instead of Tokenizers but still use my [`tokenizer.json`](https://raw.githubusercontent.com/MatousAc/aqg/main/pbe/src/assets/OptimumONNX/tokenizer.json)? My model is based on T5, so maybe I don't even need to do this?

PR #969, opened 2022-03-31
Update the builder to use earlier windows version (2022) is not understood.

Narsil:
Error on `windows-latest` is 

```
  gyp verb find VS msvs_version not set from command line or npm config
  gyp verb find VS VCINSTALLDIR not set, not running in VS Command Prompt
  gyp verb find VS unknown version "undefined" found at "C:\Program Files\Microsoft Visual Studio\2022\Enterprise"
  gyp verb find VS could not find a version of Visual Studio 2017 or newer to use
  gyp verb find VS looking for Visual Studio 2015
  gyp verb find VS - not found
  gyp verb find VS not looking for VS2013 as it is only supported up to Node.js 8
  gyp ERR! find VS 
```

I took the short route and simply use `windows-2019` to build this, which ships with MSVC 2019 instead of 2022 and works.

The cleaner fix would be to rewrite the node bindings to newer neon bindings in order to support node 16+ (and probably newer MSVC too).

https://github.com/huggingface/tokenizers/issues/911

### DIFF
--- a/.github/workflows/node-release.yml
+++ b/.github/workflows/node-release.yml
@@ -7,8 +7,8 @@ env:
 
 on:
   push:
-    # tags:
-    #   - node-v*
+    tags:
+      - node-v*
 
 jobs:
   rust_publish:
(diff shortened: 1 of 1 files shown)


n1t0 (approved):
:ok_hand:

PR #973, opened 2022-04-05
Optimize GPT-2 regex as logic for improved performance

benwtrent:
The regex used by the BPE pre-tokenizer has fairly simple logic. The tokens are split along: letters, numbers, whitespace, symbols. Each possibly being prefixed with a `' '` and with special handling for specific apostrophe scenarios. 

Transforming this into logic instead of REGEX improves performance. Here are benchmarks (for the split function only):

Ran with 10s warmup over 100 samples of the `big.txt` data set.
```
BPE split REGEX         time:   [1.1157 s 1.1181 s 1.1207 s]

BPE split Logic         time:   [640.42 ms 645.15 ms 650.22 ms]
```

Here are some benchmarks for the overall GPT2 process (I include only the improvements, the previous run was the REGEX split)

```
BPE GPT2 encode         time:   [29.143 us 29.432 us 29.861 us]
                        change: [-12.354% -10.057% -7.6869%] (p = 0.00 < 0.05)
                        Performance has improved.

BPE GPT2 encode batch   time:   [8.8120 ms 9.0704 ms 9.3106 ms]
                        change: [-7.4765% -5.5234% -3.6632%] (p = 0.00 < 0.05)
                        Performance has improved.

BPE GPT2 encode, no cache
                        time:   [35.395 us 35.681 us 36.070 us]
                        change: [-14.634% -13.082% -11.349%] (p = 0.00 < 0.05)
                        Performance has improved.

BPE GPT2 encode batch, no cache
                        time:   [10.708 ms 10.824 ms 10.952 ms]
                        change: [-10.197% -7.7006% -5.4115%] (p = 0.00 < 0.05)
                        Performance has improved.
```

All the benchmarks were ran on my local machine:

```
MacBook Pro (16-inch, 2019)
2.4 GHz 8-Core Intel Core i9
64 GB 2667 MHz DDR4
Intel UHD Graphics 630 1536 MB
```

### DIFF
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -69,6 +69,6 @@ http = ["reqwest", "cached-path"]
 cli = ["clap"]
 
 [dev-dependencies]
-criterion = { version = "0.3", features = ["html_reports"]}
+criterion = "0.3"
 tempfile = "3.1"
 assert_approx_eq = "1.1"
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -69,6 +69,6 @@ http = ["reqwest", "cached-path"]
 cli = ["clap"]
 
 [dev-dependencies]
-criterion = "0.3"
+criterion = { version = "0.3", features = ["html_reports"]}
 tempfile = "3.1"
 assert_approx_eq = "1.1"
--- a/tokenizers/benches/bpe_benchmark.rs
+++ b/tokenizers/benches/bpe_benchmark.rs
@@ -10,19 +10,33 @@ use std::path::Path;
 use criterion::Criterion;
 use tokenizers::models::bpe::{BpeTrainerBuilder, BPE};
 use tokenizers::models::TrainerWrapper;
-use tokenizers::pre_tokenizers::byte_level::ByteLevel;
+use tokenizers::pre_tokenizers::byte_level::{ByteLevel, OptimizedByteRegex};
 use tokenizers::pre_tokenizers::whitespace::Whitespace;
 use tokenizers::tokenizer::{AddedToken, EncodeInput};
-use tokenizers::Tokenizer;
+use tokenizers::{PreTokenizerWrapper, Tokenizer};
 
 use common::{iter_bench_encode, iter_bench_encode_batch, iter_bench_train};
 use std::ops::Deref;
+use tokenizers::pre_tokenizers::sequence::Sequence;
 
 static BATCH_SIZE: usize = 1_000;
 
 fn create_gpt2_tokenizer(bpe: BPE) -> Tokenizer {
     let mut tokenizer = Tokenizer::new(bpe);
     tokenizer.with_pre_tokenizer(ByteLevel::default());
+    tokenizer.with_decoder(ByteLevel::default());
+    tokenizer.add_tokens(&[AddedToken::from("ing", false).single_word(false)]);
+    tokenizer.add_special_tokens(&[AddedToken::from("[ENT]", true).single_word(true)]);
+    tokenizer
+}
+
+fn create_optimized_gpt2_tokenizer(bpe: BPE) -> Tokenizer {
+    let mut tokenizer = Tokenizer::new(bpe);
+    let byte_level = ByteLevel::default().use_regex(false).add_prefix_space(false); 
+    tokenizer.with_pre_tokenizer(Sequence::new(vec![
+        PreTokenizerWrapper::OptimizedByteRegex(OptimizedByteRegex::default()),
+        PreTokenizerWrapper::ByteLevel(byte_level),
+    ]));
     tokenizer.with_decoder(ByteLevel::default());
     tokenizer.add_tokens(&[AddedToken::from("ing", false).single_word(false)]);
     tokenizer.add_special_tokens(&[AddedToken::from("[ENT]", true).single_word(true)]);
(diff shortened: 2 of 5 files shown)


Narsil:
Hi @benwtrent ,

Thanks very much for this PR !
Do you mind sharing the initial problem you are trying to solve ?

`10%` speedup is great ! But I fear the two codes don't necessarily do the same thing and it would need a lot more tests on the unicode range + making sure the codes do exactly the same thing.
Also speedups need to be checked on various platforms (even if we limit it to the most common ones)

Before investing tons of work to make sure things are 100% similar, can we first make the case for such a change (currently `tokenizers` should be a tiny amount of compute compared to the models that run with them) ?

Also another option would be to disable the regex `ByteLevel(use_regex=False)` and use a new `pre_tokenization` layer in `Sequence` so we can do the change more seemlessly.

benwtrent:
> Do you mind sharing the initial problem you are trying to solve ?

Speed improvements. Even onig::Regex suffers when text input is large. And this saves compute. Note: the 10% is only on the `big.txt`, of which the longest input string is only 465 words. I postulate that for longer inputs this improvement is even greater. I can try to gather some numbers around improvements on longer text inputs.

> Before investing tons of work to make sure things are 100% similar, can we first make the case for such a change 

Again, long text inputs are a big pain for BPE, specifically because the regex match can be slow, thus hampering throughput.

> Also another option would be to disable the regex

Would this contribution be more readily accepted if I added a new pre-tokenization layer that is optionally available? Maybe one that users could optionally use if they know their input texts are usually exceptionally long?

I 100% understand the concern that the tokens created are not exactly the same in all cases. I can happily add more test cases.

Narsil:
> Would this contribution be more readily accepted if I added a new pre-tokenization layer that is optionally available? Maybe one that users could optionally use if they know their input texts are usually exceptionally long?

That would make it much easier to add yes, since it would be only an opt-in feature.
We would still need to make sure both are the same, but it would definitely be easier to do.

Just to emphasize:

```python
tokenizer.pre_tokenizer = ByteLevel()

# Into
tokenizer.pre_tokenizer = Sequence([OptimizedByteRegex(), ByteLevel(use_regex=False)])
```

I am still confused how `onig` or other optimized regex engine can be "slow" on long text, afaik, there are also using a DFA just like the one you crafted. I know Python regex engine is not great at optimizing its DFA, but `onig` should do a better job at it, so I am not sure to understand why this is faster.


> I 100% understand the concern that the tokens created are not exactly the same in all cases. I can happily add more test cases.

The cases that pop into mind would be odd utf-8 whitespace, I sort of remember some categories could overlap  but cannot find relevant doc atm. Also making sure `onig` and `std::char` respect the same standard. Rust declares it follows https://docs.rs/regex/latest/regex/#unicode UTS#18 I am unsure about `onig`.

benwtrent:
@Narsil 

> I am still confused how onig or other optimized regex engine can be "slow" on long text, afaik, there are also using a DFA just like the one you crafted. 

OK, I did some more tests. Specifically, I concatenated every 1000 lines from `big.txt` to dramatically increase the length of every example. 

The percentage improvements on the "optimized" vs regular onig REGEX were the same as before. onig is very clever and efficient. 

There goes my hypothesis 🤦 .

Well, this is great news for tokenizers :D. No need for this PR. 


FYI, working in this code base is a joy. Super easy to set up, run formatting, compiling, etc. ++ developer experience.

benwtrent:
@Narsil the only "big" improvement I can think of that this logic provides is the ability to lazily stream tokens before collecting them. The REGEX has to read the entire string, where the logic can just read a handful of chars at a time and produce tokens as they are read.

However, it seems that this is not a goal of tokenizers and would be much more complicated to implement as other code would have to be changed (traits updated, etc.).

Narsil:
> FYI, working in this code base is a joy. Super easy to set up, run formatting, compiling, etc. ++ developer experience.

Music to our ears !
🤗 

> @Narsil the only "big" improvement I can think of that this logic provides is the ability to lazily stream tokens before collecting them. The REGEX has to read the entire string, where the logic can just read a handful of chars at a time and produce tokens as they are read.

This is a fantastic idea, but probably hard to implement for sure, currently the tokenizer does have to do n passes on the strings to finish it's calculation, which is not ideal. I am not sure we can 100% do it in streaming mode as some parts of pretokenization might take at the outputs of previous modifications (Again not sure, but there's definitely some code looking backwards in the string to decide what to do.)

I have to mention https://github.com/microsoft/BlingFire that claims to be faster too. I have not used it since it got out nor checked accuracy. I don't think they output incoming offsets which can definitely help with speed. 

And last but not least, when checking for compliance, we like to use the XNLI dataset, as it contains tons of unicode, and has helped us catch many odd bugs in the past.

benwtrent:
@Narsil , moved to a pre-tokenizer and did benches side by side for the optimized pretokenizer vs. not. There is almost no difference statistically.

Against "big.txt"

| Kind        | encode    | encode batch | encode, no cache | encode batch, no cache |
|-------------|-----------|--------------|------------------|------------------------|
| Regular     | 33.489 us | 9.8179 ms    | 42.064 us        | 12.178 ms              |
| "Optimized" | 31.539 us | 9.4979 ms    | 38.164 us        | 11.877 ms              |

Against "long-big.txt" (every 1k lines concatenated)

| Kind        | encode    | encode batch | encode, no cache | encode batch, no cache |
|-------------|-----------|--------------|------------------|------------------------|
| Regular     | 53.279 ms | 955.08 ms    | 66.155 ms        | 1.0737 s               |
| "Optimized" | 51.058 ms | 978.08 ms    | 61.584 ms        | 1.0252 s               |

Not 100% sure why I saw an improvement when I replaced the regex inline. I am guessing it was an artifact of running it on my machine. I double checked the inline numbers and there is no discernible difference now. 

One thing that has changed before my previous benches (showing the 10% speed improvement) and now is that there was a bug 🤦 . The bug was only adjusting the offset by `1` instead of `char.len()`. 

All in all, it shows my optimized code is not that optimized at all. `onig` is indeed about as fast as you can get it :D.

benwtrent:
@Narsil as for BlingFire being faster...

Here is their bench:

https://github.com/microsoft/BlingFire/wiki/Comparing-performance-of-Bling-Fire-and-Hugging-Face-Tokenizers

From what I can tell, they are not using the "Fast" tokenizers? Of course their optimized code would be faster than python :/

benwtrent:
Just to confirm their benchmarks, here is tokenizers vs. blingfire on `big.txt`. Each padding to 512 tokens (I am padding in tokenizers using enable_padding and not appending `0` in python directly like their previous benches):
```
time -p cat big.txt | python test_tokenizers.py -s 1
real 13.82
user 14.64
sys 0.20
 time -p cat big.txt | python test_bling.py -s 1
real 1.58
user 2.41
sys 0.19
```
Here is tokenizers with no padding
```
real 6.01
user 6.86
sys 0.20
```

Blingfire pre-allocates the entire tokenization vector with padding at once. I am guessing `tokenizers` is appending to the vector after allocating.

As for the other performance discrepancies, a deeper investigation would be required.

Narsil:
> From what I can tell, they are not using the "Fast" tokenizers? Of course their optimized code would be faster than python :/

As anything performance oriented, check for yourself on meaningful workloads for you. I haven't had time to look too deeply into their code, or their benchmarks, but as you mentioned, size of the input sentences can be a factor, padding, multi threading stuff like this.

Internally, tokenization is rarely even registering in terms of time compared to model inference/training, so we might not be as sensitive to extreme performance as other workloads. `Offsets` is super important to get correct for workloads like `NER` or question answering, in order to be able to highlight the answer in the original string (since tokenization looses a lot of information).

We know that to be a source of non-optimal runtime costs, and the others is related to working on `char` and not pure bytes (but it has other advantages, mainly we're sure that everything is still a string, and offsets are Python3  friendly.

BlingFire claims to be single pass, which I don't know how they can be on real BPE tokenizers (like GPT2, BERT is WordPiece which can be single pass). Same goes for Unigram, the algorithm itself is fast, but it cannot be expressed afaik in a  DFA sort of structure.

Another maybe unrelated topic, I actually tried earlier to make Bert (WordPiece) faster by directly using a Trie structure (sort of DFA let's say, I'm never sure of exact structure names, but you probably get the idea).

Surprisingly on our benchmarks it was slower: https://github.com/huggingface/tokenizers/pull/863

So as always for perf, measure measure measure and probably measure again :).
Also check that there's no difference in actual outcomes... These algos are easy to mess up :)

benwtrent:
OK, just for fun, I got a flamegraph BERT single encoding set up (this PR has diverged I suppose...will close it).

Here is what I found for the code on `main`:

Without padding:
flamegraph

Split seems to be the dominate thing. There are probably ways to combine multiple splits (maybe punctuation && whitespace) to reduce this cost.


With padding:
flamegraph

Padding has a crazy cost here (almost taking as much time as the encoding itself...yikes). Seems like the allocation there is wrong. It may be better to simply pass a mutable vector that is defaulted with the padded values and allow the encoder to write to it. Obviously, that trick cannot be done in batch as you likely want to pad to the maximum encoded sequence length.


But, as you rightly said, 
> tokenization is rarely even registering in terms of time compared to model inference/training, so we might not be as sensitive to extreme performance as other workloads

The runtime costs are probably dominated by the cost of inference && training. 

It's just optimization problems can be addictive! :D


Gonna close this PR as it obviously isn't providing value. But, it was a great learning experience and again, it was fun to play around here.

Narsil:
> It's just optimization problems can be addictive! :D

I feel you :)

Issue #873, opened 2022-01-06
Add a `Sequence` to the processors


SaulLu:
I think it could be useful to have a `Sequence` object for (post)-processors. 

Indeed, I think a tokenizer might need to combine the `ByteLevel` post-processor and a `TemplateProcessing` processor.

Today `RobertaProcessing` is a processor that combines these 2 features in some way but we can't use our own template.

What do you think? :smile:

Narsil:
In principle, very favorable.

In practice:

```python
def process(encoding: Encoding, pair_encoding: Encoding, add_special_tokens: bool) -> Encoding
```

So with this signature, it's destructive show not really obvious how we should "chain" them.
If we do 

```python
for processor in processors:
    encoding = processor(encoding, pair_encoding, add_special_tokens)
```
Then is means, only the first one will ever know that there was a pair meaning the order of the sequence will be important in a non-obvious way IMO.

What do you think ? Should we still do it ?

Food for thought: There is request to support more than a pair of encodings: https://github.com/huggingface/tokenizers/issues/804. Maybe to take into consideration if we envision making bigger changes to `Processor` to enable sequencing them.

thomasw21:
@Narsil is there a reason no to do
```python
for processor in processors:
    encoding, pair_encoding = processor(encoding, pair_encoding, add_special_tokens)
# This last one is really just needed for backward compatilibity, pair_encoding is None in this case.
encoding, pair_encoding = post_process.default_process(encoding, pair_encoding, add_special_tokens)
```
I feel you could actually remove `<dyn PostProcessor>::default_process(encoding, pair_encoding, add_special_tokens)` from most pre-processors. and add it at the end of the sequence. Typically `ByteLevel` become `Sequence([ByteLevel, MergePair])`, but in order not to break backward compat we can do the top solution?

Narsil:
This is roughly what I suggested to @mishig25 as a solution. 

Using `Vec<Encoding>` actually instead of pairs since pair is also limiting in some form (#804 ) but roughly it's this.

I also suggested a way to implement it in a  non backward incompatible way (like I wanted to do for `Decoder` trait)

SaulLu:
Let me just re-share here a design idea that we had - I think - also discussed offline a long time ago: add a new method to all processors - for example named `process_chain` - which will have another composable signature and which will be the method called by `processors.Sequence`. :blush:

thomasw21:
[...]

PR #1005, opened 2022-05-27
Add `postprocessor::Sequence`

mishig25:
Closes https://github.com/huggingface/tokenizers/issues/873 by implementing `preprocessor::Sequence`

[as suggested](https://github.com/huggingface/tokenizers/issues/873#issuecomment-1139511282) by @SaulLu, 
> Let me just re-share here a design idea that we had - I think - also discussed offline a long time ago: add a new method to all processors - for example named process_chain - which will have another composable signature (a list) and which will be the method called by processors.Sequence. 😊

Please let me know if proposed changes look good, if so, I will add tests. Otherwise, I'm happy to make any changes as necessary. Also, please see the comments as I had some questions


tokenizers/src/tokenizer/mod.rs
```diff
@@ -100,6 +104,14 @@ pub trait PostProcessor {
         pair_encoding: Option<Encoding>,
         add_special_tokens: bool,
     ) -> Result<Encoding>;
+    /// Process method that is used when a processor is part of `processors::sequence::Sequence`
+    fn process_chain(
+        &self,
+        _encodings: Vec<Encoding>,
+        _add_special_tokens: bool,
+    ) -> Result<Vec<Encoding>> {
```
mishig25, line 112: @Narsil this is the preliminary signature we agreed upon. Please check its implementation & let me know if there is anything you wanna change

tokenizers/src/processors/template.rs
```diff
@@ -647,6 +648,26 @@ impl PostProcessor for TemplateProcessing {
...
         )
     }
+
+    fn process_chain(
+        &self,
+        encodings: Vec<Encoding>,
+        add_special_tokens: bool,
+    ) -> Result<Vec<Encoding>> {
+        let is_pair = encodings.len() == 2;
+        if is_pair {
```
mishig25, line 658: wdyt about this implementation & hardcoded number 2. 
Basically, I'm saying: if there are 2 elements in `encodings` vector, then run `self.process(encoding, pair_encoding)` with **pair_encoding** encoding.

otherwise, run `self.process(encoding, pair_encoding=None)` for each element in `encodings`

thomasw21:
Not sure I understand why we use `process_chain` ... I think all `fn process` should take `encodings: Vec<Encoding>` instead, and `Sequence` is just a for loop that calls `process` instead of adding a new endpoint?

tokenizers/src/processors/sequence.rs
```diff
@@ -0,0 +1,138 @@
...
+        match encodings.len() {
+            1 => Ok(encodings
+                .pop()
+                .ok_or(SequenceProcessorError::EncodingsVecPop)?),
+            _ => {
+                // merge encodings
+                for (i, encoding) in encodings.iter_mut().enumerate() {
+                    encoding.set_sequence_id(i);
+                }
+                Ok(Encoding::merge(encodings, false))
```
mishig25, line 56: to emulate https://github.com/huggingface/tokenizers/blob/bb6fea033932ff1fb23807a2b7c5376f5a7c4786/tokenizers/src/tokenizer/mod.rs#L123-L125

mishig25:
@Narsil please feel free to do another round of review. 
The things I've changed based on our huddle on friday are:
1. `process_chain` should not use/call `process`
2. `process_chain` should not be optional (i.e. every preprocessor must implement `process_chain` method)
3. added tests preprocessor (for bytelevel, template, sequence)

Narsil (commented):
Seems a fine direction but we need to remove the code within the `process` function in each implementor of the trait no ?

tokenizers/src/pre_tokenizers/byte_level.rs
```diff
@@ -194,6 +194,24 @@ impl PostProcessor for ByteLevel {
...
+                process_offsets(encoding, self.add_prefix_space);
+                encoding
+                    .get_overflowing_mut()
+                    .iter_mut()
+                    .for_each(|encoding| process_offsets(encoding, self.add_prefix_space));
+            }
+        }
+
+        Ok(encodings)
+    }
```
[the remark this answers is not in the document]
mishig25: do you mean, we should change `process` to:
```rust
fn process(
    &self,
    mut encoding: Encoding,
    mut pair_encoding: Option<Encoding>,
    add_special_tokens: bool,
) -> Result<Encoding> {
    ...
    self.process_chain(...)
    ...
}
```
?

bindings/python/src/processors.rs
```diff
@@ -68,6 +70,14 @@ impl PostProcessor for PyPostProcessor {
         self.processor
             .process(encoding, pair_encoding, add_special_tokens)
     }
+
+    fn process_chain(
```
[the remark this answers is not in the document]
thomasw21: Ah so I see, thanks! Would be down to know if it would make sense to build such a feature where we can construct `private abstract methods`

SaulLu:
Hello to all! 

Thanks a lot again @mishig25 for taking the lead on this feature! :raised_hands: 

I was wondering where we are on this feature! What is currently the sticking point? Is there anything we need to discuss?

mishig25:
Superceded by 1047

PR #1008, opened 2022-06-01
Changing `Decoder` trait to be more composable. (#938)

### LINKED ISSUE #872: Add an `Sequence` object to the decoders

Narsil:
* Changing `Decoder` trait to be more composable.

### DIFF
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -119,9 +119,13 @@ impl dyn PostProcessor {
     }
 }
 
-/// A `Decoder` has the responsibility to merge the given `Vec<String>` in a `String`.
+/// A `Decoder` changes the raw tokens into its more readable form.
 pub trait Decoder {
-    fn decode(&self, tokens: Vec<String>) -> Result<String>;
+    fn decode(&self, tokens: Vec<String>) -> Result<String> {
+        let results = self.decode_chain(tokens)?;
+        Ok(results.join(""))
+    }
+    fn decode_chain(&self, tokens: Vec<String>) -> Result<Vec<String>>;
 }
 
 /// A `Trainer` has the responsibility to train a model. We feed it with lines/sentences
--- a/bindings/node/lib/bindings/decoders.d.ts
+++ b/bindings/node/lib/bindings/decoders.d.ts
@@ -49,3 +49,9 @@ export function ctcDecoder(
   word_delimiter_token?: string,
   cleanup?: boolean
 ): Decoder;
+
+/**
+ * Instantiate a new Sequence Decoder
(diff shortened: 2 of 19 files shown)


mishig25:
Got 2 questions:
1. What is the difference between #938 & #1008 and why was #938 reverted in the first place?
2. Do you need to create `tokenizers::decoders::Sequence`?

Narsil:
> Got 2 questions:
> 
>     1. What is the difference between [Changing `Decoder` trait to be more composable. #938](https://github.com/huggingface/tokenizers/pull/938) & [Changing `Decoder` trait to be more composable. (#938) #1008](https://github.com/huggingface/tokenizers/pull/1008) and why was [Changing `Decoder` trait to be more composable. #938](https://github.com/huggingface/tokenizers/pull/938) reverted in the first place?
> 

This one is not breaking
>     2. Do you need to create `tokenizers::decoders::Sequence`?
Yes.

tokenizers/src/tokenizer/mod.rs
```diff
@@ -119,9 +119,13 @@ impl dyn PostProcessor {
     }
 }
 
-/// A `Decoder` has the responsibility to merge the given `Vec<String>` in a `String`.
+/// A `Decoder` changes the raw tokens into its more readable form.
 pub trait Decoder {
-    fn decode(&self, tokens: Vec<String>) -> Result<String>;
+    fn decode(&self, tokens: Vec<String>) -> Result<String> {
+        let results = self.decode_chain(tokens)?;
+        Ok(results.join(""))
```
mishig25, line 126: will this line cause any problem for the SequenceDecoder? 
Specifically, one unlikekly edge-case I'm thinking is: one of the sequence decoders is another sequence decoder, and `Ok(results.join(""))` will cause an informaition loss?

as a solution, we can implement `fn decode` for Sequence Decoder?

Narsil: It's OK I think.

The parent `Sequence` would call the child `Sequence.decode_chain` NOT the `decode` function, so we're good, no ?

I could add a test to make sure.

mishig25: ooh I see, so it is no problem in that case

mishig25 (approved):
lgtm!

Issue #935, opened 2022-03-01
Support `wasm`


Narsil:
Wasm support would be a cool issue.

- [ ] Add a feature flag `wasm`.
- [ ] Use `esaxx_rs::suffix_rs` instead of `esaxx_rs::suffix` (Maybe with a change in the crate features too to prevent C compilation)
- [ ] Find a good workaround `onig` compilation. Dropping support for any element that depends on it is the simplest, Using something like `emscripten` might be done too https://github.com/rustwasm/team/issues/291 . Big conversation around this: https://github.com/huggingface/tokenizers/issues/63
- [ ] Add a simple example on how to use when it works.

mbrunel:
Hey,
I need this feature. Is someone working on it? Otherwise I can take it over if you still want it.

mbrunel:
Thanks for the answer,
I did this : https://github.com/mithril-security/tokenizers-wasm.git
It compiles the whole library to wasm (using josephrocca's method for the regex abstraction) but only exposes very few features, nonetheless it has the merits of :

- 1 : working (seemingly)
- 2 : answering my original use-case

Do you think this is the way to go? And if so do you think it would be of interest for people if I (or others) were to do more?

Narsil:
Hi @mbrunel ,

Actually I started some work in #1009 to integrate fully the work as this feature seemed to have more traction that I expected (and it was a fun coding thing to do).

Main differences  is `pure_rust` is not nice I think since it subtracts from `esaxx-rs` so I went instead with a default `cpp` feature, that I disable for `wasm`.

Then I went with something similar for the regex abstraction but I tried to remove the copy and have as little change as much as possible.
I didn't try all features within the `wasm` example project. Your example pulling a tokenizer and running it is nice, would you be ok if I stole it ?

I took the liberty of co authoring both you and @josephrocca since I clearly stole some ideas.

Feel free to comment on the PR directly if there's any questions or things that I didn't do properly, I very much just followed the `wasm` tutorial.

mbrunel:
```
Actually I started some work in https://github.com/huggingface/tokenizers/pull/1009 to integrate fully the work as this feature seemed to have more traction that I expected (and it was a fun coding thing to do).
```
Nice

```
we could also point to your tokenizers-wasm entire directory if you want to add more features and make it more like a binding to the entire API.
```
[...]

PR #1009, opened 2022-06-06
Adding `unstable_wasm` feature + example to run `tokenizers` on wasm.

Narsil:
Should get #935 started.

Ideally we should test compliance of some well known tokenizers
Maybe a small tokenizer training.

Feature is named `unstable_wasm` (as well as the example). Should at least make sure people understand that 1:1 parity is NOT to be expected.

The reason is that `onig` the current regexp engine has to be swapped out for `fancy_regex` which *should* match, but we have no guarantee (and onig is a C dependency without any `wasm` support that we know of).

The `esaxx-rs` has been modified to allow dropping out of the C dependency requirements, it uses the rust code (which is slower than it's cpp counterpart). Then all optional features are disabled for wasm `cli`, `http` and `progressbar`.

This PR also adds `esaxx_fast` feature to enable easier control of the C version of the `esaxx-rs` code. Since features are additive there was no way to remove the `cpp`  feature only for `wasm`. So instead there's a new feature within the default feature, which gets disabled by the example `default-features = false`.


tokenizers/examples/unstable_wasm/.appveyor.yml
```diff
@@ -0,0 +1,11 @@
+install:
+  - appveyor-retry appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
```
McPatate, line 2: Why use appveyor for the CI ?

tokenizers/examples/unstable_wasm/README.md
```diff
@@ -0,0 +1,69 @@
+<div align="center">
```
McPatate, line 1: We also may want to change the README's content at some point ^^

tokenizers/examples/unstable_wasm/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+
+#[wasm_bindgen]
+extern "C" {
+    fn alert(s: &str);
+}
+
+#[wasm_bindgen]
+pub fn greet() {
+    alert("Hello, {{project-name}}!");
+}
```
McPatate, line 21: may want to remove this

tokenizers/src/pre_tokenizers/split.rs
```diff
@@ -1,4 +1,4 @@
-use onig::Regex;
+use crate::utils::SysRegex as Regex;
```
McPatate, line 1: `SysRegex` is used in other places of the code, won't this alias make the code a bit confusing ? Either work for me, `SysRegex` or `Regex`

McPatate (approved):
IMO we should remove unnecessary files like licenses and CI stuff in the example.

Was the deletion of the `Cargo.lock`s in the bindings intended ?

Also I'm curious, what does `Sys` in `SysRegex` stand for ?

Cool to bring tokenizers to wasm ! 🔥

tokenizers/examples/unstable_wasm/.appveyor.yml
Narsil: It's in the default template for `wasm` app. I don't really see a lot of downsides of keeping them.

tokenizers/src/pre_tokenizers/split.rs
Narsil: There's a lot of `Regex` everywhere in code. 

`SysRegex` tends to imply `System` meaning the real underlying regex engine.
Happy to put a better name.

And I should rename everything instead of using `as Regex` it was because of laziness :)

tokenizers/examples/unstable_wasm/.appveyor.yml
McPatate: I'd leave it up to the user to know or not to add these kinds of files, they don't showcase anything particular to `tokenizers` and are readily available on `wasm` tutorials. But that's just my 2 cents !

tokenizers/examples/unstable_wasm/README.md
Narsil: Added a simple line so it's not the pure wasm tutorial (commands explained in it are still useful I find)

McPatate (approved):
lgtm !

PR #1024, opened 2022-07-11
Add `from_bytes` approach for creating tokenizers

### LINKED ISSUE #1013: Load pre-trained tokenizer from memory
Hello guys,

`tokenizers::Tokenizer` has two methods to create a pre-trained tokenizer: `Tokenizer::from_pretrained` and `Tokenizer::from_file`. It's quite common that we need to download `tokenizer.json` from remote, now we have to save the remote data to a local file and then load the tokenizer from local file. We also have to delete it after loading. 

If we have something like `Tokenizer::from_memory(data_bytes: &[u8])`, we can just download the remote tokenizer to memory and then load it directly. It's quick and safe, you'll never need to delete the local file after using.

HaoboGu:
This PR fixes #1013.
Tokenizer can be created from bytes in memory using added `Tokenizers::from_bytes`.

Signed-off-by: HaoboGu <haobogu@outlook.com>

### DIFF
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -402,6 +402,10 @@ impl Tokenizer {
         let tokenizer = serde_json::from_str(&content)?;
         Ok(tokenizer)
     }
+    pub fn from_bytes<P: AsRef<[u8]>>(bytes: P) -> Result<Self> {
+        let tokenizer = serde_json::from_slice(bytes.as_ref())?;
+        Ok(tokenizer)
+    }
     #[cfg(feature = "http")]
     pub fn from_pretrained<S: AsRef<str>>(
         identifier: S,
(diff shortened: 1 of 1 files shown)


tokenizers/src/tokenizer/mod.rs
```diff
@@ -402,6 +402,10 @@ impl Tokenizer {
         let tokenizer = serde_json::from_str(&content)?;
         Ok(tokenizer)
     }
+    pub fn from_bytes<P: AsRef<[u8]>>(bytes: P) -> Result<Self> {
+        let tokenizer = serde_json::from_slice(bytes.as_ref())?;
+        Ok(tokenizer)
```
Narsil, line 407: NIT: `serde_json::from_slice(bytes.as_ref())` is IMO slightly better ;)

Narsil (approved):
LGTM, left a nit (feel free to ignore it if you don't agree).

Narsil:
We can consider the tests wokring, this is linked to clippy update.

tokenizers/src/tokenizer/mod.rs
HaoboGu: I tried `serde_json::from_slice(bytes.as_ref())`, but rust compiler reports error:

```
mismatched types
expected enum `std::result::Result<tokenizer::Tokenizer, std::boxed::Box<(dyn std::error::Error + std::marker::Send + std::marker::Sync + 'static)>>`
   found enum `std::result::Result<_, serde_json::Error>`
```

Issue #1029, opened 2022-07-20
Please update derive_builder to 0.10


kaustavha:
Tokenizers currently uses derive_builder 0.9 in rust. There is a 0.10 release for it out which should not cause any breaking changes. 
Would be nice to get it updated.

PR #1030, opened 2022-07-20
Update derive_builder to 0.10 in cargo.toml

kaustavha:
Update derive_builder to 0.10 in cargo.toml. 
Related to issue #1029


Narsil:
Seems it's not as innocuous as it appears. :(

PR #1034, opened 2022-07-26
feat(ci): add macos arm64 runner

McPatate:
I've added an arm64 runner that you can check out in the settings tab.

Edited the CI to make it release arm64 bins !

### DIFF
--- a/.github/SETUP_RUNNER.md
+++ b/.github/SETUP_RUNNER.md
@@ -1,0 +1,8 @@
+[Install homebrew](https://brew.sh/)
+
+then :
+```sh
+brew install python@3.8 python@3.9 python@3.10
+pip3 install awscli
+```
+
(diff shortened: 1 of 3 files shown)


Narsil:
Do you mind creating an actions where this actually ran to check what actually happened.

Just comment out the upload to pypi part, so we can iterate on it.

Narsil:
Arm64 node build can be ignored. If you can just remove it from the automated build that'd be ok :)

We need to rewrite the node bindings at some point anyway (but seems like a lot of efforts :( )

McPatate:
> We need to rewrite the node bindings at some point anyway (but seems like a lot of efforts :( )

Does anyone use them ?

Narsil:
> Does anyone use them ?

There are some issues open yes, but as we're lagging behind in node versions it's harder and harder to use them
https://github.com/huggingface/tokenizers/issues?q=is%3Aissue+is%3Aopen+node

Narsil:
Let's close in favor of https://github.com/huggingface/tokenizers/pull/1055 if you're OK with it.

PR #1047, opened 2022-08-23
Modify `Processor` trait to support chaining.

Narsil:
0 modifications yet, everything will consume the vector.
Every test should be green without any modifications.

I wanted to resurrect https://github.com/huggingface/tokenizers/pull/1005 but I found myself having a real hard time making modifications and understanding what was what and why things were different when they were.

Instead I decided to start from scratch heavily inspiring from the PR instead.

This PR is the first step. It modifies the trait, and that' s it.

- Next step making the various ` process_encodings`  actually agnostic (meaning not fusing their returned results into 1 encoding when they can).
- Then we can add `Sequence`.

Each step will be in a different PR.

Edit companion PR for code linting : https://github.com/huggingface/tokenizers/pull/1048

### DIFF
--- a/tokenizers/src/processors/bert.rs
+++ b/tokenizers/src/processors/bert.rs
@@ -1,4 +1,4 @@
-use crate::tokenizer::{Encoding, PostProcessor, Result};
+use crate::tokenizer::{Encoding, PostProcessor, ProcessorError, Result};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::iter::FromIterator;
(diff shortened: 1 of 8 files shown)


tokenizers/src/processors/bert.rs
```diff
@@ -34,20 +40,34 @@ impl PostProcessor for BertProcessing {
...
-                add_special_tokens,
-            );
+            return Ok(encodings);
         }
 
+        let (mut encoding, pair_encoding): (Encoding, Option<Encoding>) = match encodings.len() {
+            1 => (
+                encodings
+                    .pop()
+                    .ok_or(ProcessorError::InvalidEncodingsVecLength)?,
```
mishig25, line 56: nit: can you just do `pop().unwrap()` here because you've already checked for length `encodings.len() == 1` or `ok_or` would still be better?

mishig25 (approved):
lgtm!

tokenizers/src/processors/bert.rs
Narsil: I am removing this in the follow up PR anyway so shouldn't matter.

mishig25: oh I see 😊

SaulLu:
Thank you @Narsil :raised_hands:

PR #1048, opened 2022-08-23
Updating code according to clippy.

Narsil:
- Adding `Eq` where possible
- Denied the ref deref warnings as it was spamming and solution not
  really better.

### DIFF
--- a/bindings/node/native/Cargo.lock
+++ b/bindings/node/native/Cargo.lock
@@ -847,18 +847,19 @@ dependencies = [
 
 [[package]]
 name = "macro_rules_attribute"
-version = "0.0.2"
+version = "0.1.2"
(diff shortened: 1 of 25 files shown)


mishig25 (approved):
lgtm! 

What does it mean when `#[derive(PartialEq, Eq)]` both `PartialEq` & `Eq` are present? Shouldn't they be mutually exclusive, like if you have `Eq`, what is the point of having `PartialEq` as well?

Narsil:
@mishig25 ,

It's actually a `cargo clippy` warning now. If your are `PartialEq` and all your members are `Eq` then you can be `Eq`. But not every structure can be `Eq`: https://users.rust-lang.org/t/what-is-the-difference-between-eq-and-partialeq/15751

PR #1051, opened 2022-08-24
Making `process_encodings` not eat up the encodings any more.

Narsil:
This actually changes implementation (so real risk of  backward regression).

The idea is that the processors should not (as much as possible), fuse the encodings themselves, everything should be taken care of by `Encoding::merge(encodings)` instead. So they only need to create the encodings and modify them accordingly.

Benchmark is unclear: 

```python
WordPiece BERT encode   time:   [25.492 us 26.008 us 26.442 us]
                        change: [+9.6881% +12.165% +14.652%] (p = 0.00 < 0.05) 

    Performance has regressed.
----
WordPiece BERT encode batch
                        time:   [10.383 ms 10.615 ms 10.816 ms]
                        change: [-0.3604% +3.1938% +6.4601%] (p = 0.07 > 0.05)

   No change in performance detected.

----
BPE GPT2 encode         time:   [15.162 us 15.399 us 15.644 us]
                        change: [-5.9290% -3.5010% -1.0007%] (p = 0.01 < 0.05)

     Performance has improved.**
----
BPE GPT2 encode batch   time:   [9.1476 ms 9.3622 ms 9.6480 ms]                            
                        change: [+7.9795% +10.891% +13.714%] (p = 0.00 < 0.05)

Performance has regressed.
----
BPE GPT2 encode, no cache
                        time:   [27.387 us 28.347 us 29.280 us]
                        change: [+1.1100% +3.9995% +6.8846%] (p = 0.01 < 0.05)

   Performance has regressed.
----
BPE GPT2 encode batch, no cache
                        time:   [12.613 ms 12.986 ms 13.459 ms]
                        change: [-3.6782% -0.7142% +2.6011%] (p = 0.68 > 0.05)

    No change in performance detected.
```

BPE GPT2 doesn't use PostProcessor and seem both improved and worse ?? 
WordPiece Bert seems roughly slower (it uses `BertProcessing` which has been modified).

It's not entirely clear why it's slower, will try and do some simple modifications to recover perf but not too much.

### DIFF
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -34,6 +34,10 @@ harness = false
 name = "bert_benchmark"
 harness = false
 
+[[bench]]
+name = "layout_benchmark"
+harness = false
+
 [dependencies]
 lazy_static = "1.4"
 rand = "0.8"
--- a/tokenizers/benches/layout_benchmark.rs
+++ b/tokenizers/benches/layout_benchmark.rs
@@ -1,8 +1,6 @@
 #[macro_use]
 extern crate criterion;
 
-mod common;
-
 use std::fs::File;
 use std::io::{BufRead, BufReader};
 use std::path::Path;
@@ -10,7 +8,6 @@ use std::time::{Duration, Instant};
 
 use criterion::black_box;
 use criterion::Criterion;
-use tokenizers::normalizers::NormalizerWrapper;
 use tokenizers::processors::template::TemplateProcessing;
 use tokenizers::{EncodeInput, Encoding, PostProcessor, Tokenizer};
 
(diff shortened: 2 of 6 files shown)


Narsil:
Narrower benchmark: 

```
TemplateProcessing single encode
                        time:   [1.1407 us 1.1499 us 1.1626 us]
                        change: [+30.191% +36.538% +43.271%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 20 measurements (10.00%)
  2 (10.00%) high severe

TemplateProcessing pair encode
                        time:   [3.1768 us 3.2029 us 3.2455 us]
                        change: [+55.745% +70.071% +84.622%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

Different is that `Encoding::merge` is iterative instead of allocating all the vectors at once.
But the overal time is 1-3us, which is .. well not a lot.

I won't fix this (will add a TODO comment though so future readers can know how to fix later.)

mishig25 (approved):
lgtm!

PR #1052, opened 2022-08-24
Adding `Sequence` for `PostProcessor`.

Narsil:
Actually adding the `Sequence`.

Added a dummy  test since actually we didn't need it for OPT.
Still a nice addition for future models/ease of use.

### DIFF
--- a/bindings/node/lib/bindings/post-processors.d.ts
+++ b/bindings/node/lib/bindings/post-processors.d.ts
@@ -54,3 +54,11 @@ export function templateProcessing(
   pair?: string,
   specialTokens?: [string, number][]
 ): PostProcessor;
+
+/**
+ * Instantiate a new SequenceProcessing.
+ *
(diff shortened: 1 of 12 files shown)


mishig25 (approved):
lgtm!

Got one question: can there be a bad or incompatible sequence (for example: `Sequence::new(vec![TemplateProcessing, ByteLevel])`) or any sequence woks (i.e. permutationcombination of all processors should work) ?

Narsil:
> Got one question: can there be a bad or incompatible sequence (for example: Sequence::new(vec![TemplateProcessing, ByteLevel])) or any sequence woks (i.e. permutationcombination of all processors should work) ?

Yes combinations can be "bad" in the sense that I am not really sure what they are supposed to do, but ultimately they are separate in the sense that none actually consumes the encodings until the very last step.

mishig25:
got it 👍

PR #1054, opened 2022-08-24
Modify `Processor` trait to support chaining.

Narsil:
0 modifications yet, everything will consume the vector.
Every test should be green without any modifications.


Narsil:
Force approving because of this: https://github.com/huggingface/tokenizers/pull/1047

PR #1055, opened 2022-08-25
Adding m1 build to the release process for Python.

Narsil:
This was tested here: https://github.com/huggingface/tokenizers/runs/8010901199?check_suite_focus=true

Superseeds: https://github.com/huggingface/tokenizers/pull/1034
Fixes: https://github.com/huggingface/tokenizers/issues/932

### DIFF
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -152,7 +152,7 @@ jobs:
   upload_package:
     name: Upload package to PyPi
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


McPatate (approved):
Nice ! Thanks for taking care of this.

Is it repeatable ? Since the worker isn't cleaned after each run.

Maybe we should add a cleanup task.

Narsil:
> Is it repeatable ? Since the worker isn't cleaned after each run.

What do you mean ? The python versions are hard installed in `pyenv`  which makes cleanup not super necessary (the deps are hard upgraded before every run + the checkout means the thing is going to be rebuilt)

Narsil:
I restarted the jobs, they seem to work.

PR #1060, opened 2022-09-16
Turns out we introduced a regression because bad code.

Narsil:


McPatate (approved):

mishig25 (approved):
lgtm! curious how this issue was discovered?

Narsil:
https://huggingface.slack.com/archives/C02CH2YP4EQ/p1663273001253299
(Sorry it's an internal link)

Here is the PR I use to run the full test suite: https://github.com/huggingface/transformers/pull/16712

Basically this was caught by a slow pipeline test where the output didn't match the previous version because we rarely test `WordPiece` decoder, but it seems it's used in the wild and in slow tests ;)

We're running the full test suite before release now on `transformers`

PR #1063, opened 2022-09-19
Removing python3.6 from manylinux it's not supported anymore.

Narsil:
https://github.com/huggingface/tokenizers/actions/runs/3081237199/jobs/4979550754
https://github.com/huggingface/tokenizers/actions/runs/3081237204/jobs/4979550056

### DIFF
--- a/bindings/python/test2.py
+++ b/bindings/python/test2.py
@@ -1,0 +1,4 @@
+from tokenizers import Tokenizer
+
+
+tokenizer = Tokenizer.from_file("/home/nicolas/Downloads/tokenizer-wiki.txt")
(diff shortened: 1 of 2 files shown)


bindings/python/test2.py
```diff
@@ -0,0 +1,4 @@
+from tokenizers import Tokenizer
+
+
+tokenizer = Tokenizer.from_file("/home/nicolas/Downloads/tokenizer-wiki.txt")
```
mishig25, line 4: Is this file should be added/commited `bindings/python/test2.py`?
If it should be added, it has an absolute path on your 
machine "home/nicolas/Downloads/tokenizer-wiki.txt"

Narsil: I need to remove that file.

mishig25: perfect!

PR #1064, opened 2022-09-19
Enabling static interpreter embedding for manylinux.

Narsil:


mishig25:
Is it for https://pyo3.rs/latest/features.html#features-for-extension-module-authors ? 

> [extension-module](https://pyo3.rs/latest/features.html#extension-module)
This feature is required when building a Python extension module using PyO3.
>
> It tells PyO3's build script to skip linking against libpython.so on Unix platforms, where this must not be done.
>
> See the [building and distribution](https://pyo3.rs/latest/building_and_distribution.html#linking) section for further detail.

Narsil:
It's more about this https://pyo3.rs/latest/building_and_distribution.html#embedding-python-in-rust

PR #1066, opened 2022-09-20
pyo3: update to 0.17

davidhewitt:
Still a WIP; doing this I found an unintended API change in PyO3 which I think I can push a patch release for.


bindings/python/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "tokenizers-python"
 version = "0.11.0"
 authors = ["Anthony MOI <m.anthony.moi@gmail.com>"]
-edition = "2018"
+edition = "2021"
```
davidhewitt, line 5: @Narsil if you cherry-pick [8691345](https://github.com/huggingface/tokenizers/pull/1066/commits/8691345d733b40c3923b09d67b7e795024993e06) onto your branch for #1064 I think it may fix the issue.

The thinking is thatedition 2021 uses Cargo's newer `resolver = "2"` which won't add the `auto-initialize` feature from the pyo3 dev-dependency during the manylinux build, which I think will avoid the problem.

Narsil:
Hey, 

Thanks to your work I was able to make it work (by removing the `dev-dependencies` with `auto-initialize`). I added the dependency as a pure feature, which seems to not work properly when doing `cargo test --features test` but this is not the most important to fix soon (there's a bunch of failures to link to python functions, even if I am in a shared Python 3.10.7 environement).

 Thanks a ton for this work, and I'm all up for upgrading to `0.17` !

Narsil:
https://github.com/huggingface/tokenizers/actions/runs/3104447752/jobs/5028936526

davidhewitt:
@Narsil I've since released PyO3 0.17.2 and this branch now builds for me locally, should be good for review and also it would be great to have the CI run approved to confirm it passes your CI.

Narsil:
Hey @davidhewitt Thanks for the work.

I took the liberty of rebasing (to prevent merge issues) and modifying the makefile line that was failing !
If this works this will get merged !

PR #1072, opened 2022-09-26
Fixing roberta type id (everything is zero).

Narsil:
The original Roberta `type_ids` actually gives out `0` for every part:

https://github.com/huggingface/tokenizers/blob/python-v0.12.1/tokenizers/src/processors/roberta.rs#L117

This broke 1 integration test in `transformers`.

```py
from transformers import AutoTokenizer, LayoutLMForQuestionAnswering
from datasets import load_dataset
import torch

tokenizer = AutoTokenizer.from_pretrained("impira/layoutlm-document-qa", add_prefix_space=True)
model = LayoutLMForQuestionAnswering.from_pretrained("impira/layoutlm-document-qa", revision="1e3ebac")

dataset = load_dataset("nielsr/funsd", split="train")
example = dataset[0]
question = "what's his name?"
words = example["words"]
boxes = example["bboxes"]

encoding = tokenizer(
    question.split(), words, is_split_into_words=True, return_token_type_ids=True, return_tensors="pt"
)
print(encoding["input_ids"])
print(encoding.sequence_ids(0))
bbox = []
for i, s, w in zip(encoding.input_ids[0], encoding.sequence_ids(0), encoding.word_ids(0)):
    if s == 1:
        bbox.append(boxes[w])
    elif i == tokenizer.sep_token_id:
        bbox.append([1000] * 4)
    else:
        bbox.append([0] * 4)
encoding["bbox"] = torch.tensor([bbox])

word_ids = encoding.word_ids(0)
outputs = model(**encoding)
loss = outputs.loss
start_scores = outputs.start_logits
end_scores = outputs.end_logits
start, end = word_ids[start_scores.argmax(-1)], word_ids[end_scores.argmax(-1)]
data = " ".join(words[start : end + 1])
print(repr(data))
```

### DIFF
--- a/bindings/python/Cargo.lock
+++ b/bindings/python/Cargo.lock
@@ -1706,7 +1706,7 @@ checksum = "cda74da7e1a664f795bb1f8a87ec406fb89a02522cf6e50620d016add6dbbf5c"
 
 [[package]]
 name = "tokenizers"
-version = "0.13.0"
+version = "0.13.1"
 dependencies = [
  "aho-corasick",
  "cached-path",
@@ -1739,7 +1739,7 @@ dependencies = [
 
 [[package]]
 name = "tokenizers-python"
-version = "0.13.0"
+version = "0.13.1"
 dependencies = [
  "env_logger",
  "itertools 0.9.0",
--- a/bindings/python/Cargo.toml
+++ b/bindings/python/Cargo.toml
@@ -14,7 +14,7 @@ serde = { version = "1.0", features = [ "rc", "derive" ]}
 serde_json = "1.0"
 libc = "0.2"
 env_logger = "0.7.1"
-pyo3 = { version = "0.16.2", features = ["extension-module"] }
+pyo3 = { version = "0.16.2" }
 numpy = "0.16.2"
 ndarray = "0.13"
 onig = { version = "6.0", default-features = false }
@@ -28,5 +28,6 @@ path = "../../tokenizers"
 tempfile = "3.1"
 
 [features]
+default = ["pyo3/extension-module"]
 test = ["pyo3/auto-initialize"]
 
--- a/bindings/python/Makefile
+++ b/bindings/python/Makefile
@@ -20,7 +20,7 @@ TESTS_RESOURCES = $(DATA_DIR)/small.txt $(DATA_DIR)/roberta.json
 test: $(TESTS_RESOURCES)
 	pip install pytest requests setuptools_rust numpy pyarrow datasets
 	python -m pytest -s -v tests
-	cargo test --features test
+	cargo test --no-default-features --features test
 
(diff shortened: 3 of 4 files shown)


ydshieh:
Thanks, @Narsil for working on this issue.

First, should this line be changed too:
https://github.com/huggingface/tokenizers/blob/e592992d428e0b9d467cf4fa08b11f5e54558025/tokenizers/src/processors/roberta.rs#L179
?

In the 0.12.1, it was
https://github.com/huggingface/tokenizers/blob/8a9bb28f4667b63f94ed96a3c3672f5db226de4d/tokenizers/src/processors/roberta.rs#L180


If I understand correctly, `roberta_processing` is a recently added test in this file, right?

Narsil:
> First, should this line be changed too:

Yes indeed. 
Even slightly more than that. The encodings are passed with type_ids being set by default to their number (only pairs are really used, so [0, and 1] but it seemed that roberta used 0 everywhere (the default used to be 0

> If I understand correctly, roberta_processing is a recently added test in this file, right? Overall, LGTM for the change, thanks again. We should probably double check why we used all 0 before (when we get some time).

Yes there was no test basically, so I added the tests with the trait modification. That's why I put the values that seemed logic, not necessarily the ones that used to be there before (I tried to create the test with 0.12.1 version, I may have done it wrong for this one)

ydshieh (approved):
From the docstring in `RobertaTokenizer`
https://github.com/huggingface/transformers/blob/a04140afb12b5dcf1df023e2180aedf616f3da87/src/transformers/models/roberta/tokenization_roberta.py#L400
and its return value
https://github.com/huggingface/transformers/blob/a04140afb12b5dcf1df023e2180aedf616f3da87/src/transformers/models/roberta/tokenization_roberta.py#L415-L417
all 0s is expected. And the change in this PR is totally right! Thanks.

mishig25 (approved):

ydshieh:
What I found is `class LayoutLMTokenizer(BertTokenizer):`. But the checkpoint in `impira/layoutlm-document-qa` havs `"tokenizer_class": "RobertaTokenizer"`. So it loads `RobertaTokenizer` instead of `LayoutLMTokenizer`. 

See
https://huggingface.co/impira/layoutlm-document-qa/blob/main/config.json

This should answer your question @Narsil . But we are not sure why `impira/layoutlm-document-qa` has `"tokenizer_class": "RobertaTokenizer"` - a bug, or the author intend to use that tokenizer class for training etc.

julien-c:
> But we are not sure why `impira/layoutlm-document-qa` has `"tokenizer_class": "RobertaTokenizer"` - a bug, or the author intend to use that tokenizer class for training etc.

You should ask directly in the [repo discussions](https://huggingface.co/impira/layoutlm-document-qa), no @ydshieh?

PR #1079, opened 2022-10-05
Updating python formatting.

Narsil:
Linked to https://github.com/huggingface/safetensors/pull/17

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -108,7 +108,7 @@ jobs:
         working-directory: ./bindings/python
         run: |
           source .env/bin/activate
-          pip install black==20.8b1 click==8.0.4
+          pip install black==22.3 click==8.0.4
           make check-style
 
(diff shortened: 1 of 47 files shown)


.github/workflows/python.yml
```diff
@@ -108,7 +108,7 @@ jobs:
         working-directory: ./bindings/python
         run: |
           source .env/bin/activate
-          pip install black==20.8b1 click==8.0.4
+          pip install black==22.3 click==8.0.4
```
mishig25, line 111: would it make sense to make them [setup tools extras](https://github.com/huggingface/doc-builder/blob/main/setup.py#L12) & call as `pip install tokenizers[dev]` ?

Narsil: Hmm it does, just moves the location of this information. Will have to check about `click` too.

mishig25 (approved):
lgtm!

PR #1095, opened 2022-11-03
Add spaces_between_special_tokens and cleanup tokenization spaces

ArthurZucker:
Following [#19293](https://github.com/huggingface/transformers/issues/19293) in transformers, we are going to add the `spaces_between_special_tokens` argument to the `decode` function. 

cc @ankrgyl

### DIFF
--- a/bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
+++ b/bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
@@ -41,7 +41,7 @@ class BaseTokenizer:
             The vocabulary
         """
         return self._tokenizer.get_vocab(with_added_tokens=with_added_tokens)
-    
+
     def get_added_tokens(self):
         """
         Get the underlying added tokens.
--- a/bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
+++ b/bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
@@ -42,15 +42,6 @@ class BaseTokenizer:
         """
         return self._tokenizer.get_vocab(with_added_tokens=with_added_tokens)
 
-    def get_added_tokens(self):
-        """
-        Get the underlying added tokens.
-
-        Returns:
-            :obj:`List[str]`: The string representation of the added tokens
-        """
-        return self._tokenizer.get_added_tokens()
-
     def get_vocab_size(self, with_added_tokens: bool = True) -> int:
         """Return the size of vocabulary, with or without added tokens.
 
--- a/bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
+++ b/bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
@@ -264,7 +264,6 @@ class BaseTokenizer:
         self,
         ids: List[int],
         skip_special_tokens: Optional[bool] = True,
-        clean_up_tokenization_spaces: Optional[bool] = False,
         spaces_between_added_tokens: Optional[bool] = True,
     ) -> str:
         """Decode the given list of ids to a string sequence
@@ -276,8 +275,6 @@ class BaseTokenizer:
             skip_special_tokens: (`optional`) boolean:
                 Whether to remove all the special tokens from the output string
 
-            clean_up_tokenization_spaces: (`optional`) boolean:
-                Whether or not to clean up the tokenization spaces.
 
             spaces_between_added_tokens: (`optional`) boolean:
                 Whether an extra space (" ") should be prepended to special tokens
@@ -291,7 +288,6 @@ class BaseTokenizer:
         return self._tokenizer.decode(
             ids,
             skip_special_tokens=skip_special_tokens,
-            clean_up_tokenization_spaces=clean_up_tokenization_spaces,
             spaces_between_added_tokens=spaces_between_added_tokens,
         )
 
@@ -299,7 +295,6 @@ class BaseTokenizer:
         self,
         sequences: List[List[int]],
         skip_special_tokens: Optional[bool] = True,
-        clean_up_tokenization_spaces: Optional[bool] = False,
         spaces_between_added_tokens: Optional[bool] = True,
     ) -> str:
         """Decode the list of sequences to a list of string sequences
@@ -310,9 +305,6 @@ class BaseTokenizer:
 
             skip_special_tokens: (`optional`) boolean:
                 Whether to remove all the special tokens from the output strings
-
-            clean_up_tokenization_spaces: (`optional`) boolean:
-                Whether or not to clean up the tokenization spaces.
 
             spaces_between_added_tokens: (`optional`) boolean:
                 Whether an extra space (" ") should be prepended to special tokens
@@ -326,7 +318,6 @@ class BaseTokenizer:
         return self._tokenizer.decode_batch(
             sequences,
             skip_special_tokens=skip_special_tokens,
-            clean_up_tokenization_spaces=clean_up_tokenization_spaces,
             spaces_between_added_tokens=spaces_between_added_tokens,
         )
 
--- a/bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
+++ b/bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
@@ -265,7 +265,7 @@ class BaseTokenizer:
         ids: List[int],
         skip_special_tokens: Optional[bool] = True,
         clean_up_tokenization_spaces: Optional[bool] = False,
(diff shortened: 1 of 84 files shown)


bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
```diff
@@ -279,13 +294,20 @@ def decode_batch(self, sequences: List[List[int]], skip_special_tokens: Optional
             skip_special_tokens: (`optional`) boolean:
                 Whether to remove all the special tokens from the output strings
 
+            spaces_between_special_tokens: (`optional`) boolean:
+                Whether an extratr space (" ") should be prepended to special tokens
```
Narsil, line 298:
```suggestion
                Whether an extra space (" ") should be prepended to special tokens
```

tokenizers/src/tokenizer/mod.rs
```diff
@@ -795,14 +795,21 @@ where
...
         let tokens = ids
             .into_iter()
             .filter_map(|id| {
                 self.added_vocabulary
                     .id_to_token(id, &self.model)
                     .filter(|token| {
-                        !skip_special_tokens || !self.added_vocabulary.is_special_token(token)
+                        !skip_special_tokens
+                            || !self.added_vocabulary.is_special_token(token)
+                            || !spaces_between_special_tokens
```
Narsil, line 812: Huh ?

This seems a bit odd.  spaces_between_special_tokens should control how to reconstruct the string no ?

Narsil (commented):
Overall looks good.

I'm a bit surprised by the core fix.
I think adding another test with `spaces_between_special_tokens=False` should help confirm behavior.
We probably need to test against a few different decoders too. (ByteLevel and Whitespace at least).

tokenizers/src/tokenizer/mod.rs
ArthurZucker: Yep, I think I am missing this part

RomanCast:
Hello, any update on this PR? It would be really helpful if integrated into tokenizers! 

Thanks for the work!

ArthurZucker:
Hey! Should be updated this week 😉

lucasjinreal:
Please update this 1 years later.

https://github.com/huggingface/tokenizers/issues/1258

this issue indicates some REAL problems might becaused of this

tokenizers/src/tokenizer/mod.rs
```diff
@@ -795,22 +816,59 @@ where
...
+        let sub_texts_concatenated = if spaces_between_special_tokens {
+            sub_texts.join(" ")
+        } else {
+            sub_texts.join("")
+        };
+
+        let tokens = sub_texts_concatenated
+            .split(' ')
+            .map(|s| s.to_string())
+            .collect();
```
ArthurZucker, line 862: Use stateless instead of this

tokenizers/src/tokenizer/mod.rs
```diff
@@ -643,6 +643,11 @@ where
         final_vocab
     }
 
+    /// Get the added tokens
+    pub fn get_added_tokens(&self) -> Vec<String> {
+        self.added_vocabulary.get_added_tokens()
+    }
```
Narsil, line 649: Let's remove this. Vec allocs is costly, you don't want to do this.

tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -192,6 +192,14 @@ impl AddedVocabulary {
         &self.added_tokens_map
     }
 
+    /// Get only the added tokens
+    pub fn get_added_tokens(&self) -> Vec<String> {
+        self.added_tokens
+            .iter()
+            .map(|token| token.content.clone())
+            .collect()
+    }
```
Narsil, line 201: Let's remove this. This is creating an entire new vec, which isn't necessary.

tokenizers/tests/documentation.rs
```diff
@@ -451,10 +454,31 @@ fn pipeline_bert() -> tokenizers::Result<()> {
...
+#[test]
+fn spaces_between_added_tokens() -> tokenizers::Result<()> {
+    let mut bert_tokenizer = Tokenizer::from_file("data/bert-wiki.json")?;
+    bert_tokenizer.add_special_tokens(&[
+        AddedToken::from("[ABC]", false),
+        AddedToken::from("[DEF]", false),
+        AddedToken::from("GHI IHG", false),
+    ]);
+    let input_ids = bert_tokenizer.encode("[ABC][DEF][ABC]GHI IHG[DEF]", false)?;
+    let decoded_wo_spaces = bert_tokenizer.decode(input_ids.get_ids(), true, false)?;
```
Narsil, line 474: Please assert the ids so we can see what we're talking about.

tokenizers/tests/documentation.rs
```diff
@@ -451,10 +454,31 @@ fn pipeline_bert() -> tokenizers::Result<()> {
...
+    println!("{}", decoded_wo_spaces);
+    assert_eq!(decoded_wo_spaces, "[ABC][DEF][ABC]GHI IHG[DEF]");
+
+    let decoded_w_spaces = bert_tokenizer.decode(input_ids.get_ids(), true, true)?;
+    println!("{}", decoded_w_spaces);
+    // "[ABC] [DEF] [ABC] GHI IHG [DEF]"
+    assert_ne!(decoded_wo_spaces, decoded_w_spaces);
+    assert_eq!(decoded_w_spaces, "[ABC] [DEF] [ABC] GHI IHG [DEF]");
```
Narsil, line 482: Do you mind adding a special in the mix so we can check that skip still works with spaces between ?

tokenizers/src/tokenizer/mod.rs
```diff
@@ -795,22 +800,37 @@ where
...
+                    continue;
+                } else if (self.added_vocabulary.get_added_tokens().contains(&token)
+                    && spaces_between_added_tokens
+                    || push_spaces)
+                    && index < last_idx - 1
+                {
+                    tokens_to_decode.extend([token, " ".to_string()]);
+                } else {
+                    tokens_to_decode.push(token);
+                }
```
Narsil, line 827: I'm not sure this works.

This is already tons better than before, but you're still modifying "tokens_to_decode" by introducing extra spaces.

The `push_spaces` is really convoluted (it's defined as double negative) and the you use `||` in the condition, meaning we always add if there's no decoder, but we add only if it's an added token when there's a decoder.

Refactoring the conditions names might make this a bit more obvious what we're trying to achieve.

Also adding extra "tokens" containing `" "` will *definitely* have unintended consequences in the decoders....
I think this kind of invalidates the approach.

Another idea would be to "add a decoder" at the end of the chain that would deal with those added tokens or somethign like that. In any case, "faking" tokens will yield issues I think.

tokenizers/tests/documentation.rs
ArthurZucker: Sure!

tokenizers/src/tokenizer/added_vocabulary.rs
ArthurZucker: Removed it, totally agree

tokenizers/src/tokenizer/mod.rs
ArthurZucker: yep

tokenizers/src/tokenizer/mod.rs
ArthurZucker: Refactored it!

tokenizers/src/tokenizer/mod.rs
```diff
@@ -795,22 +795,37 @@ where
...
+                if self.added_vocabulary.is_special_token(&token) && skip_special_tokens {
+                    continue;
+                }
+                if self.added_vocabulary.get_vocab().contains_key(&token) {
+                    let idx = tokens_to_decode.len();
+                    if spaces_between_added_tokens && !join_on_spaces {
+                        token += " "
+                    }
+                    if join_on_spaces && idx >= 1 && !spaces_between_added_tokens {
+                        tokens_to_decode[idx - 1] += &token;
```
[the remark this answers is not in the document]
ArthurZucker: The problem is that at the end, there's going to be a `tokens_to_decode.join(" ")`, thus if I don't concatenate to the previous element, spaces will be added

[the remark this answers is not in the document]
ArthurZucker: So I'll get rid of the join

ArthurZucker:
The PR is ready, but after internal discussion it was decided that it will not be merged. 
The problem is that `decoders` don't match well with this, and the argument is rather not intuitive.

PR #1115, opened 2022-11-30
Include license file in Rust crate

ankane:
Hey, thanks for this great library!

This PR symlinks the license file so it's included in the published crate.


ankane:
Looks like the CI failure is due to a security advisory (unrelated).

ankane:
Thanks @Narsil!

PR #1130, opened 2022-12-23
Wrap rustdoc html entity in code block

hvaara:
This prevents a warning when running the linter about opened code block not having a closing block.


Narsil (approved):
LGTM.
Thank you for this

PR #1144, opened 2023-01-13
Allow to build without onig or fancy-regex

llogiq:
Not all tokenizers need the advanced features of either oniguruma or fancy-regex. Unfortunately, as things are now, tokenizers does not build unless either the "onig" or the "unstable-wasm" feature is enabled.

So this PR allows tokenizers to be built with plain regex, reducing both build time and size overhead.

### DIFF
--- a/tokenizers/src/normalizers/replace.rs
+++ b/tokenizers/src/normalizers/replace.rs
@@ -1,5 +1,8 @@
 use crate::tokenizer::{NormalizedString, Normalizer, Result};
+#[cfg(any(feature = "onig", feature = "unstable-wasm"))]
 use crate::utils::SysRegex;
+#[cfg(not(any(feature = "onig", feature = "unstable-wasm")))]
+use regex::Regex as SysRegex;
 use serde::{Deserialize, Serialize};
 
 /// Represents the different patterns that `Replace` can use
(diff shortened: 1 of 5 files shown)


llogiq:
Oops, the build failed for `--all-features`. Hope that's fixed now.

llogiq:
The error on the python-macosj build seems to be an unrelated network timeout. Can someone restart the job, please?

Narsil:
Hi @llogiq,

I sympathize with the idea, however I don't think this can work.

The biggest issue is that Regex ARE being used in many places, and that as far as I know only `onig` works exactly as it should for the existing tokenizers in the wild (the ones that have been written even before this lib).
It's been confirmed recently by @n1t0 (the original creator of this lib) that he tried several engines before settling on `onig`. 

Introducing another regex, is most likely going to bite users because the tokenization is going to start being different than what is supposed to be there. 
Even `fancy-regex` which is supposed to be the same as `onig` probably isn't (I have yet to find two different regex engines which behave exactly the same).  which is why the feature glag is called `unstable-wasm` at least you are warned with the name that this is `unstable` and not necessarily 100% correct. (But if you know your tokenizer doesn't use a regex or the regex works correctly then it's all fine).

So my take is that we probably shouldn't use `regex` at all. 

Now we can keep your PR open, see if others actually hardly want to keep the low building times, maybe there are other solutions, like ditching entirely Regex support when no feature is enabled (so loading tokenizers would fail if they actually included regex for instance).

Would that be ok for you ?

llogiq:
That's completely OK with me.

PR #1166, opened 2023-02-28
Update model.rs

thomasw21:
Remove the notion of "Trained by HF" since models can now be converted from slow tokenizers, which makes this denomination slightly misleading.


Narsil (commented):
LGTM

Narsil:
Thanks.

PR #1173, opened 2023-03-07
pyo3 v0.18 migration

mert-kurttutan:
Essentially, it involves changing pyo3 signature.
I think some of the the custom written `__text_signature__` becomes unnecessary after v0.18 since they are generated based on `signature`

Relevant link: https://pyo3.rs/v0.18.0/migration.html#from-017-to-018

### DIFF
--- a/bindings/python/src/encoding.rs
+++ b/bindings/python/src/encoding.rs
@@ -78,7 +78,7 @@ impl PyEncoding {
     /// Returns:
     ///     :class:`~tokenizers.Encoding`: The resulting Encoding
     #[staticmethod]
-    #[args(growing_offsets = true)]
+    #[pyo3(signature = (encodings, growing_offsets = true))]
     #[pyo3(text_signature = "(encodings, growing_offsets=True)")]
     fn merge(encodings: Vec<PyRef<PyEncoding>>, growing_offsets: bool) -> PyEncoding {
         tk::tokenizer::Encoding::merge(
@@ -263,7 +263,7 @@ impl PyEncoding {
     ///
     /// Returns:
     ///     :obj:`Tuple[int, int]`: The range of tokens: :obj:`(first, last + 1)`
-    #[args(sequence_index = 0)]
+    #[pyo3(signature = (word_index, sequence_index = 0))]
     #[pyo3(text_signature = "(self, word_index, sequence_index=0)")]
     fn word_to_tokens(&self, word_index: u32, sequence_index: usize) -> Option<(usize, usize)> {
         self.encoding.word_to_tokens(word_index, sequence_index)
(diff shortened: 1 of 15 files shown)


mert-kurttutan:
I also have a maturin version of the codebase working and passing some of the builds tests. It seems to be working for alot of other projects, e.g. polars with minimal configuration (e.g. no need to repair wheels for manylinux standard) and github actions.
In case this fits your future plan, I can provide a PR on this one.

Narsil:
>  In case this fits your future plan, I can provide a PR on this one.

If you already have it, that would be nice for sure ! Right now, I'm not really into investing time to debug our build pipelines, as it's tedious and error prone (manylinux is one of them, but there's m1 and conda which are also pain in the neck). 
So I'm not sure about merging, but it would be nice to unify at some point with maturin !

bindings/python/src/encoding.rs
```diff
@@ -263,7 +263,7 @@ impl PyEncoding {
     ///
     /// Returns:
     ///     :obj:`Tuple[int, int]`: The range of tokens: :obj:`(first, last + 1)`
-    #[args(sequence_index = 0)]
+    #[pyo3(signature = (word_index, sequence_index = 0))]
     #[pyo3(text_signature = "(self, word_index, sequence_index=0)")]
```
Narsil, line 267: Does the inspected signature automatically include `self` here too ? If yes then yes let's remove those `text_signature`.

There were other things to check for, but as long as the `.pyi` are consistent I'll agree to remove them (especially if they are duplicate, and the pyo3 signature is actually validated)

Narsil:
I'll merge this, don' thesitate to open another PR to remove `text_signature` (it requires some slight modification to `stub.py` because PyO3 uses `$self` instead of `self` for the text_signature.

mert-kurttutan:
> I'll merge this, don' thesitate to open another PR to remove `text_signature` (it requires some slight modification to `stub.py` because PyO3 uses `$self` instead of `self` for the text_signature.

Exactly, this actually caused a problem with formatting by `black`

bindings/python/src/encoding.rs
mert-kurttutan: You already answered this :)

Narsil:
You can just do a small `replace("$self", "self")` in the signature so that black is happy.
However, I couldn't find the equivalent signature for the class itself (to create the `__init__` in the pyi without text_signature.

I would have guessed it was `__init__` that the `#[new]` function targets, but it doesn't seem so, maybe the `__new__` (I think for the pyi we could lie and pretend it was the __init__ to avoid confusion since `__new__` is very rarely used in Python world
)

PR #1182, opened 2023-03-15
Fixing infinite loop in UnigramTrainer.

### LINKED ISSUE #1179: UnigramTrainer Stalls on Small Vocabulary Size
I'm training a Unigram model using the UnigramTrainer and inexplicably, it seems to not terminate sometimes when given a large enough coprus, but not for other slightly smaller corpus sizes.

I'm training tokenizers with different vocabulary sizes to compare models from a vocab size of 100 tokens to thousands of tokens and I suspect that it has to do with a while condition not being met somewhere for small vocabulary sizes. Perhaps the initial alphabet is larger than the vocabulary size specified and this causes the problem?

I had a similar problem when I implemented my own tokenzier. If you don't check that the initial alphabet is the same size as the pruned vocabulary size, you can keep calculating loss and trying to prune non-alphabet tokens, but there are none and so you can never reach your target vocabulary size.

Narsil:
Fix #1179.

### DIFF
--- a/tokenizers/src/models/unigram/trainer.rs
+++ b/tokenizers/src/models/unigram/trainer.rs
@@ -29,6 +29,12 @@ fn digamma(mut x: f64) -> f64 {
     result
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum UnigramTrainerError {
+    #[error("The vocabulary is not large enough to contain all chars")]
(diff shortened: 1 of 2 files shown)


sgugger (approved):
LGTM, thanks for the fix!

PR #1188, opened 2023-03-22
Adding `split` argument to Metaspace to be able to disable it

Narsil:
For Llama Tokenizer.


ArthurZucker:
As disscused offline, we are probably going to drop this in favor of a `SpmBpeMetaSpace` to restrict usage of this particular split

Narsil:
> As disscused offline, we are probably going to drop this in favor of a SpmBpeMetaSpace to restrict usage of this particular split

We misunderstood each other, I think this is better than having a separate metaspace, which has nothing to do with Bpe (Metaspace is already for SPM).

PR #1192, opened 2023-03-23
Faster `datasets` train example

lhoestq:
Using .iter() is much faster than accessing using row ids


Narsil (approved):
I don't see any performance improvement on my computer with this.

Everything is the same.

I am guessing this speedup doesn't impact the probable actual bottleneck.

Let's merge since code is just better ! :)

PR #1196, opened 2023-03-23
Adding 2 new decoders:

Narsil:
- Fuse will simply concatenate all tokens into 1 string
- Strip will remove n char from left or right

Sequence(Replace("_", " "), Fuse(), Strip(1, 0)) should be what we want
for the `Metaspace` thing.

- Note: Added a new dependency from better parsing of decoders.
This is due to untagged enums which can match anything the `MustBe`
ensure there's no issue between Fuse and ByteFallback.
Since both are new the chances for backward incompatibility is low.


bindings/python/tests/bindings/test_decoders.py
```diff
@@ -72,6 +83,30 @@ def test_decoding(self):
...
+        decoder = Fuse()
+        assert decoder.decode(["My", " na", "me"]) == "My name"
+
+
+class TestStrip:
+    def test_instantiate(self):
+        assert Strip(left=0, right=0) is not None
+        assert isinstance(Strip(left=0, right=0), Decoder)
+        assert isinstance(Strip(left=0, right=0), Strip)
+        # assert isinstance(pickle.loads(pickle.dumps(Strip(left=0, right=0))), Strip)
```
sgugger, line 103: Same here, to remove or fix?

sgugger (approved):
Thanks for adding this!

bindings/python/tests/bindings/test_decoders.py
Narsil: Hoped it would go unoticed :)
 
 Pickling is a nice touch, but cumbersome.

PR #1217, opened 2023-04-12
Add unigram bytefallback

ArthurZucker:
Adds support for bytfallback with the unigram model

### DIFF
--- a/bindings/python/tests/bindings/test_tokenizer.py
+++ b/bindings/python/tests/bindings/test_tokenizer.py
@@ -414,29 +414,26 @@ class TestTokenizer:
         assert output.tokens == ["Hey", "Ġthere", "Ġdear", "Ġfriend", "!"]
 
     def test_unigram_byte_fallback(self):
-        vocab =[
+        vocab = [
             ("<unk>", 0.0),
             ("A", -0.01),
             ("sen", -0.02),
             ("te", -0.03),
             ("n", -0.04),
             ("ce", -0.05),
-            ("<0x09>", -0.06),
-            
             ("<0xF0>", -0.06),
             ("<0x9F>", -0.06),
             ("<0xA4>", -0.06),
             ("<0x97>", -0.06),
-
             (" ", -0.4),
         ]
-        tokenizer = tokenizer = Tokenizer(Unigram(vocab, 0,byte_fallback = False))
+        tokenizer = tokenizer = Tokenizer(Unigram(vocab, 0, byte_fallback=False))
 
         output = tokenizer.encode("A sentence 🤗")
         assert output.ids[-1] == 0
-        assert output.tokens == ['A', ' ', 'sen', 'te', 'n', 'ce', ' ', '🤗']
-        
-        tokenizer = Tokenizer(Unigram(vocab, 0, byte_fallback = True))
+        assert output.tokens == ["A", " ", "sen", "te", "n", "ce", " ", "🤗"]
+
+        tokenizer = Tokenizer(Unigram(vocab, 0, byte_fallback=True))
 
         output = tokenizer.encode("A sentence 🤗")
-        assert output.tokens == ['A', ' ', 'sen', 'te', 'n', 'ce', ' ', '<0xF0>', '<0x9F>', '<0xA4>', '<0x97>']
+        assert output.tokens == ["A", " ", "sen", "te", "n", "ce", " ", "<0xF0>", "<0x9F>", "<0xA4>", "<0x97>"]
--- a/bindings/python/tests/bindings/test_tokenizer.py
+++ b/bindings/python/tests/bindings/test_tokenizer.py
@@ -430,10 +430,11 @@ class TestTokenizer:
         tokenizer = tokenizer = Tokenizer(Unigram(vocab, 0, byte_fallback=False))
 
         output = tokenizer.encode("A sentence 🤗")
-        assert output.ids[-1] == 0
+        assert output.ids == [1, 10, 2, 3, 4, 5, 10, 0]
         assert output.tokens == ["A", " ", "sen", "te", "n", "ce", " ", "🤗"]
 
         tokenizer = Tokenizer(Unigram(vocab, 0, byte_fallback=True))
 
         output = tokenizer.encode("A sentence 🤗")
+        assert output.ids == [1, 10, 2, 3, 4, 5, 10, 6, 7, 8, 9]
         assert output.tokens == ["A", " ", "sen", "te", "n", "ce", " ", "<0xF0>", "<0x9F>", "<0xA4>", "<0x97>"]
--- a/bindings/python/tests/bindings/test_tokenizer.py
+++ b/bindings/python/tests/bindings/test_tokenizer.py
@@ -5,7 +5,7 @@ import pytest
 
 from tokenizers import AddedToken, Encoding, Tokenizer
 from tokenizers.implementations import BertWordPieceTokenizer
-from tokenizers.models import BPE, Model, WordPiece
+from tokenizers.models import BPE, Model, WordPiece, Unigram
 from tokenizers.normalizers import Lowercase
 from tokenizers.pre_tokenizers import ByteLevel
 from tokenizers.processors import BertProcessing, RobertaProcessing
@@ -412,3 +412,31 @@ class TestTokenizer:
         tokenizer = Tokenizer.from_pretrained("anthony/tokenizers-test", revision="gpt-2")
         output = tokenizer.encode("Hey there dear friend!", add_special_tokens=False)
         assert output.tokens == ["Hey", "Ġthere", "Ġdear", "Ġfriend", "!"]
+
+    def test_unigram_byte_fallback(self):
+        vocab =[
+            ("<unk>", 0.0),
+            ("A", -0.01),
+            ("sen", -0.02),
+            ("te", -0.03),
+            ("n", -0.04),
+            ("ce", -0.05),
+            ("<0x09>", -0.06),
+            
+            ("<0xF0>", -0.06),
+            ("<0x9F>", -0.06),
+            ("<0xA4>", -0.06),
+            ("<0x97>", -0.06),
+
+            (" ", -0.4),
+        ]
+        tokenizer = tokenizer = Tokenizer(Unigram(vocab, 0,byte_fallback = False))
+
+        output = tokenizer.encode("A sentence 🤗")
+        assert output.ids[-1] == 0
+        assert output.tokens == ['A', ' ', 'sen', 'te', 'n', 'ce', ' ', '🤗']
+        
+        tokenizer = Tokenizer(Unigram(vocab, 0, byte_fallback = True))
+
+        output = tokenizer.encode("A sentence 🤗")
+        assert output.tokens == ['A', ' ', 'sen', 'te', 'n', 'ce', ' ', '<0xF0>', '<0x9F>', '<0xA4>', '<0x97>']
--- a/tokenizers/src/models/unigram/model.rs
+++ b/tokenizers/src/models/unigram/model.rs
@@ -91,15 +91,6 @@ fn byte_to_piece(c: u8) -> String {
     format!("<0x{:02X}>", c)
 }
 
-fn piece_to_byte(piece: &str) -> Option<u8> {
-    let mut k_map: HashMap<String, u8> = HashMap::new();
-    for i in 0..=255 {
-        let byte_piece = byte_to_piece(i);
-        k_map.insert(byte_piece.clone(), i);
-    }
-    k_map.get(piece).copied()
-}
-
 impl Unigram {
     /// Create a `Unigram` model from a given vocabulary.
     /// Vocabulary are the various tokens and their associated score which is a sort of a logprob of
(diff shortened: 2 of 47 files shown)


tokenizers/src/models/unigram/model.rs
```diff
@@ -401,16 +406,33 @@ impl Model for Unigram {
...
+                            let code = format!("<{:#04X}>", byte);
+                            if let Some(id) = self.token_to_ids.get(&code) {
+                                ids.push(*id);
+                            }
+                        }
+                        if ids.is_empty() {
+                            self.unk_id.ok_or(UnigramError::MissingUnkId)? as u32
+                        } else {
+                            *ids.first().unwrap()
+                        }
```
ArthurZucker, line 429: probably wrong, all tokens have to be pushed + offsets

chris-ha458: atleast in sentencepiece, the bytefallbacks do not use the push/offset trick and just use reserved
<0x00> to <0xFF>

chris-ha458:
Something like this could initialize initial vocabulary for byte_fallback.
This could aslo be useful for 
https://github.com/huggingface/tokenizers/pull/1183#issuecomment-1518910416

```
const UNICODE_CAPACITY: usize = 256;

pub fn create_encoded_bytes() -> Vec<String> {
    let mut bytes: Vec<String> = Vec::with_capacity(UNICODE_CAPACITY);

    for i in 0..UNICODE_CAPACITY {
        let encoded: String = format!("<0x{:02X}>", i);
        bytes.push(encoded);
    }
    bytes
}
```

ArthurZucker:
Hey! Sorry I just got back from holidays! Will be updating soon

chris-ha458:
No apology necessary! Hope you had a good vacation.

I am interested in how you plan to address the alphabet/initial tokens situation
(like my code snippet)

It would have to be ensured to not be tokenized at anytime (they are not meant to be tokenized, rather processed internally as fallback during tokenization and generation)

AFAICT there are no mechanisms like spm's control symbols that are ensured not to have a surface representation.

One way would be to assign those tokens a i32::MIN log prob so as to make it very unlikely to be tokenized.

tokenizers/src/models/unigram/model.rs
```diff
@@ -407,14 +433,22 @@ impl Model for Unigram {
...
+            } else {
+                vec![self.unk_id.ok_or(UnigramError::MissingUnkId)? as u32]
             };
-            let len = string.len();
-            let offsets = (offset, offset + len);
+            let len = string.len() - ids.len() + 1;
+            for id in ids {
+                let offsets = (offset, offset + len);
+                tokens.push(Token::new(id, self.id_to_token(id).unwrap(), offsets));
+            }
```
Narsil, line 450: Remove every `unwrap` and every `vec`.

There is 1 `collect` tolerated ( I think it's done that way in BPE) and it's *only* to check that ALL bytes have a token id (you're allowed to use a single `vec` or `collect` in that branch, not in the others.

:)

chris-ha458: I understand the `unwrap` part since it implicitly implies that potential errors are not being handled.

Can you share the reasoning regarding `vec`? is it because since this focuses on the bytefallback pieces, which should be known during compile time (256 of them) and therefore should be addressed with arrays?

ArthurZucker:
Now removing things related to training

bindings/python/tests/bindings/test_tokenizer.py
```diff
@@ -412,3 +412,28 @@ def test_from_pretrained_revision(self):
...
+            ("<0xF0>", -0.06),
+            ("<0x9F>", -0.06),
+            ("<0xA4>", -0.06),
+            ("<0x97>", -0.06),
+            (" ", -0.4),
+        ]
+        tokenizer = tokenizer = Tokenizer(Unigram(vocab, 0, byte_fallback=False))
+
+        output = tokenizer.encode("A sentence 🤗")
+        assert output.ids[-1] == 0
```
Narsil, line 433: Assert all the ids plz.

bindings/python/tests/bindings/test_tokenizer.py
```diff
@@ -412,3 +412,28 @@ def test_from_pretrained_revision(self):
...
+        assert output.tokens == ["A", " ", "sen", "te", "n", "ce", " ", "🤗"]
```
Narsil, line 434: Let's remove the tokens, please, only the ids matter.

tokenizers/src/models/unigram/model.rs
```diff
@@ -78,18 +81,35 @@ pub enum UnigramError {
...
+
+pub fn piece_to_byte(piece: &str) -> Option<u8> {
+    let mut k_map: HashMap<String, u8> = HashMap::new();
+    for i in 0..=255 {
+        let byte_piece = byte_to_piece(i);
+        k_map.insert(byte_piece.clone(), i);
+    }
+    k_map.get(piece).copied()
+}
+
```
Narsil, line 100: This is not used.

ArthurZucker: Indeed, will add it in another PR for training

tokenizers/src/models/unigram/model.rs
```diff
@@ -409,7 +431,22 @@ impl Model for Unigram {
         for string in str_tokens {
             let id: u32 = match self.token_to_ids.get(&string) {
                 Some(id) => *id,
-                None => self.unk_id.ok_or(UnigramError::MissingUnkId)? as u32,
+                None => {
+                    if self.byte_fallback {
+                        for byte in string.bytes() {
+                            let byte_string = byte_to_piece(byte);
+                            tokens.push(Token::new(
+                                self.token_to_ids[&byte_string],
```
Narsil, line 439: This will panic if `byte_string` is not  in `token_to_ids`, no ?

tokenizers/src/models/unigram/model.rs
```diff
@@ -78,18 +81,35 @@ pub enum UnigramError {
...
     /// Vocabulary are the various tokens and their associated score which is a sort of a logprob of
     /// their frequency, which will enable tokenization and sampling.
     /// unk_id, is the index within the vocabulary.
     /// For now `Unigram` *requires* at least `unk` because we might find a never seen char.
     /// Further versions might allow that part to be hidden.
-    pub fn from(vocab: Vec<(String, f64)>, unk_id: Option<usize>) -> Result<Self> {
+    pub fn from(
+        vocab: Vec<(String, f64)>,
+        unk_id: Option<usize>,
+        byte_fallback: Option<bool>,
```
Narsil, line 111: Why `Option<bool>` ? It should be `bool` no ?

There are no optional arguments in Rust (and it's good)

`unk_id` is really an Option, it's not at all forced (but it will cause errors if you haven't one and are triggering an unk).

ArthurZucker: I am too pythonic and tried to have a default to False and make it an option

tokenizers/src/models/unigram/model.rs
```diff
@@ -78,18 +81,35 @@ pub enum UnigramError {
...
     fn default() -> Self {
         let vocab = vec![("<unk>".to_string(), 0.0)];
-        Self::from(vocab, Some(0)).unwrap()
+        Self::from(vocab, Some(0), Some(false)).unwrap()
     }
 }
 
+pub fn byte_to_piece(c: u8) -> String {
+    format!("<0x{:02X}>", c)
+}
```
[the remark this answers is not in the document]
chris-ha458: My opinion is that leaving the function in but not `pub` is a good compromise
The code is difficult to read and understand as it is (not the developers fault. It's just doing a lot of difficult things)
Leaving this in would make it just a bit more understandable.
Inlining it in a way that's well commented/documented would also be a good alternative.

Either way, as long as it is not pub i think the performance part would be comparable and optimized away by the rust compiler

tokenizers/src/models/unigram/model.rs
chris-ha458: @ArthurZucker Tbh I still think pythonically too and I'm wondering what is the idiomatic Rust thing to do here

ArthurZucker: I think that it doesn't make sense to have to init a unigram with `Some(false)` which you have to do rigght now !

tokenizers/src/models/unigram/model.rs
```diff
@@ -27,6 +27,7 @@ pub struct Unigram {
 
     fuse_unk: bool,
     is_optimized: bool,
+    pub byte_fallback: bool,
```
Narsil, line 30:
```suggestion
    byte_fallback: bool,
```

tokenizers/src/models/unigram/model.rs
```diff
@@ -73,23 +76,33 @@ pub enum UnigramError {
...
     fn default() -> Self {
         let vocab = vec![("<unk>".to_string(), 0.0)];
-        Self::from(vocab, Some(0)).unwrap()
+        Self::from(vocab, Some(0), false).unwrap()
     }
 }
 
+fn byte_to_piece(c: u8) -> String {
+    format!("<0x{:02X}>", c)
+}
```
Narsil, line 92: Please make this inline. I insist.

tokenizers/src/models/unigram/model.rs
```diff
@@ -573,4 +605,191 @@ mod tests {
...
+                Token {
+                    id: 5,
+                    value: "<0x91>".to_string(),
+                    offsets: (15, 21)
+                }
+            ]
+        );
+
+        unigram.set_fuse_unk(false);
+        let tokens = unigram.tokenize("unfused ⅑⅑").unwrap();
```
Narsil, line 700: Why do we care about fuse_unk ? It's an orthogonal feature, isn't it?

tokenizers/src/models/unigram/model.rs
```diff
@@ -573,4 +605,191 @@ mod tests {
...
+            ("<0x90>".to_string(), -0.04),
+            ("<0x91>".to_string(), -0.05),
+            ("<0x9B>".to_string(), -0.06),
+            ("Hello".to_string(), -0.1),
+            ("how".to_string(), -0.2),
+            ("are".to_string(), -0.3),
+            (" ".to_string(), -0.4),
+        ];
+        let mut unigram = Unigram::from(sentencepieces, Some(0), true).unwrap();
+        let tokens = unigram.tokenize("Hello⅐how are⅑⅑").unwrap();
```
Narsil, line 627: Can we make these examples smaller and more readable ?

We only need 1 utf-8 multibyte token `"é"` could work (we can keep your tokens it's fine).
`"Hello⅐"` should be enough to prove the point.

Then instead of using a new utf-8 that we don't know how it's encoded, just recreate the vocab with only 1 missing byte.

And the `fuse_unk` shouldn't need to be tested. It only has impact on the offsets, which is quite orthogonal with what we're testing here, no ?

ArthurZucker: No `fuse_unk` is tested to make sure that if there more than 1 byte, all of them are returned (we disccused about this offline)

ArthurZucker: Plus we want to make sure that when 1 token in the unk sequence does not have bytefallback, then the entire string is returned as unk, instead of falling back for the known one!

ArthurZucker: Reduce the examples 😉

Narsil (approved):
LGTM !

chris-ha458:
Excited for this feature!

If there are plans for any formal release anytime soon, I'll wait for it and test this function thoroughly!
(There might be odd edge cases especially regarding the surface representation of byte-fallback tokens ie what the tokenizer would do when the corpus includes text that look like "<0x03>" in the orignal text)

If a new version of "tokenizers" isnt expected to hit soon i'll just get on trying it out asap

ArthurZucker:
Mmmm maybe we'll wait until a transformers realease includes umT5 see https://github.com/huggingface/transformers/pull/24477 ! But if you figure out the bug that's breaking nodes, kudos 😅

kellymarchisio:
I noticed byte_fallback is only implemented if you import from a model trained using Google's sentencepiece library (here with .from_spm).  Are there plans to add this to SentencePieceUnigramTokenizer.train to eliminate the dependency on sentencepiece?

chris-ha458:
I hope to see that as well.
I've tried a very ugly work around involving training with the byte_fallback on and then adding the byte pieces (<0x00> to <0xFF>) individually as added tokens, but it is a very brittle solution.

ArthurZucker:
Yep we discussed about having this in a follow PR! Did not have time to do it yet!

kellymarchisio:
Ok sg!  Do you have a minimal example for coaxing the current implementation to work with byte fallback without importing from SPM file?

chris-ha458:
let me know if you need any assistance in developing or testing!

gautierdag:
Also still interested in having `byte_fallback` properly implemented in Unigram, such that unigram models can be trained/used without `unk` tokens. I haven't managed to find a workaround and even manually adding the byte pieces to the vocabulary does not work.

ArthurZucker:
On my TODO list, @chris-ha458 I'd be happy to review a PR if you want to tackle this!

PR #1222, opened 2023-04-19
Printing warning to stderr.

### LINKED ISSUE #964: Print warnings to stderr, not stdout
I have a python script which reads some data from stdin and then outputs to stdout. However, my output is getting garbled (and therefore following steps in my pipeline break), because this library is outputting warnings to stdout as well. Specifically, I was getting [this warning](https://github.com/huggingface/transformers/issues/5486). I managed to solve this problem by setting TOKENIZERS_PARALLELISM as suggested, but I think in general that this problem should be resolved another way:

Can we adjust this library to print warnings (and maybe other diagnostic/logging info) to stderr instead of stdout? I believe that this is more in line with what stdout and stderr are intended for. Not sure what consequences this would have, I haven't read this code in detail and don't know how users use it.

I could submit a PR. Thanks for the consideration and your work!

Narsil:
Fixes #964

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -29,12 +29,12 @@ static mut REGISTERED_FORK_CALLBACK: bool = false;
 extern "C" fn child_after_fork() {
     use tk::parallelism::*;
     if has_parallelism_been_used() && !is_parallelism_configured() {
-        println!(
+        eprintln!(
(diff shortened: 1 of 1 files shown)


mishig25 (approved):
lgtm !

McPatate (approved):

PR #1227, opened 2023-04-27
implement a simple max_sentencepiece_length into BPE

chris-ha458:
Add a way for the BPE trainer to behave like the unigram trainer where tokens longer than a certain lenght(default 16 in SPM) to be skipped. this is implemented in unigram trainer but in a different way.

If this code were to be actually integrated some works to be done

1. Documentation describing the behavior and how it should be set. 
2. Set default==0 so it doesnt act unless set
3. provide ways in the python binding for the user to set max token length

### DIFF
--- a/tokenizers/src/models/bpe/trainer.rs
+++ b/tokenizers/src/models/bpe/trainer.rs
@@ -502,6 +502,18 @@ impl BpeTrainer {
                 }
             }
             let new_token = format!("{}{}", part_a, part_b);
+        
+            // implement sentencepiece-like merge.
+            // if this code were to be merged, integrate a way in the python bindings to communicate this variable
+            // default should be 0/None to maintain previous behavior. 16 is the spm default.
+            let max_merge_length = 16;
+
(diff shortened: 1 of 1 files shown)


chris-ha458:
I was trying to find a way to implement max_sentencepiece_length through pretokenizer split rules and to be honest, its very difficult and regexes can be real slow when operating on the whole training corpus.

Narsil:
Hi, thanks for the PR, why did you close it in the end ?

Narsil:
Unigram has a max length option already.

Here in BPE it's doable but I'm unsure how it's going to behave. When you don't have good pretokenization, the problem is that the range of the pairs is unbounded and so it keeps updating more and more pairs as it adds new merges.

We could have something like you propose but I would implement it at a different level. The idea would be that once a merge hits 16 length, we need to skip it, but we also want to cut all potential pairs that would generate those before hand so that the heap queue doesn't even consider those impossible merges.

This could potentially act as a range limitor (like space splitting does) to keep BPE efficient.

PR #1244, opened 2023-05-10
fix unigram.rs test_sample()

chris-ha458:
https://github.com/huggingface/tokenizers/blob/87230bb59b2a70b6c361943d1394f42218b0040b/tokenizers/tests/unigram.rs#LL71C1-L71C53

When running cargo test --release, the above line causes an error.

referring to https://github.com/huggingface/tokenizers/blob/87230bb59b2a70b6c361943d1394f42218b0040b/tokenizers/src/models/unigram/lattice.rs#L138

It seems that lattice::from should only take 3 arguments. If i had to guess, it should be Lattice::from("ABC", 0, 2); 

This change makes cargo test --release pass without error.

### DIFF
--- a/tokenizers/tests/unigram.rs
+++ b/tokenizers/tests/unigram.rs
@@ -68,7 +68,7 @@ fn test_train_unigram_from_file() {
 #[cfg(not(debug_assertions))]
 #[test]
 fn test_sample() {
-    let mut lattice = Lattice::from("ABC", 0, 1, 2);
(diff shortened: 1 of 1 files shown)


Narsil:
It works. `0`, `1` and `2` were special ids maintained by the lattice, 1 was dropped (and since this test is never run by the CI it was never seen)

Narsil (approved):
LGTM

Issue #1259, opened 2023-05-26
Path to implementing --max_sentence_length


chris-ha458:
With #1257 merged, I would like to further discuss the possibilities regarding  --max_sentence_length implementation.
To recap,--max_sentence_length is an option exposed in [spm](https://github.com/google/sentencepiece/blob/master/doc/options.md) that allows a length cap on individual input sentences.
The rationale for that feature is twofold and described in spm's protobuf (equivalent to json): 
[sentencepiece/src/sentencepiece_model.proto](https://github.com/google/sentencepiece/blob/master/src/sentencepiece_model.proto)
```
// Longer input tends to bring the following risks:
// * Overflow during EM training (unigram language model only)
// * Performance drop because of O(n log n) cost in BPE.
```
I feel that those are useful traits that the huggingface tokenizer could also have.

Implementing it would require some consideration.
Since this diverges from default behavior, we shouldn't implement in a way that exposes the option to use it in the first place.

Some ideas:
1. implement as new pretokenizer.(lengthcheck(min_length,max_length) )
2. simply extend upon #1257 and add a check for maximum length in the trainer.
3. Don't implement this function.

The first option would be the cleanest way, and we wouldn't have to worry about backward compatibility.
One would have to intentionally add it.
An issue with this approach could be that the lengthcheck isn't strictly a pretokenizer. It's more of a preprocessor.
Also it would have to be disabled during inference. I am not aware of a method to do this.

The second option is the easiest way, and i think I might be able to do most if not all of the work including bindings
(following #1228)
If we keep the default as no maximum length limit than previous approaches or people training new tokenizers will not be affected.
Also this would only influence the training process so there is no issue with train vs inference discrepancy.

One drawback for this approach could be from bringing preprocessing logic into the training logic, but since this preprocessing is unique to training i don't think it's a significant issue.
Another drawback could be that this wouldn't affect BPE as the implementation in spm does.
This isn't that much of a problem since we could easily implement a similar option there when the need arises.
. It is extremely rare from my corpuses for natural language corpuses to have longer than 4192 characters without any kind of linebreak or punctuation that would cause other tokenizers to separate it. Usually, a "sentence" longer than 4192 characters would be code. There are code in certain programming languages where the WHOLE file is just ascii and whitespace. 
No newline or linebreaks. It gets fed into the training algorithm as is. 
This has been a problem for unigram algorithm but not as much for BPE. So If it was up to me, I would just add this for unigram but not BPE.

[...]

PR #1262, opened 2023-05-29
Parallelize (pretokenization)

chris-ha458:
This commit parallelizes the process when an iterator is used to feed the tokenizer.
This would help in case of : long sequence being passed into tokenizer trainer even after pretokenization.

https://huggingface.co/docs/tokenizers/training_from_memory

This commit alone seems benign since map operates on each string slice without any concern for it's offsets.
Passes all tests.

more important are these parts within tokenizer/mod.rs but they might be much more difficult or involved.
```
let normalized = self.do_normalize(seq.as_ref())?;
let pre_tokenized = self.do_pre_tokenize(normalized)?;
```

I hope to look into low hanging fruit in the individual normalizers and pretokenizers and add them all up into this PR.

If there are suggestions for more deeper architectural changes feel free to point them out.

### DIFF
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -1131,7 +1131,7 @@ where
                 let pre_tokenized = self.do_pre_tokenize(normalized)?;
                 Ok(pre_tokenized
                     .get_splits(OffsetReferential::Original, OffsetType::Byte)
-                    .into_iter()
+                    .into_maybe_par_iter()
                     .map(|(s, _, _)| s.to_owned())
                     .collect())
             },


chris-ha458:
After much efforts into the unigram trainer parallelization, the bottleneck seems to be in the pretokenization algorithm neutral side. That's where I am focusing my efforts atm.

tokenizers/src/tokenizer/mod.rs
```diff
@@ -1131,7 +1131,7 @@ where
                 let pre_tokenized = self.do_pre_tokenize(normalized)?;
                 Ok(pre_tokenized
                     .get_splits(OffsetReferential::Original, OffsetType::Byte)
-                    .into_iter()
+                    .into_maybe_par_iter()
```
Narsil, line 1134: This is parallelizing the `get_splits` method, not the pretokenization itself. 

Shouldn't you modify `sequences.map` instead ?

chris-ha458: I think you are right.
If I read the code correctly, it would at best parallelize downstream of the pretokenization and not the pretokenization itself.
I don't think this part is the bottleneck and if it is, #1259 might provide a solution.

Since as used in this part of the code `sequences` is an already iterator(as per function defintion) and there is only map here.

I am looking upstream to where the iterator is fed into .train() and it seems to be here

https://github.com/chris-ha458/tokenizers/blob/21b8cbee3a3b02bb6cce9c129196b8fbdbff0c8e/bindings/python/src/tokenizer.rs#LL1221C42-L1221C60

```
        py.allow_threads(|| {
            ResultShunt::process(buffered_iter, |iter| {
                self.tokenizer
                    .train(&mut trainer, MaybeSizedIterator::new(iter, length))
                    .map(|_| {})
                    .map_err(|e| exceptions::PyException::new_err(e.to_string()))
            })?
        })
```
And the part that constructs that `|iter|` appeares right above.
```
let buffered_iter = PyBufferedIterator::new(
            iterator,
            |element| {
                // Each element of the iterator can either be:
                //  - An iterator, to allow batching
                //  - A string
                if let Ok(s) = element.downcast::<PyString>() {
                    itertools::Either::Right(std::iter::once(s.to_str().map(|s| s.to_owned())))
                } else {
                    match element.iter() {
                        Ok(iter) => itertools::Either::Left(
                            iter.map(|i| i?.extract::<String>())
                                .collect::<Vec<_>>()
                                **.into_iter(), # Change this into into_maybe_par_iter()?**
                        ),
                        Err(e) => itertools::Either::Right(std::iter::once(Err(e))),
                    }
                }
            },
            256,
        )?;
```
My hope is that changing the line with
`**.into_iter(), # Change this into into_maybe_par_iter()?**`
would do the trick.

As far as i can tell there isn't anyplace downstream that uses any sideeffects or shared state so
it would work, and the type system and tests would help guide me if I'm wrong.

chris-ha458: Okay this is much much more involved than I had hoped.
This is also uncomfortably close to the GIL

PR #1270, opened 2023-06-07
don't update added tokens map if token already in vocab

ArthurZucker:
# Issue : 
When tokens are added using `add_token`, if they were already part of the vocab, they are not completely ignored.:
- the `added_tokens_map_r` is update, but the `added_tokens` is not 
- this induces bugs, where changing `single_word` will result in a different behaviour after adding the token again. 
- erros do propagate to `transformers` : adding a token that already exists ( let's say `eos_token`) will not update the `trie` but it will update the `added_tokens_map_r`. 

A potential improvement here could be to add the feature `update_added_token` to update a token that was already added. 

Missing feature in fast that could help debug slow and vice-versa: 
- `.get_special_tokens()` to return the equivalent `added_special_tokens` of slow. 
- `.get_added_tokens()` to return the equivalent `added_tokens` of slow
- `.get_added_tokens_map()` to return the equivalent `added_tokens_encoder` of slow
- `.get_added_vocab` which would return the entire vocab
This fixes the issue when an added token is updated if it was already part of the vocab

### DIFF
--- a/tokenizers/src/tokenizer/added_vocabulary.rs
+++ b/tokenizers/src/tokenizer/added_vocabulary.rs
@@ -250,12 +250,10 @@ impl AddedVocabulary {
             } else {
                 let new_id = (model.get_vocab_size() + self.added_tokens_map.len()) as u32;
                 self.added_tokens_map.insert(token.content.clone(), new_id);
-
+                self.added_tokens_map_r.insert(new_id, token.clone());
                 if !self.special_tokens_set.contains(&token.content) {
                     self.added_tokens.push(token.clone());
                 }
-                // Update the current revert operation
-                self.added_tokens_map_r.insert(new_id, token.clone());
             };
         }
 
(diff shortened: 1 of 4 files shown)


tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -244,30 +244,19 @@ impl AddedVocabulary {
...
                 let new_id = (model.get_vocab_size() + self.added_tokens_map.len()) as u32;
                 self.added_tokens_map.insert(token.content.clone(), new_id);
 
                 if !self.special_tokens_set.contains(&token.content) {
                     self.added_tokens.push(token.clone());
                 }
-
-                new_id
+                // Update the current revert operation
+                self.added_tokens_map_r.insert(new_id, token.clone());
```
ArthurZucker, line 258: unwrap issue here?

PR #1289, opened 2023-07-05
revise type specification

hiroshi-matsuda-rit:
The type of the arg `vocab` is defined as `Option[str]` in `SentencePieceUnigramTokenizer.__init()__` but `List[Tuple[str, float]]` in `Unigram`

https://github.com/huggingface/tokenizers/blob/v0.13.4-rc2/bindings/python/py_src/tokenizers/implementations/sentencepiece_unigram.py#L19
https://github.com/huggingface/tokenizers/blob/v0.13.4-rc2/bindings/python/py_src/tokenizers/models/__init__.pyi#L245

This inconsistency is very confusing to users.


ArthurZucker (approved):
LGTM thanks for the catch

PR #1294, opened 2023-07-14
fix 'Tuple' is not defined at sentencepiece_unigram.py

fenglui:
/huggingface/tokenizers/bindings/python/py_src/tokenizers/implementations/sentencepiece_unigram.py", line 19, in SentencePieceUnigramTokenizer
    vocab: Optional[List[Tuple[str, float]]] = None,
NameError: name 'Tuple' is not defined. Did you mean: 'tuple'?

line 3 modified from
from typing import Iterator, List, Optional, Union
to
from typing import Iterator, List, Optional, Union, Tuple

fix this issue


fenglui:
fixed by another pr

PR #1303, opened 2023-07-25
Single warning for holes.

### LINKED ISSUE #1298: Is there a way to supress this warning?
```
warn!("The OrderedVocab you are attempting to save contains a hole for index {}, your vocabulary could be corrupted !", i);
```
Found here: https://github.com/huggingface/tokenizers/blob/5f6e9784526a4cd5e4f6dcdcc045cdceba5463e1/tokenizers/src/models/mod.rs#L42C53-L42C53

Our OrderedVocab has holes on purpose, so this warning is quite annoying.

Thanks!

Narsil:
Fixes #1298

### DIFF
--- a/tokenizers/src/models/bpe/serialization.rs
+++ b/tokenizers/src/models/bpe/serialization.rs
@@ -126,7 +126,6 @@ impl<'de> Visitor<'de> for BPEVisitor {
                 _ => {}
             }
         }
-        println!("Vocab {vocab} merges {merges}")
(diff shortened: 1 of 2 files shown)


ArthurZucker (approved):
I like the addition! Do we have an option to de-activate logs/warnings like LOGGING_INFO?

Narsil:
No and there won't be as mentionned in the issue. This is a hard error, and is not treated as such only because of old existing tokenizers we cannot modify.

PR #1275, opened 2023-06-12
Improve error for truncation with too high stride

boyleconnor:
Fixes #1269

@ArthurZucker I took the liberty of opening this PR. Let me know if this is a good format for the error message.

I ran tested that the message propagated through to the Python binding and also ran `make all-checks`; I'm not sure if there are any other pre-commit checks to run.

PR #1306, opened 2023-07-27
Give error when initializing tokenizer with too high stride

boyleconnor:
Related to #1269. Implements improvements requested [here](https://github.com/huggingface/tokenizers/pull/1275#pullrequestreview-1474362381).

Enabling truncation with bad parameters (specifically `stride` and `max_length`) on a tokenizer now gives an error (in the form of a proper Python exception, explaining the relation of added special tokens and max length to stride).

Note that I didn't remove the original `assert!()` for a few reasons:

1. I assumed that the number of added special tokens is always lowest when `is_pair = false`. I'm not sure if this is true/a justified assumption
2. I have not implemented any error output for the Node bindings. I haven't even looked at them yet, but I imagine they will enable truncation with bad params and ignore the `Err` output of the function. It might be nice to implement that functionality as part of this PR.

### DIFF
--- a/bindings/python/src/tokenizer.rs
+++ b/bindings/python/src/tokenizer.rs
@@ -712,8 +712,9 @@ impl PyTokenizer {
             }
         }
 
-        self.tokenizer.with_truncation(Some(params));
-
+        if let Err(error_message) = self.tokenizer.with_truncation(Some(params)) {
+            return Err(PyError(error_message.to_string()).into_pyerr::<exceptions::PyValueError>());
+        }
         Ok(())
     }
 
--- a/bindings/python/src/tokenizer.rs
+++ b/bindings/python/src/tokenizer.rs
@@ -721,7 +721,7 @@ impl PyTokenizer {
     /// Disable truncation
     #[pyo3(text_signature = "(self)")]
     fn no_truncation(&mut self) {
-        self.tokenizer.with_truncation(None);
+        let _ = self.tokenizer.with_truncation(None);
     }
 
     /// Get the currently set truncation parameters
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -902,11 +902,7 @@ where
         // 1. First we truncate if needed
         let (encoding, pair_encoding) = {
             if let Some(trunc) = &self.truncation {
-                let n_added_tokens = if let Some(processor) = &self.post_processor {
-                    processor.added_tokens(pair_encoding.is_some())
-                } else {
-                    0
-                };
+                let n_added_tokens = self.get_n_added_tokens(pair_encoding.is_some());
 
                 if add_special_tokens && n_added_tokens > 0 {
                     let params = TruncationParams {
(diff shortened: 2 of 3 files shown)


bindings/python/src/tokenizer.rs
```diff
@@ -712,15 +712,16 @@ impl PyTokenizer {
...
+            return Err(PyError(error_message.to_string()).into_pyerr::<exceptions::PyValueError>());
+        }
         Ok(())
     }
 
     /// Disable truncation
     #[pyo3(text_signature = "(self)")]
     fn no_truncation(&mut self) {
-        self.tokenizer.with_truncation(None);
+        let _ = self.tokenizer.with_truncation(None);
```
Narsil, line 724: Can you `expect` here ? no_truncation should never fail, but it's better to actually fail than be in the dark.

tokenizers/src/tokenizer/mod.rs
```diff
@@ -595,9 +599,21 @@ where
     }
 
     /// Set the truncation parameters
-    pub fn with_truncation(&mut self, trunc: Option<TruncationParams>) -> &mut Self {
+    ///
+    /// Fails if `stride` is too high relative to `max_length` and `post_processor.added_tokens()`
+    pub fn with_truncation(&mut self, trunc: Option<TruncationParams>) -> Result<&mut Self> {
+        if let Some(trunc_params) = &trunc {
+            let n_added_tokens = self.get_n_added_tokens(false);
```
Narsil, line 606: Why `false`  ? It depends.

Narsil (approved):
Thanks for the PR.

The code is good.
It's a breaking change, but it's only on the Rust side.

However I'm very curious when this is actually an issue. Having such large strides really seems a bug in the first place to me, almost looks like you're trying to get a sliding window out of it.

tokenizers/src/tokenizer/mod.rs
boyleconnor: I assumed that fewer tokens were added when `is_pair=false` (I'm not actually sure if this is a good assumption). So, when `with_truncation()` is called, we check if there is any possible case where this is a valid combination of `max_length` and `stride` for this tokenizer, and fail if it isn't.

Another reason why I kept the original `assert!()` from the inference part of the code is that it's possible that any additional special token(s) when `is_pair=true` will push it over the edge and make the `stride` too large for the `max_length`

boyleconnor:
> Having such large strides really seems a bug in the first place to me

You are right that this should probably never happen if someone is reading the documentation carefully (and properly understands HuggingFace's idiosyncratic definition of "stride").

However, it might come up with some toy examples. E.g. if you're playing around with tokenizers and you want to look at example outputs, you might try looking at the output of, say, the following

```python
tokenizer = tokenizers.Tokenizer.from_pretrained('bert-base-cased')
tokenizer.enable_truncation(5, stride=3)
print(tokenizer.encode("Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.")
```

would lead to this error

tokenizers/src/tokenizer/mod.rs
boyleconnor: @Narsil can you confirm whether my assumption in the above comment is a good one (i.e. is `added_tokens(false)` always less than `added_tokens(true)`?)

PR #1312, opened 2023-08-04
Check that processors add the number of tokens they say they will

boyleconnor:
The value output by `processor.added_tokens()` and the number of tokens added by `processor.process_encodings()` are two values that are defined separately (e.g. [with](https://github.com/huggingface/tokenizers/blob/efea6c72464428cabbf6332a16243e55963ba175/tokenizers/src/processors/bert.rs#L35) [BERT](https://github.com/huggingface/tokenizers/blob/efea6c72464428cabbf6332a16243e55963ba175/tokenizers/src/processors/bert.rs#L43)), but are required to be the same in order for `Tokenizer.encode()` to function correctly. If there is any discrepancy between these values, it can lead to wonky non-crashing behavior, such as truncating the wrong number of characters.

This PR adds a sanity check that these values are in fact the same.


Narsil:
No we cannot panic like that it's not OK.

I'm very fine if the reported numbers is not the perfect one, panicking randomly is not acceptable for it

boyleconnor:
@Narsil, I'm very surprised to read this:

> I'm very fine if the reported numbers is not the perfect one

If there is a mismatch between `added_tokens()` and the number of tokens actually added by `processor.process_encodings()`, then (if truncation is enabled) `Tokenizer.encode()` will truncate to the wrong length, which I think most Hugging Face users would consider completely unacceptable behavior (e.g. imagine if you are truncating the tokens on some input before sending it to an expensive API).

IIUC the only way that this failure (`added_tokens() != final_length - original_length`) would be possible is if a _Hugging Face contributor_ makes a mistake when implementing a `Processor` for a new tokenization system and the value of `added_tokens()` doesn't match the actual number of added tokens. I don't see how this could be "randomly" triggered without a significant systemic error being present in the `tokenizers` code base.

boyleconnor:
@ArthurZucker do you want to share thoughts on this PR directly? I believe you mentioned on the issue that you would check it out

ArthurZucker:
Hey! I don't think a new processor is planned, which is why it's not really worth the effort for now, but thanks a lot for wanting to contribute 🤗 I'm sorry in the delayed answer

boyleconnor:
Okay, I will close this PR, then

PR #1322, opened 2023-08-15
pyo3: update to 0.19

mikelui:
* python: update to pyo3 0.19

* Fix warnings due to deprecations


Narsil:
Can you run `python stub.py` and see what happens ?

It seems the actual output of the signatures has changed (which IIUC it shouldn't)

Narsil (approved):

PR #1331, opened 2023-08-24
Move to maturing mimicking move for `safetensors`.

Narsil:


Narsil:
Proving the build works
Screenshot from 2023-08-25 09-54-04

PR #1335, opened 2023-09-01
Update added tokens

### LINKED ISSUE #1334: `AddedTokens` loophole
AddedTokens is a little bit broken for a few things, will fix them. Here is an example: 

```python 
>>> from tokenizers import Tokenizer, AddedToken
>>> from tokenizers.models import BPE
>>> tokenizer = Tokenizer(BPE())
>>> tokenizer.add_tokens(["NOOO","This", " "])
3
>>> tokenizer.encode('NOOO This').tokens 
 ['NOOO', ' ', 'This']
>>> tokenizer.add_special_tokens([AddedToken("NOOO", rstrip = True)])
0
>>> tokenizer.encode('NOOO This').tokens 
['NOOO ', 'This']
>>> tokenizer.get_vocab()
{'NOOO': 0, ' ': 2, 'This': 1}
>>> tokenizer.decode([0,1,2], skip_special_tokens = True)
'This  '
```

So the output of `add_special_tokens` tells us that the token was not added. But the truth is the `added_tokens_map_r`  was updated.

ArthurZucker:
# What does this PR do?
Fixes #1334 and refactors the `AddedVocabulary`. Previously the when a token was already part of the vocabuary it was not added to the `added_tokens_map` but still added to `added_tokens_map_r` which is not consistent. Now it is added in both even if it already existed. 

Here is a small snippet of what is now possible in python

```python
>>> from tokenizers import AddedToken, Tokenizer
>>> token = AddedToken("HEY")
>>> tokenizer = Tokenizer.from_pretrained("gpt2")
>>> tokenizer.get_added_tokens_decoder()

>>> tokenizer.add_tokens([token])
1
>>> token.special = False
>>> tokenizer.add_tokens([token])
1 
>>> tokenizer.get_vocab_size()
50258
>>> content = tokenizer.decode([4])
'%'
>>> tokenizer.add_tokens([AddedToken('%')])
1
>>> tokenizer.get_added_tokens_decoder()
{4: AddedToken("%", rstrip=False, lstrip=False, single_word=False, normalized=True, special=False),
 50256: AddedToken("<|endoftext|>", rstrip=False, lstrip=False, single_word=False, normalized=True, special=True),
 50257: AddedToken("HEY", rstrip=False, lstrip=False, single_word=False, normalized=True, special=False)}
>>> tokenizer.get_vocab_size()
50258
```

### DIFF
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -668,12 +668,11 @@ where
     pub fn get_vocab_size(&self, with_added_tokens: bool) -> usize {
         // TODO ArthurZ THIS IS WRONG! We need to measure the length of the `set` because
         // now some tokens can be both in the added_tokens_encoder and in the vocab
-        self.model.get_vocab_size()
-            + if with_added_tokens {
-                self.added_vocabulary.len()
-            } else {
-                0
-            }
+        if with_added_tokens {
+            self.get_vocab(with_added_tokens).len()
+        } else {
+            self.model.get_vocab_size()
+        }
     }
 
     /// Converts a token in the corresponding id.
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -669,7 +669,7 @@ where
         // TODO ArthurZ THIS IS WRONG! We need to measure the length of the `set` because
         // now some tokens can be both in the added_tokens_encoder and in the vocab
         if with_added_tokens {
-            self.get_vocab(with_added_tokens).len()
+            self.get_vocab(true).len()
         } else {
             self.model.get_vocab_size()
         }
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -659,11 +659,9 @@ where
         final_vocab
     }
 
-    /// Get the added vocabulary only
-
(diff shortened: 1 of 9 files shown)


tokenizers/src/tokenizer/mod.rs
```diff
@@ -659,14 +659,20 @@ where
...
-        self.model.get_vocab_size()
-            + if with_added_tokens {
-                self.added_vocabulary.len()
-            } else {
-                0
-            }
+        // TODO ArthurZ THIS IS WRONG! We need to measure the length of the `set` because
+        // now some tokens can be both in the added_tokens_encoder and in the vocab
+        if with_added_tokens {
+            self.get_vocab(true).len()
```
Narsil, line 672: Why not use max(vocab_id) instead?

ArthurZucker: To account for potential holes 😢

ArthurZucker: And if we use max, we have to convert u32 to usize or return usize

Narsil (approved):
LGTM.

As said internally, I didn't have time to fully review, but most of my initial worries are addressed:

- `is_special_token` -> `special` is making things more uniform.
- The test breaking change is also fixing some other subtle bug within transformers when playing with special tokens (and we're releasing a breaking version anyway).

The rest looks OK.

PR #1341, opened 2023-09-19
Added ability to inspect a 'Sequence' pre-tokenizer.

eaplatanios:
I ran into this while attempting to create a modified copy of a `Tokenizer` instance after it has been created.

### DIFF
--- a/tokenizers/src/pre_tokenizers/sequence.rs
+++ b/tokenizers/src/pre_tokenizers/sequence.rs
@@ -6,27 +6,27 @@ use serde::{Deserialize, Serialize};
 #[derive(Clone, Debug, PartialEq)]
 #[macro_rules_attribute(impl_serde_type!)]
 pub struct Sequence {
-    pre_tokenizers: Vec<PreTokenizerWrapper>,
+    pretokenizers: Vec<PreTokenizerWrapper>,
 }
 
 impl Sequence {
-    pub fn new(pre_tokenizers: Vec<PreTokenizerWrapper>) -> Self {
-        Self { pre_tokenizers }
+    pub fn new(pretokenizers: Vec<PreTokenizerWrapper>) -> Self {
+        Self { pretokenizers }
     }
 
     pub fn get_pre_tokenizers(&self) -> &[PreTokenizerWrapper] {
-        &self.pre_tokenizers
+        &self.pretokenizers
     }
 
     pub fn get_pre_tokenizers_mut(&mut self) -> &mut [PreTokenizerWrapper] {
-        &mut self.pre_tokenizers
+        &mut self.pretokenizers
     }
 }
 
 impl PreTokenizer for Sequence {
     fn pre_tokenize(&self, pretokenized: &mut PreTokenizedString) -> Result<()> {
-        for pre_tokenizer in &self.pre_tokenizers {
-            pre_tokenizer.pre_tokenize(pretokenized)?;
+        for pretokenizer in &self.pretokenizers {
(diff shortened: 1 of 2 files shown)


eaplatanios:
Relatedly, what is the release policy for this project? Is there a possibility of making a patch release after this PR is merged (e.g., `0.14.1`)?

tokenizers/src/pre_tokenizers/sequence.rs
```diff
@@ -6,19 +6,27 @@ use serde::{Deserialize, Serialize};
...
-        Self { pretokenizers }
+    pub fn new(pre_tokenizers: Vec<PreTokenizerWrapper>) -> Self {
+        Self { pre_tokenizers }
+    }
+
+    pub fn get_pre_tokenizers(&self) -> &[PreTokenizerWrapper] {
+        &self.pre_tokenizers
+    }
+
+    pub fn get_pre_tokenizers_mut(&mut self) -> &mut [PreTokenizerWrapper] {
```
eaplatanios, line 21: These two new functions are just mimicking the interface you already have for the `Sequence` `Normalizer`.

eaplatanios:
cc @ArthurZucker @Narsil as I'm not sure what your process is for PR reviews and notifications.

ArthurZucker:
Hey, this is a feature addition so will not be in a `Patch` release but rather a minor release.

tokenizers/src/pre_tokenizers/sequence.rs
```diff
@@ -6,19 +6,27 @@ use serde::{Deserialize, Serialize};
 #[derive(Clone, Debug, PartialEq)]
 #[macro_rules_attribute(impl_serde_type!)]
 pub struct Sequence {
-    pretokenizers: Vec<PreTokenizerWrapper>,
+    pre_tokenizers: Vec<PreTokenizerWrapper>,
```
ArthurZucker, line 9: This seems like a breaking change to me.

ArthurZucker (commented):
Appart from the renaming that does not seem necessary, have no issue with this

tokenizers/src/pre_tokenizers/sequence.rs
eaplatanios: Though given this field is private I don't think it's a breaking change.

eaplatanios: I reverted it in either case.

eaplatanios:
> Hey, this is a feature addition so will not be in a `Patch` release but rather a minor release.

Sounds good, thanks @ArthurZucker!

Narsil:
Does this enable changing anything from python itself? 

I don't think you can give mutable references across FFI safely, so I'm not sure how this works.
Do you have a simple script to understand what you're trying to do?

Narsil:
I see! It's obvious now.

I'm not a huge fan of the `get_` prefixes of such methods in libs. But if there's précédent, uniformity is better!

Narsil (approved):
LGTM

eaplatanios:
Yeah likewise. I did it this way to keep things uniform with what was there already.

eaplatanios:
Will it be possible to make a release right after this PR is merged so I can directly use it? Or, if not, when is the next release scheduled for?

Narsil:
We don't have a super scheduled release plan.

Couldn´t you use directly the git branch in the meantime ?

eaplatanios:
Oh nice I didn't even know Cargo supported that. Very nice, thanks!

PR #1344, opened 2023-09-22
Let's allow hf_hub < 1.0

ArthurZucker:
An attempt to be more permissive with HF hub version

### DIFF
--- a/bindings/python/pyproject.toml
+++ b/bindings/python/pyproject.toml
@@ -26,7 +26,7 @@ dynamic = [
     'license',
     'readme',
 ]
-dependencies = ["huggingface_hub>=0.16.4,<0.17"]
+dependencies = ["huggingface_hub>=0.16.4,<1.0"]
 
 [project.urls]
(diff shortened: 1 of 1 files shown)


winglian:
@ArthurZucker when will this be merged? It's painful having to upgrade and downgrade back and forth to train and then upload  models

bindings/python/pyproject.toml
```diff
@@ -26,7 +26,7 @@ dynamic = [
     'license',
     'readme',
 ]
-dependencies = ["huggingface_hub>=0.16.4,<0.17"]
+dependencies = ["huggingface_hub>=0.16.4,<1.0"]
```
Narsil, line 29: No, we need to allow for breaking changes in huggingface_hub meaning we can do `>=0.17,<0.18` instead.

ArthurZucker, line 29:
```suggestion
dependencies = ["huggingface_hub>=0.16.4,<0.18"]
```

Narsil (approved):
LGTM

PR #1355, opened 2023-10-06
Preparing release.

Narsil:


tokenizers/src/models/bpe/trainer.rs
```diff
@@ -533,15 +533,16 @@ impl BpeTrainer {
             let changes = top
                 .pos
                 .maybe_par_iter()
-                .flat_map(|i| {
-                    let w = &words[*i] as *const _ as *mut _;
+                .flat_map(|&i| {
+                    let word = &words[i] as *const _ as *mut Word;
```
Narsil, line 537: @ArthurZucker Let's keep track of this. The newer compiler raised the casting of `&Word to &mut Word` as undefined behavior (meaning they can break it whenever they want).

Here I'm just keeping the raw pointer instead which works. But maybe things start breaking from this at some point.

PR #1356, opened 2023-10-07
fix a clerical error  in the comment

tiandiweizun:


ArthurZucker (approved):
Thanks for the catch!

PR #1357, opened 2023-10-07
[`pre_tokenizers`] Fix sentencepiece based Metaspace

ArthurZucker:
Fix the issue with meta spaces that applies the pre_tokenization to all the sub strings.
```rust
use tokenizers::pre_tokenizers::metaspace::{Metaspace, PrependScheme};
use tokenizers::{PreTokenizedString, PreTokenizer};
use regex::Regex;
let pretok = Metaspace::new_with_prepend_scheme('▁', true, PrependScheme::Always)
let mut pretokenized = PreTokenizedString::from("Hey my friend <s>how▁are you");
let re_ref = Regex::new(r"(<s>)").unwrap();
pretokenized
  .split(|_, sequence| sequence.split(&re_ref, SplitDelimiterBehavior::Isolated))
  .expect("AddedVocabulary bad split");
pretok.pre_tokenize(&mut pretokenized).unwrap();
```
with legacy:
```rust
vec![("▁Hey", (0, 6)),("▁my", (6, 11)),("▁friend", (11, 20)),("▁", (20, 23)),("<s>", (23, 26)),("how", (26, 29)),("▁are", (29, 35)),("▁you", (35, 41))]);
```
without legacy:
```rust
 vec![("▁Hey", (0, 6)),("▁my", (6, 11)),("▁friend", (11, 20)),("▁", (20, 23)),("▁<s>", (23, 29)),("▁how", (29, 35)),("▁are", (35, 41)),("▁you", (41, 47))]);
```

### DIFF
--- a/bindings/python/src/pre_tokenizers.rs
+++ b/bindings/python/src/pre_tokenizers.rs
@@ -452,7 +452,7 @@ impl PySequence {
     }
 }
 
-fn _from_string(string: String) -> Result<PrependScheme, PyErr> {
+fn from_string(string: String) -> Result<PrependScheme, PyErr> {
     let scheme = match string.as_str() {
         "first" => PrependScheme::First,
         "never" => PrependScheme::Never,
--- a/bindings/python/src/pre_tokenizers.rs
+++ b/bindings/python/src/pre_tokenizers.rs
@@ -518,7 +518,7 @@ impl PyMetaspace {
 
     #[setter]
     fn set_prepend_scheme(self_: PyRef<Self>, prepend_scheme: String) -> PyResult<()> {
-        let scheme = _from_string(prepend_scheme)?;
+        let scheme = from_string(prepend_scheme)?;
         setter!(self_, Metaspace, @set_prepend_scheme, scheme);
         Ok(())
     }
@@ -536,7 +536,7 @@ impl PyMetaspace {
 
         // If a prepend scheme is provided, set it
         if let Some(prepend_scheme) = prepend_scheme {
-            match _from_string(prepend_scheme) {
+            match from_string(prepend_scheme) {
                 Ok(prepend_scheme_enum) => new_instance.set_prepend_scheme(prepend_scheme_enum),
                 Err(err) => return Err(err),
             }
--- a/bindings/python/src/pre_tokenizers.rs
+++ b/bindings/python/src/pre_tokenizers.rs
@@ -11,7 +11,7 @@ use tk::pre_tokenizers::bert::BertPreTokenizer;
 use tk::pre_tokenizers::byte_level::ByteLevel;
 use tk::pre_tokenizers::delimiter::CharDelimiterSplit;
 use tk::pre_tokenizers::digits::Digits;
-use tk::pre_tokenizers::metaspace::Metaspace;
+use tk::pre_tokenizers::metaspace::{Metaspace, PrependScheme};
 use tk::pre_tokenizers::punctuation::Punctuation;
 use tk::pre_tokenizers::split::Split;
 use tk::pre_tokenizers::unicode_scripts::UnicodeScripts;
@@ -491,12 +491,31 @@ impl PyMetaspace {
 
     #[getter]
     fn get_prepend_scheme(self_: PyRef<Self>) -> String {
-        getter!(self_, Metaspace, prepend_scheme.to_string())
+        // Assuming Metaspace has a method to get the prepend_scheme as a string
+        let scheme: PrependScheme = getter!(self_, Metaspace, get_prepend_scheme());
+        match scheme {
+            PrependScheme::First => "First",
+            PrependScheme::Never => "Never",
+            PrependScheme::Always => "Always",
+        }.to_string()
     }
 
     #[setter]
-    fn set_prepend_scheme(self_: PyRef<Self>, prepend_scheme: String) {
-        setter!(self_, Metaspace, @set_prepend_scheme, prepend_scheme);
+    fn set_prepend_scheme(self_: PyRef<Self>, prepend_scheme: &str) -> PyResult<()> {
+        // Assuming Metaspace has a method to set the prepend_scheme from a string
+        let scheme = match prepend_scheme.to_lowercase().as_str() {
+            "first" => PrependScheme::First,
+            "never" => PrependScheme::Never,
+            "always" => PrependScheme::Always,
+            _ => {
+                return Err(exceptions::PyValueError::new_err(format!(
+                    "Unknown variant: {}",
+                    prepend_scheme
+                )))
+            }
+        };
+        setter!(self_, Metaspace, @set_prepend_scheme, scheme);
+        Ok(())
     }
 
     #[new]
@@ -505,24 +524,35 @@ impl PyMetaspace {
         replacement: PyChar,
         add_prefix_space: bool,
         _kwargs: Option<&PyDict>,
-    ) -> (Self, PyPreTokenizer) {
+    ) -> PyResult<(Self, PyPreTokenizer)>  {
         // Create a new Metaspace instance
         let mut new_instance: Metaspace = Metaspace::new(replacement.0, add_prefix_space);
 
-        // Extract values from _kwargs if present
-        let prepend_scheme: Option<String> = _kwargs.and_then(|kwargs| {
+        if let Some(prepend_scheme) = _kwargs.and_then(|kwargs| {
             kwargs
                 .get_item("prepend_scheme")
-                .and_then(|prepend_scheme| prepend_scheme.extract().ok())
-        });
-        // Perform an action if prepend_scheme is present
-        if let Some(prepend_scheme_value) = prepend_scheme {
-            // Do something with prepend_scheme_value
-            new_instance.set_prepend_scheme(prepend_scheme_value);
+                .and_then(|prepend_scheme| prepend_scheme.extract::<String>().ok())
+        }) {
+            // Unwrap the Option<String> before calling to_lowercase
+            let prepend_scheme_value:String = prepend_scheme.to_lowercase();
+
+            let prepend_scheme_enum = match prepend_scheme_value.as_str() {
+                "first" => PrependScheme::First,
+                "never" => PrependScheme::Never,
+                "always" => PrependScheme::Always,
+                _ => {
+                    return Err(exceptions::PyValueError::new_err(format!(
+                        "Unknown variant: {}",
+                        prepend_scheme_value
+                    )))
+                }
+            };
+
+            // Set the prepend_scheme in the new instance
+            new_instance.set_prepend_scheme(prepend_scheme_enum);
         }
 
-        // Return the new Metaspace instance and PyPreTokenizer
-        (PyMetaspace {}, new_instance.into())
+        Ok((PyMetaspace {}, new_instance.into()))
     }
 }
 
--- a/bindings/python/src/pre_tokenizers.rs
+++ b/bindings/python/src/pre_tokenizers.rs
@@ -497,7 +497,8 @@ impl PyMetaspace {
             PrependScheme::First => "First",
             PrependScheme::Never => "Never",
             PrependScheme::Always => "Always",
-        }.to_string()
+        }
+        .to_string()
     }
 
     #[setter]
@@ -524,7 +525,7 @@ impl PyMetaspace {
         replacement: PyChar,
         add_prefix_space: bool,
         _kwargs: Option<&PyDict>,
-    ) -> PyResult<(Self, PyPreTokenizer)>  {
+    ) -> PyResult<(Self, PyPreTokenizer)> {
         // Create a new Metaspace instance
         let mut new_instance: Metaspace = Metaspace::new(replacement.0, add_prefix_space);
 
@@ -534,7 +535,7 @@ impl PyMetaspace {
                 .and_then(|prepend_scheme| prepend_scheme.extract::<String>().ok())
         }) {
             // Unwrap the Option<String> before calling to_lowercase
-            let prepend_scheme_value:String = prepend_scheme.to_lowercase();
+            let prepend_scheme_value: String = prepend_scheme.to_lowercase();
 
             let prepend_scheme_enum = match prepend_scheme_value.as_str() {
                 "first" => PrependScheme::First,
--- a/bindings/python/src/pre_tokenizers.rs
+++ b/bindings/python/src/pre_tokenizers.rs
@@ -494,9 +494,9 @@ impl PyMetaspace {
         // Assuming Metaspace has a method to get the prepend_scheme as a string
         let scheme: PrependScheme = getter!(self_, Metaspace, get_prepend_scheme());
         match scheme {
-            PrependScheme::First => "First",
-            PrependScheme::Never => "Never",
-            PrependScheme::Always => "Always",
+            PrependScheme::First => "first",
+            PrependScheme::Never => "never",
+            PrependScheme::Always => "always",
         }
         .to_string()
     }
--- a/bindings/python/src/pre_tokenizers.rs
+++ b/bindings/python/src/pre_tokenizers.rs
@@ -523,24 +523,19 @@ impl PyMetaspace {
     }
 
     #[new]
-    #[pyo3(signature = (replacement = PyChar('▁'), add_prefix_space = true, **_kwargs), text_signature = "(self, replacement=\"_\", add_prefix_space=True)")]
+    #[pyo3(signature = (replacement = PyChar('▁'), add_prefix_space = true, prepend_scheme=None, **_kwargs), text_signature = "(self, replacement=\"_\", add_prefix_space=True)")]
     fn new(
         replacement: PyChar,
         add_prefix_space: bool,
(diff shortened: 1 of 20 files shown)


tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -65,12 +69,17 @@ impl Default for Metaspace {
...
         pretokenized.split(|_, mut normalized| {
             normalized.replace(' ', &self.str_rep)?;
-            if self.add_prefix_space && !normalized.get().starts_with(self.replacement) {
-                normalized.prepend(&self.str_rep);
+            if self.add_prefix_space{
+                if  self.legacy{normalized.prepend(&self.str_rep);}
+                else if self.add_prefix_space && first_split{
+                    normalized.prepend(&self.str_rep);
+                    first_split = false; // Set the flag to false after the first split
+                }
```
ArthurZucker, line 81: Legacy is not a new arg for BC, but can be easily set

ArthurZucker:
Serialization is not working properly yet! Will have to fix this

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -9,10 +9,20 @@ use crate::tokenizer::{Decoder, PreTokenizedString, PreTokenizer, Result, SplitD
 pub struct Metaspace {
     replacement: char,
     pub add_prefix_space: bool,
+    #[serde(skip_serializing_if = "skip_legacy_serialization")]
+    pub legacy: bool,
```
ArthurZucker, line 13: for consistency, let's try to properly always save it, gotta check FC.

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -65,12 +80,18 @@ impl Default for Metaspace {
 
 impl PreTokenizer for Metaspace {
     fn pre_tokenize(&self, pretokenized: &mut PreTokenizedString) -> Result<()> {
+        let mut first_split = true;
+
         pretokenized.split(|_, mut normalized| {
             normalized.replace(' ', &self.str_rep)?;
             if self.add_prefix_space && !normalized.get().starts_with(self.replacement) {
-                normalized.prepend(&self.str_rep);
+                if self.legacy {
```
ArthurZucker, line 88: here `!normalized.get().starts_with(self.replacement)` might make us add it

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -188,6 +211,55 @@ mod tests {
         );
     }
 
+    #[test]
```
ArthurZucker, line 214: test pre_tokenized from transformers

tokenizers/src/pre_tokenizers/mod.rs
```diff
@@ -104,6 +104,28 @@ mod tests {
...
             ]))
         );
+
+        let pre_tokenizer: PreTokenizerWrapper = serde_json::from_str(
+            r#"{"type":"Metaspace","replacement":"▁","add_prefix_space":true, "legacy":false}"#,
+        )
+        .unwrap();
+
+        let mut expected_pre_tokenizer = Metaspace::new('▁', true);
+        expected_pre_tokenizer.legacy = false;
```
ArthurZucker, line 114: add a new `new` function that supports 3 arguments

tokenizers/src/pre_tokenizers/metaspace.rs
ArthurZucker: rename "only_first" or "prepend_once" etc

ArthurZucker: ENUM prepend, once, never to support statefull-like changes to this. More descriptive

bindings/python/src/pre_tokenizers.rs
```diff
@@ -489,17 +489,40 @@ impl PyMetaspace {
...
         replacement: PyChar,
         add_prefix_space: bool,
         _kwargs: Option<&PyDict>,
     ) -> (Self, PyPreTokenizer) {
-        (
-            PyMetaspace {},
-            Metaspace::new(replacement.0, add_prefix_space).into(),
-        )
+        // Create a new Metaspace instance
+        let mut new_instance: Metaspace = Metaspace::new(replacement.0, add_prefix_space);
```
ArthurZucker, line 510: I don't use the `new_with_prepend_type` because otherwise I need to implement the PyPrependScheme, which I'd rather avoid. My question here @Narsil is whether or not I should just allow string prepend scheme in the `new_with_prepend` using the `set_prepend_scheme` internally?

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -1,6 +1,45 @@
...
+    }
+}
+
+impl From<String> for PrependScheme {
+    fn from(s: String) -> Self {
+        PrependScheme::from(s.as_str())
+    }
+}
+
+impl fmt::Display for PrependScheme {
```
ArthurZucker, line 34: required to make sure that python side only sees strings

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -44,6 +95,20 @@ impl Metaspace {
...
+    pub fn new_with_prepend_scheme(
+        replacement: char,
+        add_prefix_space: bool,
+        prepend_scheme: PrependScheme,
+    ) -> Self {
+        Self {
+            replacement,
+            str_rep: replacement.to_string(),
+            add_prefix_space,
+            prepend_scheme,
```
ArthurZucker, line 111: we could do `perpend_scheme.into()` and allow passing `<Into PrependScheme>` as type

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -65,10 +133,19 @@ impl Default for Metaspace {
...
             normalized.replace(' ', &self.str_rep)?;
             if self.add_prefix_space && !normalized.get().starts_with(self.replacement) {
-                normalized.prepend(&self.str_rep);
+                if self.prepend_scheme == PrependScheme::Always {
+                    normalized.prepend(&self.str_rep);
+                } else if self.prepend_scheme == PrependScheme::First && first_split {
+                    normalized.prepend(&self.str_rep);
+                    first_split = false;
+                }
+            } else {
```
ArthurZucker, line 147: never is not used, not a problem IMO

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -188,6 +266,116 @@ mod tests {
...
+                ("▁", (20, 23)),
+                ("▁<s>", (23, 29)),
+                ("▁how", (29, 35)),
+                ("▁are", (35, 41)),
+                ("▁you", (41, 47))
+            ]
+        );
+
+        pretok.set_prepend_scheme("first");
+        let mut pretokenized = PreTokenizedString::from(" Hey <s>how"); // test with prefix
```
ArthurZucker, line 335: make sure the prefix is not added twice

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -188,6 +266,116 @@ mod tests {
...
+                .collect::<Vec<_>>(),
+            vec![
+                ("▁Hey", (0, 6)),
+                ("▁", (6, 9)),
+                ("<s>", (9, 12)),
+                ("how", (12, 15))
+            ]
+        );
+
+        let mut pretokenized = PreTokenizedString::from(" Hey <s>how <s>are <s> you"); // test with many splits
```
ArthurZucker, line 354: make sure the prefix is added once

tokenizers/src/tokenizer/mod.rs
ArthurZucker, line 1: I revert this and add it in a seperate PR

bindings/python/src/pre_tokenizers.rs
Narsil: string for enums seems the easiest way.

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -1,6 +1,45 @@
...
+pub enum PrependScheme {
+    /// Specifies that the scheme should be prepended only once, on the first split.
+    First,
+    /// Specifies that the space should not be prepended.
+    Never,
+    /// Specifies that the scheme should always be prepended.
+    Always,
+}
+
+impl From<&str> for PrependScheme {
```
Narsil, line 17: This probably only belongs in the PyO3 layer. In pure rust the String doesn't seem very necessary.

tokenizers/src/pre_tokenizers/metaspace.rs
Narsil: This should be in the Python layer too, not here. Regular Debug implem should suffit imo in pure rust.

tokenizers/src/pre_tokenizers/metaspace.rs
Narsil: I don't think we should. Real type is better.

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -55,6 +120,9 @@ impl Metaspace {
         self.replacement = replacement;
         self.str_rep = replacement.to_string();
     }
+    pub fn set_prepend_scheme(&mut self, scheme: impl Into<PrependScheme>) {
+        self.prepend_scheme = scheme.into();
```
Narsil, line 124: Just make that `PrependScheme`.

`impl Into<T>` has an associated cost with it. Which is that there will be a concrete function for EVERY possible caller that might arise (here at least `PrependScheme`, `&str` and `String`. 

This is nice to make an API easier to use when the concrete underlying type is relatively verbose to make, or there are many structs implementing the given trait that might be useful (something like `impl Response` in a web framework.

Here everything is a simple `enum`. Importing the enum and using a variant makes the function  so that only 1 exists, prevents all kind of runtiem errors (all will become compile time errors). And since only 1 function exists, it can be inline more easily by the compiler.

Simple is better here IMO

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -1,6 +1,45 @@
...
+    Always,
+}
+
+impl From<&str> for PrependScheme {
+    fn from(s: &str) -> Self {
+        match s.to_lowercase().as_str() {
+            "first" => PrependScheme::First,
+            "never" => PrependScheme::Never,
+            "always" => PrependScheme::Always,
+            _ => panic!("Invalid value for PrependScheme: {}", s),
```
Narsil, line 23: Panicking is NOT okay in transformers. We should NEVER panic since we're a library.

For that there is `TryFrom` which returns `Result<T>`. However I don't think this should be done in the Rust layer but in the Python layer instead.

tokenizers/src/pre_tokenizers/mod.rs
```diff
@@ -104,6 +104,30 @@ mod tests {
...
+        );
+
+        let pre_tokenizer: PreTokenizerWrapper = serde_json::from_str(
+            r#"{"type":"Metaspace","replacement":"▁","add_prefix_space":true, "prepend_scheme":"always"}"#,
+        )
+        .unwrap();
+
+        assert_eq!(
+            pre_tokenizer,
+            PreTokenizerWrapper::Metaspace(Metaspace::new('▁', true))
```
Narsil, line 129: Can you switch every testcase test to `new_with_preprend_scheme` ? Just makes the test easier to understand.

`new(..) == new_with_prepend(.., Always)` could become a test too imo.

Narsil (commented):
x

ArthurZucker:
On it!

bindings/python/src/pre_tokenizers.rs
```diff
@@ -489,17 +489,71 @@ impl PyMetaspace {
...
+    #[setter]
+    fn set_prepend_scheme(self_: PyRef<Self>, prepend_scheme: &str) -> PyResult<()> {
+        // Assuming Metaspace has a method to set the prepend_scheme from a string
+        let scheme = match prepend_scheme.to_lowercase().as_str() {
+            "first" => PrependScheme::First,
+            "never" => PrependScheme::Never,
+            "always" => PrependScheme::Always,
+            _ => {
+                return Err(exceptions::PyValueError::new_err(format!(
+                    "Unknown variant: {}",
```
Narsil, line 513: Let's say which ones are OK ?

bindings/python/src/pre_tokenizers.rs
```diff
@@ -489,17 +489,71 @@ impl PyMetaspace {
...
+            PrependScheme::Never => "never",
+            PrependScheme::Always => "always",
+        }
+        .to_string()
+    }
+
+    #[setter]
+    fn set_prepend_scheme(self_: PyRef<Self>, prepend_scheme: &str) -> PyResult<()> {
+        // Assuming Metaspace has a method to set the prepend_scheme from a string
+        let scheme = match prepend_scheme.to_lowercase().as_str() {
```
Narsil, line 507:
```suggestion
        let scheme = match prepend_scheme.as_str() {
```

I know we like to be very accepting of parameters, but refusing things makes code overall for everyone more consistent which is better in my eyes (unless backward compat forces something like this)

bindings/python/src/pre_tokenizers.rs
```diff
@@ -489,17 +489,71 @@ impl PyMetaspace {
...
-    ) -> (Self, PyPreTokenizer) {
-        (
-            PyMetaspace {},
-            Metaspace::new(replacement.0, add_prefix_space).into(),
-        )
+    ) -> PyResult<(Self, PyPreTokenizer)> {
+        // Create a new Metaspace instance
+        let mut new_instance: Metaspace = Metaspace::new(replacement.0, add_prefix_space);
+
+        if let Some(prepend_scheme) = _kwargs.and_then(|kwargs| {
```
Narsil, line 532: I could be wrong, but accepting ` prepend_scheme: Option<String>`  should already get you halfway there, meaning you would have better error handling and automated handling of this arg,

bindings/python/src/pre_tokenizers.rs
```diff
@@ -489,17 +489,71 @@ impl PyMetaspace {
...
+            // Unwrap the Option<String> before calling to_lowercase
+            let prepend_scheme_value: String = prepend_scheme.to_lowercase();
+
+            let prepend_scheme_enum = match prepend_scheme_value.as_str() {
+                "first" => PrependScheme::First,
+                "never" => PrependScheme::Never,
+                "always" => PrependScheme::Always,
+                _ => {
+                    return Err(exceptions::PyValueError::new_err(format!(
+                        "Unknown variant: {}",
```
Narsil, line 546: Factor this out so we only have 1 error message to fix.

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -44,6 +67,20 @@ impl Metaspace {
             replacement,
             str_rep: replacement.to_string(),
             add_prefix_space,
+            prepend_scheme: PrependScheme::Always, // always prepend for legacy purpose
+        }
+    }
+
+    pub fn new_with_prepend_scheme(
```
Narsil, line 74: Kind of syntactic, but making `new`  call `new_with_prepend`  should make everything less error prone (since there's less room for modifying one path without modifying the other.

tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -55,6 +92,14 @@ impl Metaspace {
         self.replacement = replacement;
         self.str_rep = replacement.to_string();
     }
+
+    pub fn get_prepend_scheme(&self) -> PrependScheme {
+        self.prepend_scheme.clone()
```
Narsil, line 97: Don't clone on behalf of users, never.

Either return a reference `&PrependScheme`.
Or make `PrependScheme`  ` Copy`.

Making something Copy, is something you should do only when the size makes it worthwile.
A reference is a pointer of size `usize` so copying usize is usually much faster than copying an entire struct.
For an enum like PrependScheme, it' s only 3 possible values, encoded internally by rust as `u8` or `usize` (Don't remember which). In any case, it's prefereable to copy the value than to pass references around (which cost a pointer dereference)

bindings/python/src/pre_tokenizers.rs
```diff
@@ -489,17 +489,67 @@ impl PyMetaspace {
...
-            Metaspace::new(replacement.0, add_prefix_space).into(),
-        )
+    ) -> PyResult<(Self, PyPreTokenizer)> {
+        // Create a new Metaspace instance
+        let mut new_instance: Metaspace = Metaspace::new(replacement.0, add_prefix_space);
+
+        // If a prepend scheme is provided, set it
+        if let Some(prepend_scheme) = prepend_scheme {
+            let prepend_scheme_enum = match prepend_scheme.as_str() {
+                "First" => PrependScheme::First,
```
Narsil, line 539: Please factor this code out ! :D. You shouldn't need to have this twice, Make it a simple function.

Also why change `"First"`  ? `"first"` is much more pythonic imo.

PR #1378, opened 2023-10-27
Rename modeled `token_to_id`

chris-ha458:
Currently, there are two different public `token_to_id` methods.

One can be found in `tokenizers/src/tokenizer/added_vocabulary.rs`
as part of the `AddedVocabulary` struct

another is found in `tokenizers/src/tokenizer/mod.rs`
as part of `TokenizerImpl` struct

This causes issues with type inference especially with 
https://github.com/huggingface/tokenizers/blob/c718c53bb95884752badfbb27920396bb2df1518/tokenizers/src/tokenizer/added_vocabulary.rs#L259

where `rust-analyzer` complains of [E0107](https://doc.rust-lang.org/stable/error_codes/E0107.html)

Since it technically builds its not a real error, and the compiler itself eventually finds out the correct version to use, I think renaming one of the functions would be the best option.

Changing visibility might help the problem go away, but 
https://github.com/huggingface/tokenizers/blob/c718c53bb95884752badfbb27920396bb2df1518/tokenizers/src/tokenizer/added_vocabulary.rs#L202
```
    pub fn token_to_id(&self, token: &str, model: &impl Model) -> Option<u32> {
        self.added_tokens_map
            .get(token)
            .copied()
            .or_else(|| model.token_to_id(token))
    }
```
will still remain ambiguous in such cases.

I think multiple approaches  including more informative fn names could be helpful as well
(since `AddedVocabulary` 's `token_to_id` is used only there change it's version into `addedvocab_token_to_id`?)

### DIFF
--- a/tokenizers/src/tokenizer/added_vocabulary.rs
+++ b/tokenizers/src/tokenizer/added_vocabulary.rs
@@ -199,7 +199,7 @@ impl AddedVocabulary {
     }
 
     /// Get the id matching one of our token if it exists
-    pub fn token_to_id(&self, token: &str, model: &impl Model) -> Option<u32> {
+    pub fn token_to_id_model(&self, token: &str, model: &impl Model) -> Option<u32> {
         self.added_tokens_map
             .get(token)
             .copied()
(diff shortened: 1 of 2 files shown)


ArthurZucker:
Hey! renaming this function can easily break backwards compatibility WDYT?

PR #1382, opened 2023-11-03
Fix truncation length assertion

### LINKED ISSUE #1326: `stride` assertion check no longer catches all invalid truncation parameters
Why was [this line](https://github.com/huggingface/tokenizers/blob/9a93c50c25c1e0b73a85584f327113bcbef5ac80/tokenizers/src/tokenizer/mod.rs#L608) changed in 9a93c50? I am pretty sure it should be `<=`, not `<`. [Further downstream](https://github.com/huggingface/tokenizers/blob/9a93c50c25c1e0b73a85584f327113bcbef5ac80/tokenizers/src/tokenizer/encoding.rs#L319), when the truncation is actually run, it will panic if `!(stride < max_len)`, i.e. `max_len <= stride`, so I don't see why this check shouldn't match.

I tested with the following code (reminder: BERT adds exactly 2 special tokens in this case, so the effective max length is actually `8`):

```python
tokenizer = Tokenizer.from_pretrained('bert-base-cased')
tokenizer.enable_truncation(max_length=10, stride=8)  # This should fail but doesn't
```

which doesn't create an error (even though it should). But it will (correctly) produce an error when the following is run:

```python
tokenizer.encode("This piece of text is at least ten tokens long. In fact, it is likely many more than that.")
```

```
  File "/home/connor/PycharmProjects/tokenizers/test_stride_warning.py", line 11, in <module>
    main()
  File "/home/connor/PycharmProjects/tokenizers/test_stride_warning.py", line 7, in main
    tokenizer.encode("This piece of text is at least ten tokens long. In fact, it is likely many more than that.")
pyo3_runtime.PanicException: `stride` must be strictly less than `max_len=8` (note that `max_len` may be shorter than the max length of the original model, as it subtracts the number of special characters
```

boyleconnor:
Fixes #1326 

I haven't seen any explanation from @Narsil why [the line in question](https://github.com/huggingface/tokenizers/blob/9a93c50c25c1e0b73a85584f327113bcbef5ac80/tokenizers/src/tokenizer/mod.rs#L608) was changed, so I just went ahead and made this PR reversing it.

I've tested my version of this locally and can confirm that there are no longer two different errors at different places for slightly different invalid stride values:

```python

tokenizer = Tokenizer.from_pretrained('bert-base-cased')
tokenizer.enable_truncation(max_length=10, stride=9)  # This line still fails (as it should)
print(tokenizer.encode("This piece of text is at least ten tokens long. In fact, it is likely many more than that."))

tokenizer = Tokenizer.from_pretrained('bert-base-cased')
tokenizer.enable_truncation(max_length=10, stride=8)  # Now this line (correctly) fails too
print(tokenizer.encode("This piece of text is at least ten tokens long. In fact, it is likely many more than that."))
```

### DIFF
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -605,7 +605,7 @@ where
         if let Some(trunc_params) = &trunc {
             let n_added_tokens = self.get_n_added_tokens(false);
             let effective_max_length = trunc_params.max_length - n_added_tokens;
-            if effective_max_length < trunc_params.stride {
+            if effective_max_length <= trunc_params.stride {
                 return Err(Box::new(TruncationParamError(format!(
                     "tokenizer stride set to {}, which is greater than or equal to its effective max length of {} (= {} original max length - {} added special tokens), ",
                     trunc_params.stride, effective_max_length, trunc_params.max_length, n_added_tokens


ArthurZucker:
Sorry @boyleconnor I'll see if this should be merged or not, I am not sure either 😉 This was not really breaking but the fix is breaking in a way

boyleconnor:
@ArthurZucker I'm not sure what you what you mean, would you mind elaborating?

PR #1396, opened 2023-11-21
fuzz: Add a BPE training fuzzer

silvergasp:


silvergasp:
See #1397 for context. Also note that I've found a couple of non-critical bugs with the fuzzer's already. I'll open up some issues with further details shortly :)

PR #1414, opened 2023-12-13
Faster HF dataset iteration in docs

mariosasko:
`Dataset.iter` is orders of magnitude faster for iterating over "contiguous" HF datasets.


ArthurZucker (approved):
LGTM thanks

PR #1419, opened 2023-12-18
add option to skip special tokens

### LINKED ISSUE #1347: Mitigate prompt injection attacks by supporting "safe" encoding (encoding without special tokens)
*There may already exist a way of accomplishing what I'm going to describe but I didn't find it by reading the documentation.*

In certain applications, we should be careful about how special tokens are encoded as they can be used to trigger special capabilities in models, or give them special positional clues (system prompt, etc.). Hence, when serving a model to end users, we need to prevent injection attacks, in which the user sends the representation of a special token as plain text (eg. `<SYSTEM>`), and the tokenizer interprets this text as a special token. In this regard, OpenAI's `tiktoken` tokenizer has a very safe default of raising an exception if it encounters text that corresponds to a special token, [see the corresponding docstring](https://github.com/openai/tiktoken/blob/39f29cecdb6fc38d9a3434e5dd15e4de58cf3c80/tiktoken/core.py#L75). This effectively forces the developer to be very intentional about how special tokens should be handled, thus preventing injection attacks.

Such a default behavior would break existing code, an alternative would be to have a `.safe_encode` method that would throw an exception if it encounters text that corresponds to a special token, mirroring what `tiktoken` is doing, and allowing/disallowing special tokens using a whitelist/blacklist argument. Disallowed special tokens should be treated as plain text and NOT as representation of special tokens, ie. `<SYSTEM>` should be tokenized as `["<", "SYSTEM", ">"]` or otherwise depending on the vocabulary, but most importantly it should NOT be interpreted as the `<SYSTEM>` token unless explicitly enabled by the developer.

Is there an existing way of mirroring `tiktoken` behavior, and if not, would such a feature be useful to the library?

### LINKED ISSUE #1368: How can we ignore special tokens when encoding text
Hi, I'm using tokenizer for pretraining these days and I add some special tokens which I want to disable when I'm encoding the pretrain text. I tried the `add_special_tokens = False` for tokenizer.encode but it didn't work.
Any ideas or solutions will be appeciated.

### LINKED ISSUE #1391: How to split special token in encode?
i have converted a slow tokenizer into PreTrainedTokenizerFast, and get a tokenizer.json file.But i found that this tokenizer did not split special tokens.Here is my add_tokens in tokenizer.json:
` tokenizer.add_special_tokens(
                [
                    AddedToken("[gMASK]", normalized=True, single_word=False),
                    AddedToken("sop", normalized=True, single_word=False),
                ]
            )
`

ArthurZucker:
Allow skipping special tokens when encoding 

fixes #1347, fixes #1391 fixes #1368

### DIFF
--- a/bindings/python/py_src/tokenizers/__init__.pyi
+++ b/bindings/python/py_src/tokenizers/__init__.pyi
@@ -838,6 +838,18 @@ class Tokenizer:
 
         """
         pass
+    @property
+    def encode_special_tokens(self):
+        """
+        Modifies the tokenizer in order to use or not the special tokens
+        during encoding.
+
+        Args:
+            value (:obj:`bool`):
+                Whether to use the special tokens or not
+
+        """
+        pass
     @staticmethod
     def from_buffer(buffer):
         """
@@ -1032,19 +1044,6 @@ class Tokenizer:
 
             pretty (:obj:`bool`, defaults to :obj:`True`):
                 Whether the JSON file should be pretty formatted.
-        """
-        pass
-    def set_encode_special_tokens(self, value):
-        """
-        Modifies the tokenizer in order to use or not the special tokens
-        during encoding.
-
-        Args:
-            value (:obj:`bool`):
-                Whether to use the special tokens or not
-
-        Returns:
-            :obj:`Optional[str]`: An optional token, :obj:`None` if out of vocabulary
         """
         pass
     def to_str(self, pretty=False):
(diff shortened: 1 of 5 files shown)


ArthurZucker:
This works as expected for now: 
```python
>>> from transformers import AutoTokenizer
>>> tokenizer = AutoTokenizer.from_pretrained("gp2") 
>>> tokenizer.tokenize("<|endoftext|>")
['<', '|', 'end', 'of', 'text', '|', '>']
```

ArthurZucker:
TODO : 
- [ ] Evaluate on a benchmark if this does not slow down too much

ArthurZucker:
Before merging I just want to add a getter, and make sure we can just set it with `tokenizer.encode_special_tokens = True`.

ArthurZucker:
PR is not on the correct branch lol

PR #1443, opened 2024-01-22
Added ability to inspect a 'Sequence' decoder and the `AddedVocabulary`.

eaplatanios:
This PR is in similar spirit to #1341 and adds a couple more functions that allow one to construct a modified version of an existing `Tokenizer`. I've followed the existing style and conventions for newly introduced functions.

### DIFF
--- a/tokenizers/src/decoders/sequence.rs
+++ b/tokenizers/src/decoders/sequence.rs
@@ -13,6 +13,14 @@ impl Sequence {
     pub fn new(decoders: Vec<DecoderWrapper>) -> Self {
         Self { decoders }
     }
+
+    pub fn get_decoders(&self) -> &[DecoderWrapper] {
+        &self.decoders
+    }
+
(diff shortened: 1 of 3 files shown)


eaplatanios:
@ArthurZucker @Narsil gentle ping about this PR. This PR should not be controversial and is in similar spirit to #1341 (and has the same motivation).

ArthurZucker:
really sorry about all the delays, lot happening on transformers, I'll free some time

ArthurZucker (approved):
Actually doesn't look bad and would help me in transformers if we add the layer to update the added vocab. This is not super stateless, so leave it up to @Narsil

Narsil (approved):
LGTM.

It's ok because it doesn´t seem like we open up any function that have potentially bad internal state (Deserialize is not implemented for AddedVocabulary directly, itś always parsed from the entire Tokenizer).

ArthurZucker:
Thanks for your contribution 🤗

PR #1462, opened 2024-02-28
feat: support custom regexes for GPT pre-tokenizer

gcampax:
In order to properly support GPT3.5/GPT4 models which changed the regex compared to GPT2

Existing tokenizer files are not affected. New tokenizer files can be created that copy the new regex from the tiktoken sources.

I'm new to Rust, so the code probably doesn't look very idiomatic. Happy to adjust it.

### DIFF
--- a/tokenizers/src/pre_tokenizers/byte_level.rs
+++ b/tokenizers/src/pre_tokenizers/byte_level.rs
@@ -37,7 +37,7 @@ fn bytes_char() -> HashMap<u8, char> {
 lazy_static! {
     /// Regex that matches exactly one token.
     /// See https://github.com/openai/gpt-2/blob/master/src/encoder.py#L98
-    static ref RE: SysRegex = SysRegex::new(
+    static ref GPT2_RE: SysRegex = SysRegex::new(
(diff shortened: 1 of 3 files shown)


Narsil:
There's no need for that, just use   `use_regex:false` and use a regex externally if you want.

gcampax:
I'm sorry, but this is not correct. use_regex false does not apply in this context: you do need to apply the regular expression. Applying the regex externally also doesn't apply: token splitting by regex is done inside the preprocessor. I don't know how you would apply it externally.
It's also quite inconvenient to have special case external logic for OpenAI tokenizers, instead of being able to specify a JSON file that actually works.

The PR is not that big, I would kindly ask you to reconsider.

ArthurZucker:
Hey! I am down to re-consider, I think that Narsil meant is that you can have a sequence of `pre_processor`, to first do the `regex` then apply bytelevel  (this would be "external"). But to be honest, since there is a regex inside, why waste it!

ArthurZucker:
@gcampax do you want to rebase and I'll review?

PR #1488, opened 2024-04-07
Fix data directory for test

atupone:


atupone:
python binding test fails (to me) due to a wrong path of the data directory.
Both works:
either this commit,
or a symbolic link of the data directory at the same level of tests

PR #1492, opened 2024-04-12
Fix unsoundness in `tokenizers::utils::parallelism`

### LINKED ISSUE #1491: Unsound use of unsafe in `src/utils/parallelism.rs`
The static variable `USED_PARALLELISM` is accessed in `has_parallelism_been_used`, and modified in `MaybeParallelIterator::into_maybe_par_iter` and `MaybeParallelBridge::maybe_par_bridge`.
All these cases are unsafe because if they are done at the same time from different threads, they could cause a data race which has undefined behaviour.
The issue is that all of these are safe functions, and no checks are made to ensure that the unsafe operations are in fact safe.
This means that it would be possible to cause UB in safe rust by calling these functions from separate threads.
There maybe reasons to believe that this is unlikely or impossible given the rest of the library (I don't know the code base well enough to say), but that does not change the fact that this is unsound.

The easiest way to fix this would be to place the variable in a `Mutex`. I propose an implementation of this in #1492 .

albertsgarde:
Fixes #1491 by turning `static mut USED_PARALLELISM: bool` into a `Mutex`, i.e. `static USED_PARALLELISM: Mutex<bool>`. This allows the value to be accessed and modified safely and avoids unsoundness. There is a slight performance penalty, but given that the variable is only modified in the context of other parallel work, it is almost certainly negligible.

This PR fails several tests, but only the tests `main` also fails.

### DIFF
--- a/tokenizers/src/utils/parallelism.rs
+++ b/tokenizers/src/utils/parallelism.rs
@@ -2,6 +2,8 @@
 //! This module defines helpers to allow optional Rayon usage.
 //!
 
+use std::sync::Mutex;
+
 use rayon::iter::IterBridge;
 use rayon::prelude::*;
 use rayon_cond::CondIterator;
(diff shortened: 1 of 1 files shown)


ArthurZucker:
Hey sorry for the delay I'll try to review for next week!

albertsgarde:
@ArthurZucker I think #1532 is a superior solution, so this PR should probably just be closed

PR #1493, opened 2024-04-15
Add more support for tiktoken based tokenizers

ArthurZucker:
Adds a check before using merges, returing the token if it is part of the vocab

### DIFF
--- a/tokenizers/src/models/bpe/model.rs
+++ b/tokenizers/src/models/bpe/model.rs
@@ -462,7 +462,7 @@ impl BPE {
     fn tokenize_with_cache(&self, sequence: &str) -> Result<Vec<Token>> {
         if let Some(ref hit) = self.cache.as_ref().and_then(|c| c.get(sequence)) {
             Ok(self.word_to_tokens(hit).collect())
-        } else if self.vocab.contains_key(sequence) && self.ignore_merges {
+        } else if let Some(id) = self.vocab.get(sequence) && self.ignore_merges {
             let id = self.vocab.get(sequence);
             let ret = Token::new(*id.unwrap(), sequence.to_string().clone(), (0, 0));
             Ok(vec![ret])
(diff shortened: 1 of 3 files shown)


Narsil (approved):
LGTM preemptively.

Not sure about this name (even if I agree it definitely looks like a bug.)

tokenizers/src/models/bpe/model.rs
```diff
@@ -449,6 +462,10 @@ impl BPE {
     fn tokenize_with_cache(&self, sequence: &str) -> Result<Vec<Token>> {
         if let Some(ref hit) = self.cache.as_ref().and_then(|c| c.get(sequence)) {
             Ok(self.word_to_tokens(hit).collect())
+        } else if self.vocab.contains_key(sequence) && self.ignore_merges {
```
Narsil, line 465:
```suggestion
        } else if let Some(id) = self.vocab.get(sequence) && self.ignore_merges {
```

ArthurZucker (commented):
Alright finalized

tokenizers/src/models/bpe/model.rs
```diff
@@ -449,6 +462,17 @@ impl BPE {
...
+        } else if let Some(id) = self.vocab.get(sequence) {
+            if self.ignore_merges {
+                Ok(vec![Token::new(*id, sequence.to_string().clone(), (0, 0))])
+            } else {
+                let word = self.merge_word(sequence)?;
+                let ret = self.word_to_tokens(&word).collect();
+                if let Some(ref cache) = self.cache {
+                    cache.set(sequence.to_owned(), word);
+                }
+                Ok(ret)
```
Narsil, line 474: No don't repeat yourself :)

Narsil: Ideally check for the flag before looking into the hashmap (should be faster because of less branch mispredictions).
Not a huge deal.

ArthurZucker: Ok

ArthurZucker: But transfomrers is all about repeat yourself 😐

Narsil (approved):

PR #1521, opened 2024-04-30
remove enforcement of non special when adding tokens

ArthurZucker:
Fix the issue that prevented us from adding special and non special tokens using a single call to `add_tokens`. 
Very important because calling this multiple times adds a huge slowdown (regex re-compilation).

```python 
>>> from transformers import AutoTokenizer, AddedToken
>>> tokenizer = AutoTokenizer.from_pretrained("meta-llama/Meta-Llama-3-8B")
Special tokens have been added in the vocabulary, make sure the associated word embeddings are fine-tuned or trained.

>>> tokenizer._tokenizer.add_tokens([AddedToken("<me-and-you>", special=True, normalized=False)])
1

>>> tokenizer._tokenizer.decode([128256], skip_special_tokens=True)
''
```
before it would output `'<me-and-you>'`. and the token was always special when added.
This will allow us to do the following cleanup: https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils_fast.py#L169-L192 

```diff 
-        encoder = list(self.added_tokens_encoder.keys()) + [str(token) f
-        # if some of the special tokens are strings, we check if we don'
-        tokens_to_add += [
-            token for token in self.all_special_tokens_extended if token
-        ]
-        if len(tokens_to_add) > 0:
-            # super hack: if a token.special is set, tokenizer ignores i
-            # Accumulate added tokens into batches of special/non-specia
-            # individual tokens would repeatedly rebuild a trie, which c
-            is_last_special = None
-            tokens = []
-            special_tokens = self.all_special_tokens
-            for token in tokens_to_add:
-                is_special = (
-                    (token.special or str(token) in special_tokens)
-                    if isinstance(token, AddedToken)
-                    else str(token) in special_tokens
-                )
-                if is_last_special is None or is_last_special == is_spec
-                    tokens.append(token)
-                else:
-                    self._add_tokens(tokens, special_tokens=is_last_spec
-                    tokens = [token]
-                is_last_special = is_special
-            if tokens:
-                self._add_tokens(tokens, special_tokens=is_last_special)
+ self._add_tokens(tokens_to_add)
```

this issue comes from https://github.com/huggingface/tokenizers/commit/c02d4e2202314cfc4e23bce16f7991261f4b2c93 which was introduced in v0.8

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -102,7 +102,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: audit
-          args: -D warnings -f ./bindings/python/Cargo.lock
+          args: -D warnings --ignore RUSTSEC-2024-0335 -f ./bindings/python/Cargo.lock
 
       - name: Install
         working-directory: ./bindings/python
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -95,6 +95,9 @@ jobs:
           command: clippy
           args: --manifest-path ./bindings/python/Cargo.toml --all-targets --all-features -- -D warnings
 
+      - name: Install cargo-audit
+        run: cargo install cargo-audit
+
       - name: Run Audit
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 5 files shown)


Narsil (approved):
LGTM.

Not sure if we should mark it as breaking chnage (it is technically, although we're more respectful of user intent which could categorize as bugfix)

PR #1532, opened 2024-05-18
Make `USED_PARALLELISM` atomic

### LINKED ISSUE #1491: Unsound use of unsafe in `src/utils/parallelism.rs`

nathaniel-daniel:
Fixes #1491

This fixes UB with the `USED_PARALLELISM` variable, rationale described in #1491. Additionally, `static mut` is difficult to use correctly, to the point where it is likely to be removed in a future edition. Instead, I have replaced the `bool` with an `AtomicBool`, which can be used without unsafe. 

This approach is superior to #1492, as using a mutex for a bool is wasteful.

### DIFF
--- a/tokenizers/src/utils/parallelism.rs
+++ b/tokenizers/src/utils/parallelism.rs
@@ -5,14 +5,15 @@
 use rayon::iter::IterBridge;
 use rayon::prelude::*;
 use rayon_cond::CondIterator;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
 
 // Re-export rayon current_num_threads
 pub use rayon::current_num_threads;
 
 pub const ENV_VARIABLE: &str = "TOKENIZERS_PARALLELISM";
 
-// Reading/Writing this variable should always happen on the main thread
-static mut USED_PARALLELISM: bool = false;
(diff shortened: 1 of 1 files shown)


ArthurZucker (approved):
Looks good to me! @narsil WDYT?

Narsil (approved):

Narsil:
Thanks a lot for this ! 

Much better indeed than the existing code.

Narsil:
Let's ignore the clippy lints, I'll tackle them in another PR.

PR #1535, opened 2024-05-18
feat: add support for pyarrow arrays as input

### LINKED ISSUE #1415: Support PyArrow arrays as tokenizer input
Most data processing libraries (Datasets, Polars, Pandas, DuckDB, etc.) are integrated with PyArrow, so native (zero-copy if possible) support for PyArrow arrays as input to avoid the unnecessary PyArrow to Python/NumPy conversion (pretty slow for string arrays) would be nice.
 
PS: PyArrow has recently added support for the PyCapsule interface, which should help with the implementation.

notjedi:
closes #1415

### DIFF
--- a/bindings/python/src/tokenizer.rs
+++ b/bindings/python/src/tokenizer.rs
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 use std::collections::{hash_map::DefaultHasher, HashMap};
 use std::hash::{Hash, Hasher};
+use std::slice;
 
 use numpy::{npyffi, PyArray1};
 use pyo3::class::basic::CompareOp;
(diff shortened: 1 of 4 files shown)


notjedi:
yet to add support for `pyarrow.LargeString`

notjedi:
pulling `arrow` from git because the current stable version `v51.0` links against `v0.20` of `pyo3` while the bindings link against `v0.21`. there is already a merged pull request(apache/arrow-rs#5566) which migrates from `v0.20` to `v0.21` of `pyo3` which i think will be released in `v52.0`

bindings/python/src/tokenizer.rs
```diff
@@ -260,13 +262,43 @@ impl PyAddedToken {
     }
 }
 
+struct PyArrowScalarStringInput<'s>(Cow<'s, str>);
```
ArthurZucker, line 265:
```suggestion
[cfg(feature="pyarrow")]
struct PyArrowScalarStringInput<'s>(Cow<'s, str>);
```

ArthurZucker (commented):
I think we can leave the dep as optional wdyt?

notjedi:
> I think we can leave the dep as optional wdyt?

makes sense, will get it changed @ArthurZucker

notjedi:
i've added `pyarrow` as a feature in `pyproject.toml`. not sure if that is what we want here. let me know if i need to fix anything else. @ArthurZucker

ArthurZucker (commented):
I think we want to avoid having to use git for the repo 😢 
ALso do you have any bench / examples of it being use and fast? 🚀

strategy155:
Good afternoon everyone. How do you think the optimal benchmark should look like? Native pyarrow vs numpy conversion for string arrays?

ArthurZucker:
something like that yeah, I have no idea in what context pyarrow is used as I have not used it, but in optimal context usage

strategy155:
I'm not sure it should be closed. The pr is looking fine.

notjedi:
yes, just update to arrow to the latest stable version `v52.0`

Issue #1541, opened 2024-05-25
Deserializing BPE tokenizer failure


mcognetta:
I am trying to serialize and deserialize a tokenizer and am getting an error:

```
Exception: data did not match any variant of untagged enum ModelWrapper at ...
```

First, I want to clarify that I have seen the other related issues to this (e.g., #1342, #566, #909, #1297, etc.), and none of the fixes in them apply here (I will detail that below). I have tested this on two different corpora and over 3 different versions of tokenizers (`v0.12`, `v0.13`, and `v0.19`) with the same results.

Here are the things I have checked/tried (to confirm that I have checked the other bugs):
- (#909) There are no spaces in any of the tokens
- (#909) All merges have exactly one space in their representation
- (#566) I've tried with no pre_tokenizer, `Whitespace()`, `WhitespaceSplit()`, and `Split(pattern='\w+|[^\w\s]+', behavior='isolated')`
- (#1342) I've tried with no `post_processor` and a `Template` processor
- I've tried loading from a string vs from a file (both immediately after the tokenization was created)
- I've tried with many different vocabulary sizes (including so small that no merges are present)

My model training code is:

```python
def build_bpe_tokenizer(path, vocab_size):
    tokenizer = Tokenizer(BPE(unk_token="[UNK]"))
    tokenizer.pre_tokenizer = WhitespaceSplit()
    special_tokens = ["[UNK]", "[PAD]", "[CLS]", "[SEP]", "[MASK]"]
    trainer = BpeTrainer(
        vocab_size=vocab_size,
        continuing_subword_prefix="##",
        show_progress=False,
        special_tokens=special_tokens,
    )
    tokenizer.train([path], trainer)
    tokenizer.save("tokenizer.json")
    return tokenizer
```
which was taken basically verbatim from the documentation.

**EDIT: I actually had `tokenizer.model.dropout = 0.0` in my code, which was the cause of the failure (see the closing comment).**

Immediately reloading `Tokenizer.from_file("tokenizer.json")` fails with the above error.

Any ideas on how to work around this?

----------------------------------

Below is an example json output from the tokenizer training code above. Note that I set the vocabulary size to be such that only one merge was added. I did not modify anything in the json file. Please try to load this (and let me know if it works, it fails for me).

```
{
  "version": "1.0",
  "truncation": null,
  "padding": null,
  "added_tokens": [
    {
      "id": 0,
      "content": "[UNK]",
      "single_word": false,
      "lstrip": false,
[...]

PR #1550, opened 2024-06-07
Enable `dropout = 0.0` as an equivalent to `none` in BPE

mcognetta:
This is related to the discussion in #1541.

This PR allows for `0.0` to be used as the dropout value in BPE models with equivalent functionality to `none`. Previously, the docs and implementation were inconsistent:
- the docs said 0.0 to 1.0 were acceptable values for dropout and that the default was 0
    - the default was actually `none`
- it was possible to set dropout to 0.0 once the model was built (and it would work the same as if it was `none`)
    - however, in this case caching was disabled
    - on a small benchmark, i measured a 50% increase in encoding time with dropout = 0.0 compared to none (this discrepancy goes away after this PR is applied)
- serialization worked, but deserialization failed, since there was a check that `dropout \in (0.0, 1.0]`
- initializing a python BPE model with dropout 0.0 failed (e.g. `BPE(dropout = 0.0)`)

This simply allows for `0.0` to be an acceptable value during initialization and enables caching when tokenizing if `dropout == 0.0`.

E.g., now the following works

```python
>>> from tokenizers import Tokenizer, models
>>> tokenizer = Tokenizer(models.BPE(dropout = 0.0))
>>> s = tokenizer.to_str()
>>> s
'{"version":"1.0","truncation":null,"padding":null,"added_tokens":[],"normalizer":null,"pre_tokenizer":null,"post_processor":null,"decoder":null,"model":{"type":"BPE","dropout":0.0,"unk_token":null,"continuing_subword_prefix":null,"end_of_word_suffix":null,"fuse_unk":false,"byte_fallback":false,"ignore_merges":false,"vocab":{},"merges":[]}}'
>>> deserialized = Tokenizer.from_str(s)
>>> deserialized.model.dropout
0.0
```

whereas before it errored.

-----------------------------

As future work, I think that dropout should be made non-optional, with the default being 0.0. This would remove the checks for `dropout.is_none()`, etc, but keep the functionality the same. However, I guess this would be a breaking change (since then all tokenizers serialized before this change would be invalid?).

### DIFF
--- a/bindings/python/tests/bindings/test_models.py
+++ b/bindings/python/tests/bindings/test_models.py
@@ -73,6 +73,7 @@ class TestBPE:
         model = BPE(dropout=0.0)
         assert model.dropout == 0.0
 
+
 class TestWordPiece:
     def test_instantiate(self, bert_files):
         assert isinstance(WordPiece(), Model)
--- a/bindings/python/tests/bindings/test_models.py
+++ b/bindings/python/tests/bindings/test_models.py
@@ -73,7 +73,6 @@ class TestBPE:
         model = BPE(dropout=0.0)
         assert model.dropout == 0.0
 
-
 class TestWordPiece:
     def test_instantiate(self, bert_files):
         assert isinstance(WordPiece(), Model)
--- a/bindings/python/tests/bindings/test_models.py
+++ b/bindings/python/tests/bindings/test_models.py
@@ -69,6 +69,9 @@ class TestBPE:
         model.byte_fallback = True
         assert model.byte_fallback == True
 
+    def test_dropout_zero(self):
+        model = BPE(dropout=0.0)
+        assert model.dropout == 0.0
 
 class TestWordPiece:
     def test_instantiate(self, bert_files):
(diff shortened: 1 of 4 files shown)


ArthurZucker (approved):
LGTM! Thanks for adding tests in python and rust.

ArthurZucker:
rebasing on main should fix clippy issues

mcognetta:
Thanks. I fixed one of the lint errors which was a range readability thing.

mcognetta:
Fixed one more formatting issue. _Now_ I think it should be all good!

.github/workflows/trufflehog.yml
ArthurZucker, line 1: Actually this should not be in the diff

mcognetta:
I goofed, and now its getting worse 😬

mcognetta:
Ok, it's all good now, unless you want me to squash the commits.

ArthurZucker (approved):
No don't worry, this LGMT! thanks for updating

PR #1554, opened 2024-06-17
make sure we don't warn on empty tokens

ArthurZucker:
When activating warnings, there was an issue with added tokens that are not part of the vocab. I am pretty sure that the main point about added tokens is that they don't have to be in the vocab, so fixing this.

### DIFF
--- a/tokenizers/src/tokenizer/serialization.rs
+++ b/tokenizers/src/tokenizer/serialization.rs
@@ -177,7 +177,7 @@ where
 mod tests {
     use crate::tokenizer::Tokenizer;
     use std::str::FromStr;
-    use tracing_test::traced_test;
+    use tracing_subscriber::fmt;
 
     #[test]
     fn test_deserialization_serialization_invariant() {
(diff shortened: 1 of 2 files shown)


tokenizers/src/tokenizer/serialization.rs
```diff
@@ -233,4 +232,15 @@ mod tests {
...
+    #[cfg(feature = "http")]
+    #[test]
+    fn test_from_pretrained() {
+        fmt()
+            .with_max_level(tracing::Level::DEBUG)
+            .with_target(false)
+            .init();
+        let _ = Tokenizer::from_pretrained("Qwen/Qwen2-7B-Instruct", None);
+        warn!("This should be the first warning");
+    }
```
McPatate, line 245: What is the purpose of this test?

McPatate (approved):
lgtm, I'll trust you that this is the correct behaviour, not too versed on this topic 😅

ArthurZucker:
Will fix clippy on main and merge!

PR #1555, opened 2024-06-18
Add bytelevel normalizer to fix decode when adding tokens to BPE

### LINKED ISSUE #1392: added_tokens with bytemap charaters in ByteLevel could not be decoded correctly
I just found that if added tokens contain some characters that exist in the byte map for ByteLevel preprocessor could not be decoded correctly.
This is a script to reproduce the problem with version 0.14.1
```python
from tokenizers import normalizers
from tokenizers.pre_tokenizers import (
    ByteLevel,
)
from tokenizers import decoders
tokenizer = Tokenizer(BPE())
tokenizer.normalizer = normalizers.Sequence([])

tokenizer.pre_tokenizer = Sequence(
    [
        ByteLevel(add_prefix_space=False, use_regex=False),
    ])
tokenizer.add_tokens(["ilÖveyou"])
# Ö is the character representing for 0xf6
tokenizer.decoder = decoders.ByteLevel()
encode_result = tokenizer.encode("ilÖveyou")
print(encode_result.ids)
print(tokenizer.decode(encode_result.ids))
```
the output wil be 
```
[0]
il�veyou
```

I believe the problem comes from 
https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/mod.rs#L832-L836
I don't think added token should be sent to bytelevel decoder for it is extacted before pretokenize.

ArthurZucker:
This revert the previous breaking change. 

Also add a new `ByteLevel` normalizer, which replaces the ByteLevel pre_tokenizer.
Checked that we can add chines / Cyrillic tokens which are properly encoded and decoder. 

Fixes #1392

### DIFF
--- a/tokenizers/src/normalizers/byte_level.rs
+++ b/tokenizers/src/normalizers/byte_level.rs
@@ -1,0 +1,180 @@
+use crate::processors::byte_level::bytes_char;
+use crate::tokenizer::{NormalizedString, Normalizer, Result};
+use serde::{Deserialize, Serialize};
+use std::collections::{HashMap, HashSet};
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(tag = "type")]
+pub struct ByteLevel {}
+
+lazy_static! {
+    static ref BYTES_CHAR: HashMap<u8, char> = bytes_char();
+    static ref CHAR_BYTES: HashMap<char, u8> =
+        bytes_char().into_iter().map(|(c, b)| (b, c)).collect();
+}
+
+impl Default for ByteLevel {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ByteLevel {
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    pub fn alphabet() -> HashSet<char> {
+        BYTES_CHAR.values().copied().collect()
+    }
+}
+
+impl Normalizer for ByteLevel {
+    /// Strip the normalized string inplace
+    fn normalize(&self, normalized: &mut NormalizedString) -> Result<()> {
+        if !normalized.is_empty() {
+            let s = normalized.get();
+            let mut transformations: Vec<(char, isize)> = Vec::with_capacity(s.len());
+            let mut i = 0;
+            for cur_char in s.chars() {
+                let size = cur_char.len_utf8();
+                let bytes = s[i..i + size].as_bytes();
+                i += size;
+                transformations.extend(
+                    bytes
+                        .iter()
+                        .enumerate()
+                        .map(|(i, b)| (BYTES_CHAR[b], isize::from(i > 0))),
+                );
+            }
+            normalized.transform(transformations, 0);
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+
+    #[test]
+    fn test_byte_level_normalize() {
+        let original = "Hello 我今天能为你做什么";
+        let normalized = "HelloĠæĪĳä»Ĭå¤©èĥ½ä¸ºä½łåģļä»Ģä¹Ī";
+        assert_ne!(original, normalized);
+        let mut n = NormalizedString::from(original);
+        let byte_level = ByteLevel::new();
+        byte_level.normalize(&mut n).unwrap();
+        assert_eq!(&n.get(), &normalized);
+        assert_eq!(
+            n,
+            NormalizedString::new(
+                original.to_string(),
+                normalized.to_string(),
+                vec![
+                    (0, 1),
+                    (1, 2),
+                    (2, 3),
+                    (3, 4),
+                    (4, 5),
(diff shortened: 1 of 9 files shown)


tokenizers/src/tokenizer/mod.rs
```diff
@@ -847,35 +847,22 @@ where
...
+                self.added_vocabulary
+                    .id_to_token(*id, &self.model)
+                    .filter(|token| {
+                        !skip_special_tokens || !self.added_vocabulary.is_special_token(token)
+                    })
+            })
+            .collect::<Vec<_>>();
+
+        if let Some(decoder) = &self.decoder {
+            decoder.decode(tokens)
```
ArthurZucker, line 862: reverted to what we originally had

tokenizers/src/pre_tokenizers/byte_level.rs
```diff
@@ -11,7 +11,7 @@ use crate::utils::macro_rules_attribute;
 
 /// Converts bytes to unicode characters.
 /// See https://github.com/openai/gpt-2/blob/master/src/encoder.py#L9
-fn bytes_char() -> HashMap<u8, char> {
+pub fn bytes_char() -> HashMap<u8, char> {
```
McPatate, line 14:
```suggestion
pub(crate) fn bytes_char() -> HashMap<u8, char> {
```

tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -216,17 +216,14 @@ impl AddedVocabulary {
     }
 
     /// Get the token matching the given id if it exists
-    #[deprecated(
-        since = "0.19.0",
-        note = "please use `added_vocabulary.simple_id_to_token(id).or_else(|| model.id_to_token(id)` instead"
-    )]
```
McPatate, line 222: Is this not deprecated anymore?

tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -216,17 +216,14 @@ impl AddedVocabulary {
...
     pub fn id_to_token(&self, id: u32, model: &impl Model) -> Option<String> {
         self.added_tokens_map_r
             .get(&id)
             .map(|t| t.content.clone())
             .or_else(|| model.id_to_token(id))
     }
 
+    //
```
McPatate, line 226:
```suggestion
```

tokenizers/src/tokenizer/pre_tokenizer.rs
```diff
@@ -51,7 +51,7 @@ impl From<(NormalizedString, Option<Vec<Token>>)> for Split {
 /// original string.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct PreTokenizedString {
-    original: String,
+    pub original: String,
```
McPatate, line 54: I'm not seeing `.original` anywhere, why do you need this?

Also if it is for internal usage, `pub(crate)`

tokenizers/src/normalizers/byte_level.rs
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    pub fn alphabet() -> HashSet<char> {
+        BYTES_CHAR.values().copied().collect()
+    }
+}
+
+impl Normalizer for ByteLevel {
+    /// Strip the normalized string inplace
+    fn normalize(&self, normalized: &mut NormalizedString) -> Result<()> {
```
McPatate, line 34: Why doesn't this live in `NormalizedString` like so:
```rs
impl NormalizedString {
  fn normalize_byte_level(&mut self) -> Result<()> { // ... }
}
```
?
Feels a bit weird to have an empty stateless struct for just a function, but may be due to the structure of tokenizers / python.

tokenizers/src/models/bpe/trainer.rs
```diff
@@ -282,8 +282,8 @@ impl BpeTrainer {
         for c in &self.initial_alphabet {
             alphabet
                 .entry(*c)
-                .and_modify(|cnt| *cnt = std::usize::MAX)
-                .or_insert(std::usize::MAX);
+                .and_modify(|cnt| *cnt = usize::MAX)
+                .or_insert(usize::MAX);
```
McPatate, line 286: did rustfmt do that?

tokenizers/src/tokenizer/added_vocabulary.rs
ArthurZucker: will rebase on main to fix this

tokenizers/src/tokenizer/pre_tokenizer.rs
ArthurZucker: good catch, I think I no longer need this!

tokenizers/src/models/bpe/trainer.rs
ArthurZucker: clippy but it breaks stuff

PR #1568, opened 2024-07-11
[Fix metaspace prepending scheme] ⛓️‍💥⛓️‍💥

ArthurZucker:
Fixes Metaspace again

image

### DIFF
--- a/bindings/python/tests/bindings/test_tokenizer.py
+++ b/bindings/python/tests/bindings/test_tokenizer.py
@@ -540,7 +540,7 @@ class TestTokenizer:
 
     def metaspace_pre_tokenizer(self):
         tokenizer = Tokenizer.from_pretrained("t5-base")
-        tokenizer.pre_tokenizer.prepend_scheme = "firs"
+        tokenizer.pre_tokenizer.prepend_scheme = "first"
         tokenizer.add_tokens(["<bos>"])
         encoding = tokenizer.encode("<bos>New").ids
         assert encoding == [4, 32300, 123]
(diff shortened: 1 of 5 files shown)


tokenizers/src/pre_tokenizers/metaspace.rs
```diff
@@ -115,29 +118,50 @@ impl Default for Metaspace {
...
     fn pre_tokenize(&self, pretokenized: &mut PreTokenizedString) -> Result<()> {
-        pretokenized.split(|_, mut normalized| {
-            normalized.replace(' ', &self.str_rep)?;
-            match self.prepend_scheme {
-                PrependScheme::Always => {
-                    if !normalized.get().starts_with(self.replacement) {
-                        normalized.prepend(&self.str_rep);
-                    }
+        let result = if self.prepend_scheme == PrependScheme::First {
+            let first = &mut pretokenized.splits[0];
```
Narsil, line 122: This will panic if splits is empty, isn't 

```rust
pretokenized.split(|_, mut normalized| {
->
pretokenized.split(|i, mut normalized| {
...
if i == 0
```

Enough for a change ?

bindings/python/tests/bindings/test_tokenizer.py
```diff
@@ -493,6 +493,7 @@ def test_splitting(self):
         tokenizer.pre_tokenizer.split = False
         tokenizer.add_tokens([AddedToken("<REPR_END>", rstrip=True, lstrip=True)])
         assert tokenizer.encode("<REPR_END>inform<s>. Hey.       .", add_special_tokens=False).tokens == [
+            "▁",
```
Narsil, line 496: This seems like a bug no ?

<REPR_END> is a token on it's own, we shouldn't prepend with `_` no ?

tokenizers/src/pre_tokenizers/metaspace.rs
Narsil: Also for good measure, the reason it checks offsets.original == 0 instead of `i==0` is because users can send pre pretokenized strings, meanings you can have split n still being "the first token". (Since it's transparent for us at this stage if it was pretiokenized strings from us, or from the users, only the offsets retain that information)

PR #1570, opened 2024-07-12
[Feature] support Assign token to update the content of a token

### LINKED ISSUE #1473: Assign `<unusedXX>` tokens with `special_tokens` without growing vocab size
``
I'm trying to modify `google/gemma-7b` tokenizer for instruction tuning purposes. My goal is to replace some of the "unused" tokens that were specifically added to the tokenizer for my own defined "custom" tokens. I want these custom tokens to be treated as "special" (i.e. not normalized, stripped, etc.), however this seems like an impossible task.

What I would like to do is some version of the following,

```python
tokenizer = AutoTokenizer.from_pretrained("google/gemma-7b")

custom_tokens = ['<|im_start|>', '<|im_end|>']
unused_tokens = ['<unused1>', '<unused2>']

tokenizer.add_special_tokens({'additional_special_tokens': custom_tokens, 'tokens_to_replace': unused_tokens})
```

Given that many models/tokenizers being open-sourced specifically reserve some set of unused tokens for this purpose, I would like to make use of them without growing the vocabulary, and subsequently not having to adjust the model's embedding dimensions.

I've tried manually manipulating the vocab, and assigning appropriate dicts on the forward and reverse pass (encoder, decoder), but nothing seems to work.

How can I achieve my goals of making use of unused tokens, ensuring they are treated as "special", and not growing the vocabulary of the tokenizer and model embedding?

ArthurZucker:
Very draft for now:
- [ ] handle potential collisions

fixes #1473

### DIFF
--- a/bindings/python/py_src/tokenizers/__init__.pyi
+++ b/bindings/python/py_src/tokenizers/__init__.pyi
@@ -725,6 +725,23 @@ class Tokenizer:
         """
         pass
 
+    def assing_tokens(self, old_tokens, new_tokens):
+        """
+        Add the given tokens to the vocabulary
+
+        The given tokens are added only if they don't already exist in the vocabulary.
+        Each token then gets a new attributed id.
+
+        Args:
+            tokens (A :obj:`List` of :class:`~tokenizers.AddedToken` or :obj:`str`):
+                The list of tokens we want to add to the vocabulary. Each token can be either a
+                string or an instance of :class:`~tokenizers.AddedToken` for more customization.
+
+        Returns:
+            :obj:`int`: The number of tokens that were created in the vocabulary
+        """
+        pass
+
     def decode(self, ids, skip_special_tokens=True):
         """
         Decode the given list of ids back to a string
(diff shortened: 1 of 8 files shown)


pritam-dey3:
We really need this feature! Is it tracked anywhere else?

ArthurZucker:
oups sorry, yeah here is the good place will come back in a bit!

ArthurZucker:
Okay, so this PR works for assigning, but the issue is that for `Unigram` based models if the world is not removed from the internal lattice, it can still be "encoded". Might be an issue. 
We are also updating the `added_tokens` layer on top of `vocab`

ArthurZucker:
Related to #1437 that does not work for unigram as if the token is in the vocabulary, even if it is special, it's still added to the unigram algorithm!

jp1924:
related issue: #1473, [#31475](https://github.com/huggingface/transformers/issues/31475)

ArthurZucker:
there is kind of a critical issue which is that a general solution is hard to find given how unigram works. Once backtracking BPE is merged i'll have another look

PR #1585, opened 2024-08-01
Tiny improvement

Narsil:
Rational is as follows, before we used to hash the sequence, fail to find in cache, then lookup in vocab, hit and return.
This `ignore_merges` doesn't insert in cache (it's already in the vocab so no need to duplicate the data), we can put it beforehand limiting the amount of cache reads.

Before:
```
==============
num_threads: 8, data size: 24.04 MB, documents: 10000 Avg Length: 1659
tiktoken 	61.73 MB  / s
huggingface 	23.32 MB / s
==============
num_threads: 8, data size: 1.11 MB, documents: 10000 Avg Length: 116
tiktoken 	6.65 MB  / s
huggingface 	20.20 MB / s
```

After: 
```
==============
num_threads: 8, data size: 24.04 MB, documents: 10000 Avg Length: 1659
tiktoken 	59.51 MB  / s
huggingface 	25.36 MB / s
==============
num_threads: 8, data size: 1.11 MB, documents: 10000 Avg Length: 116
tiktoken 	7.48 MB  / s
huggingface 	20.93 MB / s
```

### DIFF
--- a/bindings/python/benches/test_tiktoken.py
+++ b/bindings/python/benches/test_tiktoken.py
@@ -25,12 +25,12 @@ def format_byte_size(num_bytes: int) -> Tuple[str, str]:
     return f"{num_bytes_f:.2f} PB", "PB"
 
 
-def benchmark_batch(model: str, documents: list[str], num_threads: int) -> None:
+def benchmark_batch(model: str, documents: list[str], num_threads: int, document_length: float) -> None:
     os.environ["RAYON_NUM_THREADS"] = str(num_threads)
     num_bytes = sum(map(len, map(str.encode, documents)))
     readable_size, unit = format_byte_size(num_bytes)
     print(f"==============")
-    print(f"num_threads: {num_threads}, data size: {readable_size}, documents: {len(documents)}")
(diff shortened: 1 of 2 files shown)


bindings/python/benches/test_tiktoken.py
```diff
@@ -82,24 +82,30 @@ def benchmark_batch(model: str, documents: list[str], num_threads: int) -> None:
...
     for num_threads in threads:
-        for length, fuse in input_lengths:
+        for length, fuse, long in input_lengths:
             documents = []
             for i, item in enumerate(dataset_xnli["train"]):
                 if i >= length:
                     break
-                documents.append("".join(item["premise"].values()))
+                if long:
+                    documents.append("".join(item["premise"].values()))
```
ArthurZucker, line 95: long is non english only, good that you added the average length as well

ArthurZucker (approved):

bindings/python/benches/test_tiktoken.py
Narsil: Yes but the principal factor affecting speed seems to be document length (I tried user arabic and had similar results)

PR #1586, opened 2024-08-01
Enable fancy regex

Narsil:
`tokenizers` enable the use of 2 regex engines.
The main supported one is `onig`, the other one is unstable and used by `unstable_wasm` and it's `fancy_regex` (onig being C cannot be compiled as is. on wasm targets).

This PR fixes the code so that it leverages `tokenizers` regexes instead of `onig` directly in the python bindings, which allows swapping one for the other.

Unfortunately `FancyRegex` is slower.

### DIFF
--- a/tokenizers/src/utils/fancy.rs
+++ b/tokenizers/src/utils/fancy.rs
@@ -1,3 +1,5 @@
+use crate::tokenizer::pattern::Pattern;
+use crate::Offsets;
 use fancy_regex::Regex;
 use std::error::Error;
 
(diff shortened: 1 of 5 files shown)


tokenizers/src/utils/fancy.rs
```diff
@@ -31,3 +33,31 @@ impl<'r, 't> Iterator for Matches<'r, 't> {
...
+            }
+            splits.push(((start, end), true));
+            prev = end;
+        }
+        if prev != inside.len() {
+            splits.push(((prev, inside.len()), false))
+        }
+        Ok(splits)
+    }
+}
```
ArthurZucker, line 63: is copied from onig?

ArthurZucker (commented):
Nice there were some complaints about compiling without onig while the doc says it's optional

ArthurZucker (approved):

tokenizers/src/utils/fancy.rs
Narsil: Indeed.

PR #1587, opened 2024-08-02
[Breaking Change] Perf improvement 16% by removing offsets.

Narsil:
Offsets calculation are always calculated in Python land. By changing it to not being calculated, we win 16% of the runtime.

This is not the total extent of it because offsets are still calculated in bytes.


tokenizers/src/tokenizer/pre_tokenizer.rs
```diff
@@ -146,6 +147,19 @@ impl PreTokenizedString {
...
                 OffsetType::Char => Some(BytesToCharOffsetConverter::new(&self.original)),
                 OffsetType::Byte => None,
+                OffsetType::None => {
+                    let tokens = self
+                        .splits
+                        .into_iter()
+                        .flat_map(|split| {
+                            split.tokens.unwrap().into_iter().map(|token| {
+                                // Replace this with the actual fields you need for the Encoding type
+                                (token.id, String::new(), (0, 0), None, 0)
```
ArthurZucker, line 157:
```suggestion
                                (token.id,String::with_capacity(0), (0, 0), None, 0)
```
for no allocation

tokenizers/src/tokenizer/mod.rs
```diff
@@ -759,6 +759,48 @@ where
         }
     }
 
+    /// Encode the given input. This method accepts both single sequences, as well as pair
+    /// sequences. Also, a sequence can be a string, or already pre-tokenized input directly:
+    ///
```
ArthurZucker, line 764:
```suggestion
    /// Contrarily to `encode`, it does not compute offsets
```

ArthurZucker (approved):

PR #1588, opened 2024-08-02
Using serde (serde_pyo3) to get __str__ and __repr__ easily.

Narsil:
Dummy implementation everywhere, but it's a showcase.


ArthurZucker (approved):
Fu**ng amazing 🚀 Instead of deriving at compile time a display function for each and every single class in rust, we just rely on the already derived `Serialize` trait, which give `obj.serialize()`. so smart, really thanks for finding a much more elegant solution thant #1542

PR #1589, opened 2024-08-03
Recursive ellipsis for serde_pyo3

EricLBuehler:
This is probably wrong because it's hard to control when we recurse?


Narsil:
This looks good I think.

Might need to tweaks but that's all.

PR #1590, opened 2024-08-04
Support `None` to reset pre_tokenizers and normalizers, and index sequences

ArthurZucker:
image


Narsil (approved):
LGTM

itazap:
Nice! works great! I am able to iter over a `normalizer.Sequence` as in the screenshot, but not over `tokenizer.Sequence`. Is it possible to please add support for that? 🙏

PR #1599, opened 2024-08-06
Improve decoder deserialization

Narsil:
Improves the actual error message when `type` is present, but fields are invalid.

### DIFF
--- a/tokenizers/src/decoders/mod.rs
+++ b/tokenizers/src/decoders/mod.rs
@@ -10,7 +10,7 @@ pub mod wordpiece;
 pub use super::pre_tokenizers::byte_level;
 pub use super::pre_tokenizers::metaspace;
 
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::decoders::bpe::BPEDecoder;
 use crate::decoders::byte_fallback::ByteFallback;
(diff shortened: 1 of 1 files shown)


tokenizers/src/decoders/mod.rs
```diff
@@ -98,7 +208,7 @@ mod tests {
         match parse {
             Err(err) => assert_eq!(
                 format!("{err}"),
-                "data did not match any variant of untagged enum DecoderWrapper"
+                "data did not match any variant of untagged enum DecoderUntagged"
```
ArthurZucker, line 211: We could change the custom to properly output decoder wrapper

ArthurZucker (approved):
Think we are losing a little bit with the custom, worth to declare outside error

tokenizers/src/decoders/mod.rs
Narsil: I think the new message is better, since `Untagged` is more descriptive than wrapper (both mean nothing to regular users, but at least untagged points to the fact that the type is missing.

We can override the message as needed. Do you have a good suggestion for the message?

ArthurZucker: Out of the box, something like "The `type` field is not present and we could not load into any of the variants of DecoderUntagged"

PR #1605, opened 2024-08-08
Fast regex

ArthurZucker:
Try the fast regex


ArthurZucker:
It was not that fast / not huge perf boost

PR #1610, opened 2024-08-09
fix benchmark file link

152334H:
it was broken


ArthurZucker (approved):
Thanks for updating! 🤗

PR #1638, opened 2024-09-29
Fix off-by-one error in tokenizer::normalizer::Range::len

rlanday:
I was reading through this code and observed an apparent bug. This pull request fixes the bug and adds test cases. Note that the case where the start bound is exclusive is not tested, as range expressions always create ranges with inclusive start bounds.

### DIFF
--- a/tokenizers/src/tokenizer/normalizer.rs
+++ b/tokenizers/src/tokenizer/normalizer.rs
@@ -45,8 +45,8 @@ where
 
         match range.start_bound() {
             Bound::Unbounded => Some(end),
-            Bound::Included(i) => Some(end - (*i + 1)),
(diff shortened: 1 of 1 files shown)


ArthurZucker (commented):
Hey! Could you rebase on main for the failing rust build. 
Otherwise, I am not entirely sure where this is used but logically this makes sense yes!

rlanday:
> Hey! Could you rebase on main for the failing rust build. Otherwise, I am not entirely sure where this is used but logically this makes sense yes!

Done

rlanday:
Hi @ArthurZucker, do I need to do anything else?

ArthurZucker (approved):
Nope sorry!

PR #1651, opened 2024-10-11
Add safety comments

Manishearth:
I recently performed safety review on 0.19. Besides finding the already-fixed https://github.com/huggingface/tokenizers/issues/1491, I also found that these cases of unsafety were a bit hard to disentangle, and documented them further.

### DIFF
--- a/tokenizers/src/models/bpe/trainer.rs
+++ b/tokenizers/src/models/bpe/trainer.rs
@@ -532,7 +532,7 @@ impl BpeTrainer {
             // Merge the new pair in every words
             // Safety: This is just a type assertion, the code below may no longer be safe
             // if the type of `pos` changes
-            let ref pos: HashSet<usize> = top.pos;
+            let pos: &HashSet<usize> = &top.pos;
 
             let words_len = words.len();
             struct WordPtr(*mut Word);
--- a/tokenizers/src/models/bpe/trainer.rs
+++ b/tokenizers/src/models/bpe/trainer.rs
@@ -534,7 +534,6 @@ impl BpeTrainer {
             // if the type of `pos` changes
             let ref pos: HashSet<usize> = top.pos;
 
-
             let words_len = words.len();
             struct WordPtr(*mut Word);
             // Safety: We do not actually use this for concurrent access to the same memory,
--- a/tokenizers/src/models/bpe/trainer.rs
+++ b/tokenizers/src/models/bpe/trainer.rs
@@ -530,13 +530,19 @@ impl BpeTrainer {
             merges.push((top.pair, new_token_id));
 
             // Merge the new pair in every words
-            let changes = top
-                .pos
+            // Safety: This is just a type assertion, the code below may no longer be safe
+            // if the type of `pos` changes
+            let ref pos: HashSet<usize> = top.pos;
+            let changes = pos
                 .maybe_par_iter()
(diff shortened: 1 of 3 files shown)


sftse:
Does the referenced UCG discussion [412](https://github.com/rust-lang/unsafe-code-guidelines/issues/412) apply to this case? Exercising the offending code under miri with and without the flags mentioned in the issue does not make a difference.
```
MIRIFLAGS='-Zmiri-env-set=TOKENIZERS_PARALLELISM=false -Zmiri-recursive-validation' cargo +nightly miri test bpe::trainer::tests::test_train
```

Manishearth:
@sftse IIRC miri doesn't currently do anything special around Vecs in this way, so it's not useful to rely on it for whether the rules apply. I did mention "this and related issues" because I know this type of thing has been the subject of discussion. For now I'm comfortable giving a sense of "this is potentially but not probably problematic"

Manishearth:
Oh! The & to &mut transmute is the problem, I see. We can use `&raw` or `addr_of` there instead. The recursive validity issue made me not notice the other issue there.

Given that I might just write it with manual unsafe (which would not be open to recusrive validity issues or transmute issues) then.

sftse:
Thanks!

Before seeing your PR I had made #1664, trying to hit the same angle but a bit more verbose, I should have mentioned that. #1651 and #1664 have a bit of overlap, the maintainers may want to deduplicate how they see fit.

tokenizers/src/models/bpe/trainer.rs
```diff
@@ -530,14 +530,30 @@ impl BpeTrainer {
...
-                    // We can merge each of these words in parallel here because each position
-                    // can be there only once (HashSet). So this is safe.
+                    // Safety:
+                    // We are producing a valid pointer since we are indexing in bounds
+                    //
+                    // We can access each `word` here in parallel because each position
+                    // can be there only once (pos is a HashSet).
                     unsafe {
+                        assert!(i < words_len);
+                        // This is words[i], but avoids needing to go through &T (which triggers UB)
```
ArthurZucker, line 555: FMI, why is `&T` UB?

ArthurZucker (approved):
Thanks a lot both! 🤗 quite honored to see core rust people here 👀
LGTM!

ArthurZucker:
`cargo fmt` should be enough for the CI!

ArthurZucker:
```
Error:    --> src/models/bpe/trainer.rs:535:17
    |
535 |             let ref pos: HashSet<usize> = top.pos;
    |             ----^^^^^^^--------------------------- help: try: `let pos: &HashSet<usize> = &top.pos;`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#toplevel_ref_arg
    = note: `-D clippy::toplevel-ref-arg` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::toplevel_ref_arg)]`
```
seems to cause an issue with compilation, not sure if it's actionable on my side

tokenizers/src/models/bpe/trainer.rs
Manishearth: It's UB to mutate data that currently has an active `&T`. This persists through raw pointers in some ways (and not in others).

ArthurZucker:
Thanks!

Issue #1612, opened 2024-08-13
`RefMutContainer` is unsound


CheaterCodes:
Since I was made aware of an unsoundness in this code base today (see #1485), I decided to have a bit of a look around.

I found that `RefMutContainer` erases lifetimes, resulting in more *unsafe* code.
Note that I didn't check if there is actually any unsoundness here. However, since the safety here can't be locally guaranteed, the methods should at least be marked as `unsafe`, but even better, fixed to be correct.

Here is an example code that shows undefined behavior when checked with Miri:
```rust
use std::sync::{Arc, Mutex};

#[derive(Clone)]
pub struct RefMutContainer<T> {
    inner: Arc<Mutex<Option<*mut T>>>,
}

impl<T> RefMutContainer<T> {
    pub fn new(content: &mut T) -> Self {
        Self {
            inner: Arc::new(Mutex::new(Some(content))),
        }
    }

    pub fn map<F: FnOnce(&T) -> U, U>(&self, f: F) -> Option<U> {
        let lock = self.inner.lock().unwrap();
        let ptr = lock.as_ref()?;
        Some(f(unsafe { ptr.as_ref().unwrap() }))
    }

    pub fn map_mut<F: FnOnce(&mut T) -> U, U>(&mut self, f: F) -> Option<U> {
        let lock = self.inner.lock().unwrap();
        let ptr = lock.as_ref()?;
        Some(f(unsafe { ptr.as_mut().unwrap() }))
    }
}

fn main() {
    let mut text = String::from("foo");
    let container = RefMutContainer::new(&mut text);
    drop(text);
    
    container.map(|text| {
      // Triggers UB
      println!("{text}");
    });
}
```

Link to playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=326897debac51f4e11f68833600f084e

Error produced by miri:
```txt
error: Undefined Behavior: out-of-bounds pointer use: alloc1256 has been freed, so this pointer is dangling
   --> /playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/raw.rs:138:9
    |
138 |         &*ptr::slice_from_raw_parts(data, len)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ out-of-bounds pointer use: alloc1256 has been freed, so this pointer is dangling
    |
[...]

Issue #1485, opened 2024-04-03
error: casting `&T` to `&mut T` is undefined behavior


Jipok:
ERROR: Failed building wheel for tokenizers:

```
Building wheels for collected packages: tokenizers
  Building wheel for tokenizers (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Building wheel for tokenizers (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [592 lines of output]
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib.linux-x86_64-cpython-312
      creating build/lib.linux-x86_64-cpython-312/tokenizers
      copying py_src/tokenizers/__init__.py -> build/lib.linux-x86_64-cpython-312/tokenizers
      creating build/lib.linux-x86_64-cpython-312/tokenizers/models
      copying py_src/tokenizers/models/__init__.py -> build/lib.linux-x86_64-cpython-312/tokenizers/models
      creating build/lib.linux-x86_64-cpython-312/tokenizers/decoders
      copying py_src/tokenizers/decoders/__init__.py -> build/lib.linux-x86_64-cpython-312/tokenizers/decoders
      creating build/lib.linux-x86_64-cpython-312/tokenizers/normalizers
      copying py_src/tokenizers/normalizers/__init__.py -> build/lib.linux-x86_64-cpython-312/tokenizers/normalizers
      creating build/lib.linux-x86_64-cpython-312/tokenizers/pre_tokenizers
      copying py_src/tokenizers/pre_tokenizers/__init__.py -> build/lib.linux-x86_64-cpython-312/tokenizers/pre_tokenizers
      creating build/lib.linux-x86_64-cpython-312/tokenizers/processors
      copying py_src/tokenizers/processors/__init__.py -> build/lib.linux-x86_64-cpython-312/tokenizers/processors
      creating build/lib.linux-x86_64-cpython-312/tokenizers/trainers
      copying py_src/tokenizers/trainers/__init__.py -> build/lib.linux-x86_64-cpython-312/tokenizers/trainers
      creating build/lib.linux-x86_64-cpython-312/tokenizers/implementations
      copying py_src/tokenizers/implementations/__init__.py -> build/lib.linux-x86_64-cpython-312/tokenizers/implementations
      copying py_src/tokenizers/implementations/base_tokenizer.py -> build/lib.linux-x86_64-cpython-312/tokenizers/implementations
[...]

PR #1664, opened 2024-10-22
Unsound call of `set_var`

sftse:
There have been multiple attempts to communicate that the pattern `&mut *(&T as *const T as *mut T)` is not sound, as in #1485 .

~~This is the most straightforward modification of the code that passes `cargo miri`.~~ See #1651 for the fix.
To detect the unsoundness, the parallelism must be turned off under miri, otherwise it falls over because of some unrelated issue in `crossbeam-epoch`:
```
MIRIFLAGS='-Zmiri-env-set=TOKENIZERS_PARALLELISM=false' cargo +nightly miri test bpe::trainer::tests::test_train
```

There ~~are two soundness issues~~is one soundness issue fix and one code cleanup in this PR.

The ~~first~~ soundness issue ~~is transforming & to &mut, the second~~ is hidden by the fact that std until recently had a safe `env::set_var` function. This was a breaking change and is as of now an unsafe function to call.
Lastly, some elided lifetimes were making it quite hard to see whether the use of `RefMutContainer` as mentioned in #1612 was sound. It does not seem to be unsound.

### DIFF
--- a/bindings/python/src/utils/pretokenization.rs
+++ b/bindings/python/src/utils/pretokenization.rs
@@ -39,7 +39,7 @@ fn normalize(pretok: &mut PreTokenizedString, func: &Bound<'_, PyAny>) -> PyResu
     } else {
         ToPyResult(pretok.normalize(|normalized| {
             let norm = PyNormalizedStringRefMut::new(normalized);
-            func.call((norm.get(),), None)?;
+            func.call((norm.get().clone(),), None)?;
             Ok(())
         }))
         .into()
--- a/bindings/python/src/utils/pretokenization.rs
+++ b/bindings/python/src/utils/pretokenization.rs
@@ -272,7 +272,7 @@ impl DestroyPtr for PyPreTokenizedStringRefMut {
 }
 
 impl PyPreTokenizedStringRefMut {
-    pub fn new(pretok: &mut tk::PreTokenizedString) -> RefMutGuard<Self> {
+    pub fn new(pretok: &mut tk::PreTokenizedString) -> RefMutGuard<'_, Self> {
         // SAFETY: This is safe because we return a RefMutGuard here.
         // The compiler will make sure the &mut stays valid as necessary.
         RefMutGuard::new(Self {
--- a/bindings/python/src/normalizers.rs
+++ b/bindings/python/src/normalizers.rs
@@ -553,7 +553,7 @@ impl tk::tokenizer::Normalizer for CustomNormalizer {
         Python::with_gil(|py| {
             let normalized = PyNormalizedStringRefMut::new(normalized);
             let py_normalized = self.inner.bind(py);
-            py_normalized.call_method("normalize", (normalized.get(),), None)?;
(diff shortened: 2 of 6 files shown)


tokenizers/src/models/bpe/trainer.rs
```diff
@@ -8,6 +8,63 @@ use serde::{Deserialize, Serialize};
...
 use std::collections::{BinaryHeap, HashMap, HashSet};
 
+mod unsafevec {
+    use std::cell::UnsafeCell;
+    use std::marker::PhantomData;
+
+    use super::Word;
+
+    pub(super) struct UnsafeVec<'v> {
+        value: UnsafeCell<*mut Word>,
```
Manishearth, line 18: I don't think this UnsafeCell is doing anything here?

Manishearth (commented):
not really convinced this abstraction pulls its weight compared to the approach in #1651 , it doesn't  really internalize any safety invariants

tokenizers/src/models/bpe/trainer.rs
sftse: I wasn't certain when I wrote this, does the borrow on self of `fn get_unchecked_mut` not affect anything?
One is not normally permitted to mutate past a &, you can't use a `&&mut T` to mutate `T`, so I suspected one cannot easily mutate `T` behind `&*mut T`. Miri seems fine with it, so I guess this UnsafeCell is not needed and you're correct.

ArthurZucker:
Thanks a lot for opening the PR and contributing either ways! 🤗

bindings/python/src/utils/pretokenization.rs
```diff
@@ -272,10 +270,10 @@ impl DestroyPtr for PyPreTokenizedStringRefMut {
 }
 
 impl PyPreTokenizedStringRefMut {
-    pub fn new(pretok: &mut tk::PreTokenizedString) -> RefMutGuard<Self> {
+    pub fn new(pretok: &mut tk::PreTokenizedString) -> super::RefMutGuard<'_, Self> {
         // SAFETY: This is safe because we return a RefMutGuard here.
         // The compiler will make sure the &mut stays valid as necessary.
-        RefMutGuard::new(Self {
+        super::RefMutGuard::new(Self {
```
ArthurZucker, line 276: any reason not to import it from super directly?

ArthurZucker (approved):
LGTM had no idea set var was unsafe, we might add this mirin check to our CI!

bindings/python/src/utils/pretokenization.rs
sftse: Commit slipped in from a different PR I was working on, it's an immaterial change, undone.

sftse: When making refactors, I find it helps the readability of changes to have diffs that are as localized as possible. If subsequent commits touch these lines, I might move the imports to be more local.

ArthurZucker (approved):
Thanks 🤗

Issue #1539, opened 2024-05-23
Memory leak for large strings


noamgai21:
This snippet will cause memory usage to rise indefinitely:

```python
from transformers import AutoTokenizer
import gc

tokenizer = AutoTokenizer.from_pretrained("TinyLlama/TinyLlama-1.1B-Chat-v1.0", use_fast=True)
refresh_every = 100000

for i in range(100000):
  s = f'{i} {i} ' * 10000
  tokenizer.encode(s)
  gc.collect()
  if i % 100 == 0:
    print(i)
  if i % refresh_every == 0:
    tokenizer = AutoTokenizer.from_pretrained("TinyLlama/TinyLlama-1.1B-Chat-v1.0", use_fast=True)
```

If you set `refresh_every` to 100000 (like it is in the snippet), the memory usage will keep on rising. [This colab notebook](https://colab.research.google.com/drive/1YDuPXviLqmwFAaROiVMCLIwPGjLTkQ_Y#scrollTo=3eIklMiRFq6J) crashes after about 15 minutes of executing.

If you set `refresh_every` to 100, the memory consumption will be stable.

noamgai21:
Related to https://github.com/huggingface/tokenizers/issues/1495

tomaarsen:
Hello!

I am also experiencing a memory leak with these tokenizers when processing long sequences without any spaces. This has been reported as a memory leak in Sentence Transformers, and affects some of my users: https://github.com/UKPLab/sentence-transformers/issues/1795

## Reproduction
```python
import random
import string
import time
import psutil
from tokenizers import Tokenizer

tokenizer = Tokenizer.from_pretrained('xlm-roberta-base')

def random_string(length: int) -> str:
    return ''.join(random.choices(string.ascii_uppercase + string.digits, k=length))

for iteration in range(99999999):
    start_t = time.time()
    tokenizer.encode_batch([random_string(12345) for _ in range(200)])
    memory_usage_in_MiB = psutil.Process().memory_info().rss / (1024 * 1024)
    delta_t = time.time() - start_t
    print(f"{iteration:02d}: {memory_usage_in_MiB:.2f}MB, {delta_t:.2f}s")
```

## Outputs
```
00: 353.12MB, 0.35s
01: 421.64MB, 0.51s
02: 492.77MB, 0.68s
03: 571.88MB, 0.93s
04: 623.66MB, 1.02s
05: 710.28MB, 1.35s
06: 803.41MB, 1.31s
[...]

PR #1676, opened 2024-11-06
Disable caching for long strings.

Narsil:
Fixes: https://github.com/huggingface/tokenizers/issues/1539


ArthurZucker (approved):
Sounds good, might be "breaking" a tad bit for performances (maybe huge system prompt case no?)

Narsil:
I don't think so, huge prompts are still usually followed by some kind of prompt. So caching never works unless it's exactly the same prompt.

I think the downside should be minimal, and it's not visibly breaking to users anyway.

PR #1678, opened 2024-11-07
Decode stream python

Narsil:
Follow up of https://github.com/huggingface/tokenizers/pull/1677

### DIFF
--- a/bindings/python/src/decoders.rs
+++ b/bindings/python/src/decoders.rs
@@ -1,6 +1,7 @@
 use std::sync::{Arc, RwLock};
 
 use crate::pre_tokenizers::from_string;
+use crate::tokenizer::PyTokenizer;
 use crate::utils::PyPattern;
 use pyo3::exceptions;
 use pyo3::prelude::*;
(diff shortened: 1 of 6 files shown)


bindings/python/src/decoders.rs
```diff
@@ -591,9 +592,71 @@ pub fn decoders(m: &Bound<'_, PyModule>) -> PyResult<()> {
...
     m.add_class::<PySequenceDecoder>()?;
+    m.add_class::<PyDecodeStream>()?;
     Ok(())
 }
 
+/// Class needed for streaming decode
+///
+#[pyclass(module = "tokenizers.decoders", name = "DecodeStream")]
+#[derive(Clone)]
+pub struct PyDecodeStream {
```
ArthurZucker, line 603: intersting, so we don't use the rust `DecodeStream` object. I am guessing it's for ownership reasons? Otherwise we needs to wrap the `PyDecodeStream {stream: DecodeStream}` with arc?

ArthurZucker (approved):
Sounds good, I am a bit curious about the rational behind not using / creating a rust DecodeStream

Narsil:
> Sounds good, I am a bit curious about the rational behind not using / creating a rust DecodeStream

Lifetimes cannot cross the boundary. `Stream<'a>` is not possible to use directly in the bindings, so reborrowing the tokenizer on every call is the only way for python.

bindings/python/src/decoders.rs
Narsil: Arc doesn't save us. The issue is with the borrow of the tokenizer's lifetime.

You can technically reborrow in every call in rust too, it's just not very "rusty".
For Python, we need to get access to the tokenizer on every call, and cloning into an Arc feels super wasteful (and breaks every update you might do on the tokenizer afterwards).

This seems innocuous enough since currently users have to hold already a reference to the tokenizer anyway.

bindings/python/src/decoders.rs
```diff
@@ -590,9 +591,71 @@ pub fn decoders(m: &Bound<'_, PyModule>) -> PyResult<()> {
...
+    /// Prefix is the second one that was already emitted to discard the part
+    /// of the text of all the ids
+    /// read is the prefix kept only for starting side effects of the prefix
+    read_index: usize,
+}
+
+#[pymethods]
+impl PyDecodeStream {
+    #[new]
+    #[pyo3(signature = (skip_special_tokens), text_signature = "(self, skip_special_tokens)")]
```
robertgshaw2-neuralmagic, line 633: @Narsil @ArthurZucker 

Thanks for this API! Im working on integrating it into VLLM.

QQ - this API is clear for `step()`. However, would it be possible to pass the prefill tokens to `new()`?

IIUC, the current API requires me to call step N times for N prefill tokens before I get into the decode phase. Is that right?

ArthurZucker: Yep we'll add if not already possible!

ArthurZucker: https://github.com/huggingface/tokenizers/pull/1725 for the record!

PR #1673, opened 2024-11-05
fix pylist

ArthurZucker:
Fixes #1672, which was probably broken by #1665

PR #1679, opened 2024-11-08
Fix encode_batch and encode_batch_fast to accept ndarrays again

diliop:
This is a follow up to my comment from [here](https://github.com/huggingface/tokenizers/pull/1665#discussion_r1833391386) which essentially reverts the `PyList` change from #1665 and `PySequence` change from #1673 with regards to the `input` arg for `encode_batch` and `encode_batch_fast`, back to `Vec<..>`. This allows passing `ndarray` together with `list` and `tuple` as `input` types. I also turned on the tests that were turned off before to make sure that this change for `encode_batch` is covered. I will follow up with adding more tests for `encode_batch_fast` but prefer to get this out sooner than later.

### DIFF
--- a/bindings/python/tests/bindings/test_tokenizer.py
+++ b/bindings/python/tests/bindings/test_tokenizer.py
@@ -152,8 +152,6 @@ class TestTokenizer:
         assert len(output) == 2
 
     def test_encode_formats(self, bert_files):
-        print("Broken by the change from std::usize::Max to usixeMax")
-        return 0
         with pytest.deprecated_call():
             tokenizer = BertWordPieceTokenizer(bert_files["vocab"])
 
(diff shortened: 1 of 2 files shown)


diliop:
@ArthurZucker mind giving this a look since I think sooner than later someone is going to complain about `encode_batch` or `encode_batch_fast` not accepting `ndarray` types 😞

ArthurZucker (approved):
LGTM would be nice to add a case with ndarrays in the test encode format 🤗 LGTM otherwise, good catch

ArthurZucker:
(Sorry about the delay we were on a company wide offsite 😅 🌴

ArthurZucker:
Could you update with clippy as well? 🤗

bindings/python/tests/bindings/test_tokenizer.py
```diff
@@ -152,8 +152,6 @@ def test_encode(self):
         assert len(output) == 2
 
     def test_encode_formats(self, bert_files):
```
diliop, line 154: @ArthurZucker there are already a set of tests covering `np.array` - did you have something else in mind?

ArthurZucker: Nope I checked and found then so approved and good to go!

PR #1685, opened 2024-11-21
Fix type notation of merges in BPE Python binding

Coqueue:
Heya, this change is to correct type notations of `merges` in Python binding of BPE as the [counterpart in Rust implementation](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/models/bpe/model.rs#L18) expects a file name string or `Vec<(String, String)>` instead of a mapping from a tuple of integers to another.

### DIFF
--- a/bindings/python/py_src/tokenizers/implementations/byte_level_bpe.py
+++ b/bindings/python/py_src/tokenizers/implementations/byte_level_bpe.py
@@ -16,7 +16,7 @@ class ByteLevelBPETokenizer(BaseTokenizer):
     def __init__(
         self,
         vocab: Optional[Union[str, Dict[str, int]]] = None,
-        merges: Optional[Union[str, Dict[Tuple[int, int], Tuple[int, int]]]] = None,
+        merges: Optional[Union[str, List[Tuple[str, str]]]] = None,
         add_prefix_space: bool = False,
         lowercase: bool = False,
         dropout: Optional[float] = None,
--- a/bindings/python/py_src/tokenizers/implementations/char_level_bpe.py
+++ b/bindings/python/py_src/tokenizers/implementations/char_level_bpe.py
@@ -25,7 +25,7 @@ class CharBPETokenizer(BaseTokenizer):
     def __init__(
         self,
         vocab: Optional[Union[str, Dict[str, int]]] = None,
-        merges: Optional[Union[str, Dict[Tuple[int, int], Tuple[int, int]]]] = None,
+        merges: Optional[Union[str, List[Tuple[str, str]]]] = None,
         unk_token: Union[str, AddedToken] = "<unk>",
         suffix: str = "</w>",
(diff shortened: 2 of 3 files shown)


Narsil:
Why do you think the annotations are incorrect ? The 2 things are unlinked and the signature here is correct.

Coqueue:
Hi Nicolas, thanks for taking a look.

>  The 2 things are unlinked and the signature here is correct.

I wonder if this is true.
Taking the Python [SentencePieceBPETokenizer](https://github.com/Coqueue/huggingface_tokenizers/blob/f0c48bd89a442819b39605ca117ecabd293bfdd7/bindings/python/py_src/tokenizers/implementations/sentencepiece_bpe.py#L26-L29) class as an example, the `merges` annotated as `Optional[Union[str, Dict[Tuple[int, int], Tuple[int, int]]]]` is only used to initialize the [`BPE`](https://github.com/Coqueue/huggingface_tokenizers/blob/f0c48bd89a442819b39605ca117ecabd293bfdd7/bindings/python/py_src/tokenizers/models/__init__.pyi#L82) class:
```python
        tokenizer = Tokenizer(BPE(vocab, merges, dropout=dropout, unk_token=unk_token, fuse_unk=fuse_unk))
``` 
where the `BPE` class is generated from the Rust implementation mentioned in the original post.

`BPE` class also [contains below comment](https://github.com/Coqueue/huggingface_tokenizers/blob/f0c48bd89a442819b39605ca117ecabd293bfdd7/bindings/python/py_src/tokenizers/models/__init__.pyi#L90):
```python
class BPE(Model):
    """
    An implementation of the BPE (Byte-Pair Encoding) algorithm

    Args:
        vocab (:obj:`Dict[str, int]`, `optional`):
            A dictionary of string keys and their ids :obj:`{"am": 0,...}`

        merges (:obj:`List[Tuple[str, str]]`, `optional`):
            A list of pairs of tokens (:obj:`Tuple[str, str]`) :obj:`[("a", "b"),...]`
```
At least this comment gives a different type annotation for `merges`.

> Why do you think the annotations are incorrect?

I encountered an issue initializing `SentencePieceBPETokenizer` with the given type annotation, but succeeded following the type annotation given in the comment of `BPE` class.

The type annotation `Dict[Tuple[int, int], Tuple[int, int]]]]` should actually be for the [`MergesMap`](https://github.com/huggingface/tokenizers/blob/f0c48b/tokenizers/src/models/bpe/model.rs#L17C10-L17C18) type in Rust as an alias of `HashMap<Pair, (u32, u32)>`. `MergesMap` is built internally in Rust implementation from the merges.


As you have closed the PR as a repo collaborator, I'll provide the above context and keep the PR in the same status as I can be wrong.

Coqueue:
@Narsil Hi Nicolas, appreciate it if you could take a look at the above reply.

Issue #1564, opened 2024-07-10
Decode regression


daulet:
This is not a recent regression, and perhaps it won't be fixed for that reason, but I thought I'd file it anyway.

I maintain Go bindings for this library, and by sheer luck I had benchmarks when I started. At some point I've noticed a [regression](https://github.com/daulet/tokenizers/issues/14) in decoding, but only now got around to investigating it. Long story short, I've bisected this repo and root caused it to [this PR](https://github.com/huggingface/tokenizers/pull/938). Below is a benchmark used to find it.

Regression details:
```
decode                  time:   [3.9277 µs 3.9409 µs 3.9558 µs]
                        change: [+241.37% +242.64% +244.06%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

While decode is pretty fast (order of microseconds), +240% slowdown is fairly big and I wonder if we can gain back that performance.

Benchmark code (`tokenizers/benches/decode_benchmark.rs`):
```
use criterion::{black_box, criterion_group, criterion_main, Criterion};
use tokenizers::tokenizer::Tokenizer;

fn decode(tokenizer:&Tokenizer, ids_slice: Vec<u32>, skip_special_tokens: bool) -> String {
    tokenizer.decode(ids_slice, skip_special_tokens).expect("failed to decode input")
}

fn criterion_benchmark(c: &mut Criterion) {
    let tokenizer = Tokenizer::from_file("./test/data/bert-base-uncased.json").expect("failed to create tokenizer");
    c.bench_function("decode", 
    |b| b.iter(
        || decode(&tokenizer, black_box([2829, 4419, 14523, 2058, 1996, 13971, 3899].to_vec()), black_box(true))));
}

criterion_group!(benches, criterion_benchmark);
criterion_main!(benches);
```

Add this to `Cargo.toml` and run with `cargo bench decode`.
```
[[bench]]
name = "decode_benchmark"
harness = false
```

The tokenizer file is copied from [here](https://huggingface.co/google-bert/bert-base-uncased/raw/main/tokenizer.json).

ArthurZucker:
Ow wow, thanks a mile for the clear report! 
Yeah 240% is not really good! Okay, I'lll investigate but in the mean time if you have a PR with a fix, would be super good!

ArthurZucker:
I'll add this to benches! 🆙

ArthurZucker:
[...]

PR #1707, opened 2024-12-28
Bpe clones

sftse:
This PR tightens up the allocations by a bit. It's unlikely there are any juicy gains here, but the changes also happen to lead to terser, more clear code, so this might be an uncontroversial win.


sftse:
It looks like this touches the codepath that lead to the slowdown in #1564, I'll look into whether this is something easy to fix.

sftse:
Rebased on main to get the new clippy fixes, should fix CI run.

ArthurZucker:
> It looks like this touches the codepath that lead to the slowdown in https://github.com/huggingface/tokenizers/issues/1564, I'll look into whether this is something easy to fix.

Hey! did you have anuy lead regarding this? 🤗

Narsil (approved):
LGTM. I don't see any performance difference, but it definitely feels more idiomatic.

PR #1710, opened 2024-12-31
Bump to hf-hub 0.4.0

TD-Sky:
I would be appreciated if you could release a new version after merging this PR.


ArthurZucker (commented):
Thanks, will merge #1732 as it adds a small feature, but thanks for contributing! 🤗

PR #1712, opened 2025-01-03
Draft backtrack

ArthurZucker:
The goal of the PR is to allow anyone to use a `BPE` serialized model into a `BacktrackingBpe` class. This way you only have to use the correct class to leverage the new method. 

Serialization should be of `BacktrackingBpe`  type if we want to save time on initializing the double aho corasicks. TBH I am not sure it's worth it, so we'll see. 

Full credits to https://github.com/github/rust-gems


ArthurZucker:
TODOs: 
- use a trait for BPETrainer to run on BacktrackingBpe
- implement a converter directly, instead of the deserialization I am currently doing 
- implement `model.use_backtracking_bpe()` for `BPE` that just returns a converted BacktrackingBpe

Issue #1697, opened 2024-12-10
Request for pre-tokenizer that creates words based length alone.


filbeofITK:
Hello! I would like to request a fast pre-tokenizer to be implemented, which only splits the input to continuous pre-defined length segments. I know that this is not a common issue in NLP, but for my use-case it is necessary. I'm trying to process DNA data and that has no spaces or any type of separators, so I'm trying to use fixed length tokens.

Implementing this for someone that actually knows Rust and the backend would probably take less than half an hour but I don't want to learn a new language for this. 

Biggest thanks!

jonvet:
hey @filbeofITK you're right this is not common in NLP. but is it common in the biomedical domain? are there any papers where other people use something like this? if it is we might be able to merge this PR.
otherwise, if this is more of a unique research effort, then you might be better off implementing this pre tokenizer yourself in python or checkout the feature branch in the linked PR.

filbeofITK:
Well the [DNA Transformer](https://www.nature.com/articles/s41592-024-02523-z) uses this technique, and naturally my research would build on this. I have a custom tokenizer in Python already, but unfortunately it's rather slow.

jonvet:
the paper also talks about overlapping chunks, is that required/useful?

PR #1713, opened 2025-01-05
Fixed Length Pre-Tokenizer

jonvet:
Introduces a pre-tokenizer to split text in fixed length chunks (closes https://github.com/huggingface/tokenizers/issues/1697).
The method `pre_tokenize` could be more made more concise by creating a vector with indices first like so

```let char_positions: Vec<_> = text.char_indices().collect();
    let mut splits = Vec::new();

    for chunk in char_positions.chunks(self.length) {
        let start = chunk.first().map(|(i, _)| *i).unwrap_or(0);
        let end = chunk.last().map(|(i, c)| i + c.len_utf8()).unwrap_or(text.len());
        splits.push(normalized.slice(Range::Normalized(start..end))
            .ok_or("Failed to slice normalized text")?);
    }
```

 but that would take a bit more memory, so I went for my approach instead.


Narsil:
Thanks for this. The code looks working, but I think it could be simplified quite a lot.

Is there any source/paper for trying to do fixed sized chunking ? Before adding anything to the library usually we try to make sure it's used in the wild and would benefits actual users of models (not necessarily researchers exploring new ideas, for this they can try out your branch or create their own pre_tokenizer directly in Python).

jonvet:
You're right, I simplified it along the lines of my initial comment.

I also asked the author of the issue whether this is a common approach in the literature or not (I'm not aware of it either). Should have probably clarified this before jumping on it ;)

jonvet:
according to the author it's used in [DNA Transformer](https://www.nature.com/articles/s41592-024-02523-z)

bindings/python/tests/bindings/test_pre_tokenizers.py
```diff
@@ -195,6 +196,21 @@ def test_can_modify(self):
...
+        assert isinstance(FixedLength(), FixedLength)
+        assert isinstance(pickle.loads(pickle.dumps(FixedLength())), FixedLength)
+
+    def test_can_modify(self):
+        pretok = FixedLength(length=5)
+        assert pretok.length == 5
+
+        pretok.length = 10
+        assert pretok.length == 10
+
```
ArthurZucker, line 212: we'd also want to make sure that it does it's job as a pretokenizer! so testing with the same string, that it splits in 5 then 10!

ArthurZucker (commented):
Same as my colleague! would be nice if we can get a reference to the paper it was used in the documentation of the class! (like a arxiv link) 
Otherwise we can also keep this issue open an let the community upvote ! if it get's traction we merge 🤗

ArthurZucker:
I'll fix the CI and merge this, sorry for being slow!

ArthurZucker (approved):

PR #1733, opened 2025-02-10
Add FxHash and ShortStringOptimization.

MeetThePatel:
This PR contains performance optimizations (spreadsheet of benchmark comparisons linked below). The two main optimizations are:

- Switching to FxHash instead of standard library default hasher. This is provided by the `rustc_hash` crate, which the compiler uses internally.

- Switch from `String` to `CompactString`, provided by the `compact_str` crate, which has short-string-optimization for up to 24 bytes, which many tokens can fit inside of.

# Progress:
This PR is not fully complete. At this point in time, the following tasks have been completed:
- [X] Convert base crate.
  - Tests are passing, and benchmarks were run (linked below).
- [X] Convert Python bindings.
  - The same tests that pass on HEAD are passing on this branch. There seems to be 2 failing tests on HEAD, but this branch fails the exact same tests, and in the same manner (i.e. same output).
- [ ] Convert Node bindings.
- [ ] Cleanup.

# Benchmarks:
The benchmarks are at the following link: [Tokenizer Benchmarks](https://docs.google.com/spreadsheets/d/14_nPSj6sZVBaJ8zLbCp-gekYOC8fYUZDjB5a3uOA0TU/edit?usp=sharing). All benchmarks were run on a Macbook Pro with the M2 Pro chip and 16GB memory.

Additionally, the `BPE Train vocabulary (huge)` benchmark is one that I added (not committed, as that would warrant its own PR). This benchmark is using the dataset: [One Billion Word Challenge](https://www.statmt.org/lm-benchmark/), which clocks in at 4.15GB.

### DIFF
--- a/bindings/node/src/models.rs
+++ b/bindings/node/src/models.rs
@@ -4,7 +4,7 @@ use crate::trainers::Trainer;
 use napi::bindgen_prelude::*;
 use napi_derive::napi;
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use rustc_hash::FxHashMap;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, RwLock};
 use tokenizers as tk;
@@ -95,7 +95,7 @@ impl tk::Model for Model {
     self.model.as_ref()?.read().unwrap().id_to_token(id)
   }
 
-  fn get_vocab(&self) -> HashMap<String, u32> {
+  fn get_vocab(&self) -> FxHashMap<String, u32> {
     self
       .model
       .as_ref()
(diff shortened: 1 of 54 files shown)


ArthurZucker:
Sounds good! Node bidings are really not necessary!

ArthurZucker:
Do you want this to be reviewed now?

MeetThePatel:
Not yet. I'm still working on a few things:
- See if there actually is a bottleneck in the Python-Rust interface (regarding extra type conversions), or if the benchmarks aren't pushing the crate hard enough to see a meaningful difference. It should be ~10-15% faster (according to the base crate BPE encode benchmark).
- Clean up code quality.

MeetThePatel:
I think this is ready for review.

Also, these are the distributions for the benchmark runs (blue is this PR, red is HEAD). Besides the speed being a bit faster, it seems to be more consistent as well (as least on my machine).

[HEAD vs PR base crate benchmarks.pdf](https://github.com/user-attachments/files/18759687/HEAD.vs.PR.base.crate.benchmarks.pdf)

PR #1737, opened 2025-02-19
Use ApiBuilder::from_env() in from_pretrained function

BenLocal:
Use ApiBuilder::from_env() for builder initialization, can download a file by modifying the environment variable HF_ENDPOINT


ArthurZucker (approved):
Should superseed #1778  yours was open earlier! THanks both

PR #1756, opened 2025-03-25
Itertools upgrade

sftse:


sftse:
This might not be apparent from this PR, but there currently exist two different `itertools` versions in the transitive dependencies of `tokenizers`. This upgrade removes one of them and ensures they are both on the most recent version.

Narsil (approved):

PR #1758, opened 2025-03-31
Implement `from_bytes` and `read_bytes` Methods in WordPiece Tokenizer for WebAssembly Compatibility

sondalex:
**Description:**  

This Pull Request introduces the implementation of `from_bytes` and `read_bytes` methods within the WordPiece tokenizer. These methods were added to improve integration with WebAssembly (Wasm). As model downloading is managed on the JavaScript side in a Wasm target environment, methods compatible with `u8` types facilitates the transfer of a `Uint8Array` from JavaScript to Rust.

### DIFF
--- a/tokenizers/src/models/wordpiece/mod.rs
+++ b/tokenizers/src/models/wordpiece/mod.rs
@@ -182,9 +182,10 @@ impl WordPiece {
 
         Ok(vocab)
     }
-    pub fn from_bytes(vocab: &[u8]) -> Result<WordPieceBuilder> {
-        let vocab = WordPiece::read_bytes(vocab)?;
-        Ok(WordPiece::builder().vocab(vocab))
+
+    pub fn from_bytes<P: AsRef<[u8]>>(bytes: P) -> Result<Self> {
+        let tokenizer = serde_json::from_slice(bytes.as_ref())?;
+        Ok(tokenizer)
     }
 
     /// Initialize a `WordPiece` model from a vocab mapping file.
--- a/tokenizers/src/models/wordpiece/mod.rs
+++ b/tokenizers/src/models/wordpiece/mod.rs
@@ -171,6 +171,22 @@ impl WordPiece {
         Ok(vocab)
     }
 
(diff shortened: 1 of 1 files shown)


tokenizers/src/models/wordpiece/mod.rs
```diff
@@ -171,6 +171,22 @@ impl WordPiece {
...
+            let line = line?;
+            vocab.insert(line.trim_end().to_owned(), index as u32);
+        }
+
+        Ok(vocab)
+    }
+    pub fn from_bytes(vocab: &[u8]) -> Result<WordPieceBuilder> {
+        let vocab = WordPiece::read_bytes(vocab)?;
+        Ok(WordPiece::builder().vocab(vocab))
+    }
```
ArthurZucker, line 188: why don't we use something like:
```
    pub fn from_bytes<P: AsRef<[u8]>>(bytes: P) -> Result<Self> {
        let tokenizer = serde_json::from_slice(bytes.as_ref())?;
        Ok(tokenizer)
    }
```
I trust serde a lot!

sondalex: Hello @ArthurZucker , thank you for the suggestion. 

Sure, let's do this!

tokenizers/src/models/wordpiece/mod.rs
```diff
@@ -171,6 +171,23 @@ impl WordPiece {
...
+    pub fn read_bytes(vocab: &[u8]) -> Result<Vocab> {
+        let file = BufReader::new(vocab);
+
+        let mut vocab = HashMap::new();
+        for (index, line) in file.lines().enumerate() {
+            let line = line?;
+            vocab.insert(line.trim_end().to_owned(), index as u32);
+        }
+
+        Ok(vocab)
```
ArthurZucker, line 183: this currently breaks on main! We switch to using AHashmap, I suspect the branch was not up to date

ArthurZucker: will fix

PR #1763, opened 2025-04-18
Make unigram cache optional

wangrunji0408:
We observed that the Unigram model's cache introduces significant contention overhead under multi-threaded scenarios. To mitigate this, we made the Unigram cache optional following BPE model's approach. Users can now disable it via `resize_cache(0)`. Additionally, we added `TokenizerImpl::get_model_mut` to obtain a mutable reference of model.


tokenizers/src/tokenizer/mod.rs
```diff
@@ -602,6 +602,11 @@ where
         &self.model
     }
 
+    /// Get a mutable reference to the model
+    pub fn get_model_mut(&mut self) -> &mut M {
+        &mut self.model
+    }
+
     /// Set the added vocabulary.
```
ArthurZucker, line 610: as this is unused here I would not put it in this PR!

ArthurZucker (commented):
Sorry for my late review! happy to merge, can you also update the `python` bindings to expose this in `python` as well! Can come in handy!

ArthurZucker (commented):
Ah actually I am not sure I understand the point here, you can just already resize to 0 should not make. a difference

PR #1764, opened 2025-04-18
Update __init__.pyi: fix 525: SyntaxWarning: invalid escape sequence '\w'

wyattscarpenter:
Using a raw string.


wyattscarpenter:
Thanks. So it's something to do with the transformation from a rust docstring.

```rust
/// This pre-tokenizer simply splits using the following regex: `\w+|[^\w\s]+`
```

This is a valid rust docstring, but whatever turns it into a python docstring needs to be updated to respect the python escape rules. (Or, maybe, it wants to let you use python escapes, and so we must correct this docstring to escape the escapes? Hmm...)

wyattscarpenter:
I don't really know how to deal with that myself, offhand, so I guess I'll just pivot to opening an issue about it instead.

PR #1771, opened 2025-05-01
Upgrade onig, to get it compiling with GCC 15

414owen:
Distributions are starting to ship with GCC 15, which won't build `onig`. We can upgrade `onig` to an unreleased version, which fixes the issue. We should make a mental note to upgrade to a stable point release, when one is available (although at the moment, the onig repository is archived, so I'm not sure when that will happen).

### DIFF
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -43,7 +43,7 @@ harness = false
 
 [dependencies]
 rand = "0.8"
-onig = { version = "6.4", default-features = false, optional = true }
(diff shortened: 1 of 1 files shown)


ArthurZucker (commented):
I'd be in favor of waiting a bit as this means we are introducing a non-create.io deps, preventing from publishing there AFAIK

danieldk:
Seems like there were new releases since this PR: https://github.com/rust-onig/rust-onig/releases

So an update to onig 6.5 should fix this.

ArthurZucker:
Thanks @414owen 🤗

Issue #1729, opened 2025-01-31
Building without `onig` feature fails


414owen:
`cargo build --no-default-features` fails with `unresolved import onig`.
If this isn't an optional feature, let's not mark it as such.

DCNick3:
I was able to build it by enabling `unstable_wasm` feature. Would be nice if just disabling `onig` and enabling `fancy-regex` would be enough

xangelix:
This is especially important now that `onig` doesn't build on arch and fedora systems.

https://github.com/rust-onig/rust-onig/issues/196

414owen:
Agreed.

I'm on fedora, the workaround I've been using is to use `export CC=gcc-14`, when compiling anything that depends on `tokenizers`.

414owen:
https://github.com/huggingface/tokenizers/pull/1771

Issue #1730, opened 2025-01-31
Slow compile times


414owen:
Compiling `tokenizers` takes around 35 seconds in release mode.
Most of this (25s) is spend compiling `onig_sys`.

An alternative to using `onig_sys` would be appreciated. This crate already uses the `regex-*` ecosystem, so if that could be used, it would be a big win for compile times and binary sizes.

Otherwise, if you point me to where other regex libraries fall down, I'm happy to look into alternatives.

PR #1772, opened 2025-05-01
Fix no-onig no-wasm builds

414owen:
Now, you can build with:

```
--no-default-features --features=fancy-regex
```

Which previously didn't work. You had to enable the `unstable_wasm` flag.

I think using `fancy_regex` without wasm is a valid use-case, as I've seen extremely slow build times using `onig`.
See: https://github.com/huggingface/tokenizers/issues/1730

Onig also breaks, sometimes, with compiler updates. See: https://github.com/huggingface/tokenizers/pull/1771

Closes https://github.com/huggingface/tokenizers/issues/1729

### DIFF
--- a/tokenizers/src/utils/mod.rs
+++ b/tokenizers/src/utils/mod.rs
@@ -2,14 +2,20 @@ pub(crate) mod cache;
 #[cfg(feature = "http")]
 pub(crate) mod from_pretrained;
 
-#[cfg(feature = "unstable_wasm")]
+#[cfg(feature = "fancy-regex")]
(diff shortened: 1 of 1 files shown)


adamreichold:
Considering the archiving of the upstream Oniguruma repository and its build problems using GCC 15, I wonder whether the better course of action here isn't to just unconditionally depend on fancy-regex further simplifying the whole setup.

ArthurZucker:
This breaks `cargo build --all-targets --all-features` I'll fix it to have both

Issue #1767, opened 2025-04-21
Support free-threaded Python and ship 3.13t wheels


ngoldbaum:
It looks like when I build `tokenizers` on a free-threaded Python, it chokes because of a dependency on `rust-numpy` 0.23.

If I manually update the dependencies on `pyo3` and `rust-numpy`:

```diff
diff --git a/bindings/python/Cargo.toml b/bindings/python/Cargo.toml
index 6e8b0c34..3dfdf2ca 100644
--- a/bindings/python/Cargo.toml
+++ b/bindings/python/Cargo.toml
@@ -14,8 +14,8 @@ serde = { version = "1.0", features = ["rc", "derive"] }
 serde_json = "1.0"
 libc = "0.2"
 env_logger = "0.11"
-pyo3 = { version = "0.23", features = ["abi3", "abi3-py39", "py-clone"] }
-numpy = "0.23"
+pyo3 = { version = "0.24", features = ["abi3", "abi3-py39", "py-clone"] }
+numpy = "0.24"
 ndarray = "0.16"
 itertools = "0.12"
 
@@ -24,7 +24,7 @@ path = "../../tokenizers"
 
 [dev-dependencies]
 tempfile = "3.10"
-pyo3 = { version = "0.23", features = ["auto-initialize"] }
+pyo3 = { version = "0.24", features = ["auto-initialize"] }
 
 [features]
 defaut = ["pyo3/extension-module"]
```

Then everything builds. I didn't try running the tests or anything else beyond that.

I'm a PyO3 maintainer and helped initially add support for the free-threaded build in PyO3. Happy to help out with adding support here.

0xDEADFED5:
thanks for the details @ngoldbaum! i just came here to figure out what i needed to do for python 3.13t

ArthurZucker:
Happy to update, and sorry did not have the time to do it yet! Will have a look

Qubitium:
@ngoldbaum  @ArthurZucker Env is Ubuntu 24.04 + python3.13t

Even with latest pyo3 `0.24.2` I cannot compile `tokenizers` on my local `python3.13t` env. Please help since I need this pkg to actually test transformers with free threading. 

```bash
(vm313t) root@gpu-base:~/tokenizers/bindings/python# pip install . -v
Using pip 25.1 from /root/vm313t/lib/python3.13t/site-packages/pip (python 3.13)
Processing /root/tokenizers/bindings/python
  Running command pip subprocess to install build dependencies
[...]

PR #1774, opened 2025-05-05
Update pyo3 and rust-numpy depends for no-gil/free-threading compat

Qubitium:
Fix: https://github.com/huggingface/tokenizers/issues/1767

Update `pyo3` and `rust-numpy` versions (depends) so `binding/python` can build for `python3.13t` no-gil edition.


ArthurZucker (approved):
Good idea! I think numpy 0.23 was already latest at the time so happy to update!

ArthurZucker:
Not 100% sure why we need to ping this specifc one, but fine by me

PR #1778, opened 2025-05-21
Fix ApiBuilder in from_pretrained to use env variables

vdebergue:
## Bug

Using `Tokenizer::from_pretrained` does not download the tokenizer into the `HF_HOME` folder.
Instead it will download it in the user home cache directory.

## Fix

Use env variable to load the ApiBuilder so it will read `HF_HOME` variable. 
Currently they are ignored and the default cache location is used `$HOME/.cache/huggingface`

### DIFF
--- a/tokenizers/src/utils/from_pretrained.rs
+++ b/tokenizers/src/utils/from_pretrained.rs
@@ -59,7 +59,7 @@ pub fn from_pretrained<S: AsRef<str>>(
         .into());
     }
 
-    let mut builder = ApiBuilder::new();
+    let mut builder = ApiBuilder::from_env();
     if let Some(token) = params.token {
         builder = builder.with_token(Some(token));
     }


ArthurZucker (approved):
Not super familiar with API builder, if no evn var is defined I am guessing this is equivalent right? It won't just error out instead?

vdebergue:
yes, without env it will fallback to the default and will be equivalent to `ApiBuilder::new()`

tokenizers/src/utils/from_pretrained.rs
```diff
@@ -59,7 +59,7 @@ pub fn from_pretrained<S: AsRef<str>>(
         .into());
     }
 
-    let mut builder = ApiBuilder::new();
+    let mut builder = ApiBuilder::from_env();
```
ArthurZucker, line 62:
```suggestion
    let mut builder = ApiBuilder::from_env("HF_HOME");
```
should we not set that?

ArthurZucker: Oh shit okay, I was looking at https://docs.rs/env_logger/latest/env_logger/struct.Builder.html#method.from_env, thanks for the heads up

ArthurZucker:
Can you run cargo clippy!

ArthurZucker:
Sorry found #1737 which was opened sooner!

vdebergue:
closing then, thanks !

PR #1785, opened 2025-05-27
[docs] Whitespace

stevhliu:
Improves documentation for the `Whitespace` function (see https://github.com/huggingface/transformers/issues/38180 for more details)

cc @itazap

### DIFF
--- a/bindings/python/py_src/tokenizers/pre_tokenizers/__init__.pyi
+++ b/bindings/python/py_src/tokenizers/pre_tokenizers/__init__.pyi
@@ -528,30 +528,29 @@ class Whitespace(PreTokenizer):
     whitespaces (punctuation such as hyphens, apostrophes, commas, etc.).
 
     Example:
+        Use the `Whitespace` function as shown below::
 
-        Use the `Whitespace` function as shown below:
+            ```python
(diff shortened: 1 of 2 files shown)


ArthurZucker (approved):
Nice thanks 🤗

itazap:
Awesome! 💯

stevhliu:
Hmm, having a bit of trouble here? 😅 

- I had to make some changes because running `python stub.py` gives the error `python: can't open file '/Users/stevhliu/tokenizers/stub.py': [Errno 2] No such file or directory`.
- Instead, I run the command `python bindings/python/stub.py` and it creates a new `py_src/tokenizers/` directory (with a bunch of empty files) instead of writing the changes to the existing one.
- So I modify `stub.py` to write the changes to `"bindings/python/py_src/tokenizers/"` but then it just overwrites and replaces everything in those files with a blank line which doesn't seem right.

Narsil:
I'll merge this a fix in a follow-up stub.py is right even if it's just a newline.

Narsil:
Oh you may have forgotten to rebuild tokenizers, stub.py looks at the binary and extracts the pyi from the built binary, so if your binary is outdated you may not see the new strings.

PR #1788, opened 2025-06-03
add group capture to replace

### LINKED ISSUE #1760: normalizers.Replace able to support regex group capture
Is it possible to have the Replace function support the group capture in the replacement string ?

In the following dummy example, I want to add a space between letters l and e. It works with the re package but not with normalizers.

```python
import re
from tokenizers import normalizers, Regex

pattern = r"(l)(e)"
replacement = r"\1 \2"

text = "le travail est totalement pénible"

text1 = normalizers.Replace(Regex(pattern), replacement).normalize_str(text)
text2 = re.sub(pattern, replacement, text)

print(f"{text  = }")
print(f"{text1 = }")
print(f"{text2 = }")
```
execution result :
```
text  = 'le travail est totalement pénible'
text1 = '\\1 \\2 travail est tota\\1 \\2ment pénib\\1 \\2'
text2 = 'l e travail est total ement pénibl e'
```

cboseak:
Fixes #1760

### Description
Per issue #1760's request, this is a feature to have Replace function support the group capture in the replacement string

---
### Important Notes
* In Rust's regex implementation, the replacement string syntax is custom and not identical to Perl or other languages that use backslash (\1, \2, etc.) for backreferences in replacement strings. Instead `$1 $2` is valid for a group replace string.
  * \1 replace syntax was converted to $1 syntax to allow that   

---
### Testing
* added tests
* all tests passing


ArthurZucker (commented):
Actually down for this! Do you mind updating the bindings as well?

PR #1798, opened 2025-06-13
Fix features blending into a paragraph

bionicles:
i was wondering why the main example didn't work, required the "http" feature, but i had glanced at the features list and just seen "progressbar" and didn't notice the "http" feature existed because it began at the end of a line here

this minor docs change puts these features in an unordered list with newlines between them

image


Narsil (approved):
LGTM.

Thanks for this.

PR #1805, opened 2025-06-19
Fix: Refactor BpeTrainer to use chunked pair counting

MeryylleA:
This commit refactors the `BpeTrainer` to address an Out-of-Memory (OOM) bug that occurred when training on large-alphabet corpora.

The `count_pairs` function previously loaded all unique token pairs and their counts into a single `AHashMap`, leading to excessive memory usage with large alphabets.

This fix introduces a new `count_pairs_chunked` function that processes words in chunks, significantly reducing memory consumption. The `do_train` function has been updated to use this new chunked approach.

The library compiles successfully with these changes, and this new strategy should prevent OOM errors during training on large datasets.

### DIFF
--- a/tokenizers/src/models/bpe/trainer.rs
+++ b/tokenizers/src/models/bpe/trainer.rs
@@ -447,7 +447,7 @@ impl BpeTrainer {
         // 4. Count pairs in words
         //
         self.update_progress(&progress, words.len(), "Count pairs");
-        let (mut pair_counts, mut where_to_update) = self.count_pairs(&words, &counts, &progress);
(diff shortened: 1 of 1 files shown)


ArthurZucker:
sorry did not have time to go though it, did you end up finding a fix ?

MeryylleA:
@ArthurZucker Hi Arthur, this commit was actually a test involving Google Jules in a new version. I'm very sorry for having opened this PR without definitively correcting the problem. This PR is not complete, so I closed it. Sorry for the delay in responding and for any inconvenience.

ArthurZucker:
No worries 🤗

PR #1809, opened 2025-06-23
Add 3.13t CI using pytest-run-parallel

ngoldbaum:
This PR updates the tests to be runnable in a thread pool under pytest-run-parallel. It then makes use of pytest-run-parallel to do multithreaded testing on the free-threaded build in CI.

The updates to the tests are mostly to avoid sharing filesystem paths between test runners. I also marked all the multiprocessing tests as thread-unsafe because they rely on `fork` and they rely on mutating `os.environ`.

To avoid running into runtime borrow-checker errors, I wrapped the internal state of the `PyEncoder` pyclass in a `Mutex`. I don't think it's possible for any panics to happen while mutexes are locked that weren't already possible and in all cases I think it's correct to immediately propagate panics with `unlock()` rather than trying to handle poisoned mutexes.

### DIFF
--- a/bindings/python/Makefile
+++ b/bindings/python/Makefile
@@ -21,8 +21,13 @@ TESTS_RESOURCES = $(DATA_DIR)/small.txt $(DATA_DIR)/roberta.json
 
 # Launch the test suite
 test: $(TESTS_RESOURCES)
+	pip install pytest requests setuptools_rust numpy pyarrow datasets
+	python -m pytest -s -v tests
+	cargo test --no-default-features
+
+nogil-test: $(TESTS_RESOURCES)
 	pip install pytest requests setuptools_rust numpy pyarrow datasets pytest-run-parallel
-	python -m pytest -s -v tests --parallel-threads=4
+	PYTHON_GIL=0 python -m pytest -s -v tests --parallel-threads=4
 	cargo test --no-default-features
 
 $(DATA_DIR)/big.txt :
--- a/bindings/python/Makefile
+++ b/bindings/python/Makefile
@@ -21,8 +21,8 @@ TESTS_RESOURCES = $(DATA_DIR)/small.txt $(DATA_DIR)/roberta.json
 
 # Launch the test suite
 test: $(TESTS_RESOURCES)
-	pip install pytest requests setuptools_rust numpy pyarrow datasets
-	python -m pytest -s -v tests
+	pip install pytest requests setuptools_rust numpy pyarrow datasets pytest-run-parallel
+	python -m pytest -s -v tests --parallel-threads=4
 	cargo test --no-default-features
 
 $(DATA_DIR)/big.txt :
--- a/bindings/python/conftest.py
+++ b/bindings/python/conftest.py
@@ -1,13 +1,25 @@
 import pytest
 
 
+try:
+    import pytest_run_parallel  # noqa:F401
+    PARALLEL_RUN_AVAILABLE = True
+except ImportError:
+    PARALLEL_RUN_AVAILABLE = False
+
 def pytest_addoption(parser):
     parser.addoption("--runslow", action="store_true", default=False, help="run slow tests")
 
 
 def pytest_configure(config):
     config.addinivalue_line("markers", "slow: mark test as slow to run")
-
+    # Don't warn for `pytest-run-parallel` markers if they're not available
+    breakpoint()
+    if not PARALLEL_RUN_AVAILABLE:
+        config.addinivalue_line(
+            "markers",
+            "thread_unsafe: mark the test function as single-threaded",
+        )
 
 def pytest_collection_modifyitems(config, items):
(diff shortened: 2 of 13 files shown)


bindings/python/conftest.py
```diff
@@ -1,13 +1,25 @@
...
+
 def pytest_addoption(parser):
     parser.addoption("--runslow", action="store_true", default=False, help="run slow tests")
 
 
 def pytest_configure(config):
     config.addinivalue_line("markers", "slow: mark test as slow to run")
-
+    # Don't warn for `pytest-run-parallel` markers if they're not available
+    breakpoint()
```
ngoldbaum:
```suggestion
```

Oops! Just reading this over a couple week later...

ngoldbaum:
HI all, anything I can do here to move this along? I'll go ahead and rebase in case that helps.

ArthurZucker:
Sorry for the delay!

ngoldbaum:
> Sorry for the delay!

No worries. I might look at adding 3.14t as well in the meantime...

WSL0809:
any plans to add support 3.14t ?

ngoldbaum:
> any plans to add support 3.14t ?

That would probably have to wait on at least pytorch and/or jax adding 3.14t support.

ArthurZucker (commented):
> be runnable in a thread pool under pytest-run-parallel.
I don't know if this is a strong enough motivations but happy to work towards thread free! I am quite surprised that we only need the mutex on the encodings, which are kinda the output, after everything was run !

ngoldbaum:
> That would probably have to wait on at least pytorch and/or jax adding 3.14t support.

Argh, sorry, I thought I was commenting on the `safetensors` repo...

That said, installing on 3.14t is still broken. It looks like `hf-xet` still needs to do a release that bumps the pyo3 version to prevent a build error during the tokenizers install. If you disable the `hf-xet` dependency then everything seems to build fine.

> I don't know if this is a strong enough motivations but happy to work towards thread free!

IMO it's important to have at least some multithreaded test coverage over a good chunk of the API while adding free-threaded support. pytest-run-parallel is an imperfect but automated way to do that.

You can do things like set up a Python and/or Rust build with thread sanitizer instrumentation to get finer-grained thread safety checking. Unfortunately it's not as straightfoward to use TSAN with rust code as it is for a pure C/C++ extension because the Rust toolchain has its own vendored LLVM toolchain.

> I am quite surprised that we only need the mutex on the encodings, which are kinda the output, after everything was run !

I don't think I tried to intentionally break tokenizers like that, I'm only adding mutexes in spots that I triggered runtime borrow-checker errors when I did the pytest-run-parallel mulithreaded testing exercise. Looking closer at the other pyclasses, I bet there are lots more opportunities to trigger issues.

One thing we can do is try to construct some multithreaded tests that look like real workflows where you expect users to use Python threads. We can also try to write some stress tests if there are any scenarios you're worried about where multithreaded mutation of shared state would possibly break things or where you want tokenizers to be robust.

Do you have any scenarios like that in mind?

ngoldbaum:
I'll try to find a quick contribution elsewhere to make so that CI runs in this PR automatically 🙃

ngoldbaum:
[4971dc2](https://github.com/huggingface/tokenizers/pull/1809/commits/4971dc2c75e328692b8f37a53b233a08ab416d2a) does a little bit more of an extensive change on top of what was already here and marks some of the pyclasses as frozen. See https://pyo3.rs/v0.25.1/class.htmlfrozen-classes-opting-out-of-interior-mutability for more about this.

ngoldbaum:
I opened https://github.com/huggingface/tokenizers/pull/1860 to see what the appetite is to make a bigger change toward using frozen pyclasses.

bindings/python/Makefile
```diff
@@ -25,6 +25,11 @@ test: $(TESTS_RESOURCES)
 	python -m pytest -s -v tests
 	cargo test --no-default-features
 
+nogil-test: $(TESTS_RESOURCES)
+	pip install pytest requests setuptools_rust numpy pyarrow datasets pytest-run-parallel
+	PYTHON_GIL=0 python -m pytest -s -v tests --parallel-threads=4
+	cargo test --no-default-features
+
```
Narsil, line 32: General comment: I'm not a huge fan of treating freethreaded any different than any other python implementations. The promise of freethreaded is that it should just enable more use cases. 

If we have "bugs" in the current test pipeline because we don't address parallelism properly (like with environment variables for instance) then we should fix the tests instead imho.

The more we have the same setup for both environment to tests the more likely we don't introduce bugs in either without knowing.

ngoldbaum:
See https://github.com/huggingface/tokenizers/pull/1860#issuecomment-3254352153

I'm going to focus on documentation and multithreaded tests for all builds, then add free-threaded support in a second pass that just adds new CI builds and sets `gil_used=false` on the pymodules.

PR #1830, opened 2025-07-22
Add enforce_utf8_boundaries option to BpeTrainer

sanderland:
This PR introduces a new boolean option, enforce_utf8_boundaries, to the BpeTrainer. In [recent work](https://arxiv.org/abs/2505.24689)  we've shown that preventing BPE merges which cross UTF-8 character boundaries leads to higher-quality tokenizers.
Although we have provided our [own implementation](https://github.com/sanderland/script_bpe), at ICML TokShop several people suggested that the community would benefit a lot more if this was in 🤗 tokenizers.

The flag ensures that every token learned corresponds to either a full character sequence or a valid, contiguous byte sequence prefix within a single character, leading to:
*    Improved Token Quality: It prevents the creation of unintuitive and potentially harmful tokens that mix partial and complete characters (e.g., `<0x95>\n\n` as seen in GPT-4o). 
*    Better Compression: Counter-intuitively, this constraint can also lead to a slight improvement in the final compression rate, as the model is guided toward more semantically meaningful merges.

To maintain backward compatibility, this option is disabled by default.

Unfortunately the option requires some interactions with the pretokenizer which does the encoding. I've tried to keep this clean, but open to suggestions, especially as my skills in the language are a bit rusty.

## TODO
* Node bindings. Will see if I can add this after initial review, or get help on it.

### DIFF
--- a/tokenizers/src/models/bpe/trainer.rs
+++ b/tokenizers/src/models/bpe/trainer.rs
@@ -4,14 +4,13 @@ use super::{Pair, WithFirstLastIterator, Word, BPE};
 use crate::parallelism::*;
 use crate::tokenizer::{AddedToken, Result, Trainer};
 use crate::utils::progress::{ProgressBar, ProgressStyle};
-use crate::pre_tokenizers::byte_level::bytes_char;
+use crate::pre_tokenizers::byte_level::CHAR_BYTES;
 use ahash::{AHashMap, AHashSet};
 use compact_str::CompactString;
 use dary_heap::OctonaryHeap;
 use serde::{Deserialize, Serialize};
 use std::cmp::Ordering;
 use std::collections::HashSet;
-use std::sync::LazyLock;
 
 #[derive(Debug, Eq)]
 struct Merge {
@@ -226,10 +225,6 @@ impl Default for BpeTrainer {
     }
 }
 
-/// for utf8 boundaries, we need to map gpt2 encoded bytes back
-static CHAR_BYTES: LazyLock<AHashMap<char, u8>> =
-    LazyLock::new(|| bytes_char().into_iter().map(|(b, c)| (c, b)).collect());
-
 impl BpeTrainer {
 
     pub fn new(min_frequency: u64, vocab_size: usize) -> Self {
@@ -327,18 +322,17 @@ impl BpeTrainer {
         let is_a_complete = std::str::from_utf8(&bytes_a).is_ok();
         let is_b_complete = std::str::from_utf8(&bytes_b).is_ok();
 
-        // Rule 1: Allow merging two complete tokens.
+        // - Allow merging two complete tokens.
+        // - Any mix of complete and incomplete is disallowed.
         if is_a_complete && is_b_complete {
             return true;
         }
-
-        // Rule 3 (Implicit): Any mix of complete and incomplete is disallowed.
-        if is_a_complete || is_b_complete {
+        if is_a_complete ^ is_b_complete {
             return false;
         }
 
-        // Rule 2: Both tokens are incomplete. Allow merge only if building a valid
-        // UTF-8 prefix by appending a single byte.
+        // Here we know both tokens are incomplete.
+        // Allow merge only if building a valid UTF-8 prefix by appending a single byte.
         if bytes_b.len() == 1 {
             let mut merged = bytes_a;
             merged.extend_from_slice(&bytes_b);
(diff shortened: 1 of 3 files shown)


tokenizers/src/pre_tokenizers/byte_level.rs
```diff
@@ -12,7 +12,7 @@ use crate::utils::macro_rules_attribute;
 
 /// Converts bytes to unicode characters.
 /// See https://github.com/openai/gpt-2/blob/master/src/encoder.py#L9
-pub(crate) fn bytes_char() -> AHashMap<u8, char> {
+pub fn bytes_char() -> AHashMap<u8, char> {
```
sftse: Why is this made `pub`, this seems like a mistake. In fact, it seems even the previous `pub(crate)` is misplaced. Can you make this private and reuse the `CHAR_BYTES` static in this module by making it `pub(crate)`?

tokenizers/src/models/bpe/trainer.rs
```diff
@@ -831,4 +922,102 @@ mod tests {
         .collect();
         assert_eq!(trained_vocab, expected_vocab)
     }
+
+    // The CHAR_TO_BYTE mapping is kept here *only* for the debug printing helper,
+    // to make the test output readable. It is not used in the core test logic.
+    static BYTE_TO_CHAR: LazyLock<AHashMap<u8, char>> = LazyLock::new(bytes_char);
+    static CHAR_TO_BYTE: LazyLock<AHashMap<char, u8>> =
+        LazyLock::new(|| BYTE_TO_CHAR.iter().map(|(b, c)| (*c, *b)).collect());
```
sftse: Can you reuse the statics already in the codebase, see other comment?

tokenizers/src/models/bpe/trainer.rs
```diff
@@ -270,6 +292,68 @@ impl BpeTrainer {
...
+
+        // Rule 1: Allow merging two complete tokens.
+        if is_a_complete && is_b_complete {
+            return true;
+        }
+
+        // Rule 3 (Implicit): Any mix of complete and incomplete is disallowed.
+        if is_a_complete || is_b_complete {
+            return false;
+        }
```
sftse: Can this be changed to use xor? I had a bit longer than necessary to grasp that the `||` works the same as `^` because the `is_a_complete && is_b_complete` check has already been done. With `^` the operator matches the comment more closely.

tokenizers/src/tokenizer/mod.rs
```diff
@@ -534,6 +540,29 @@ where
...
+                    return Err(
+                        "`enforce_utf8_boundaries=True` can only be used with a `ByteLevel` pre-tokenizer."
+                        .into()
+                    );
+                }
+            }
+        }
+        Ok(())
+    }
+
```
sftse, line 546: This code does not compile. The casting `pretok as &dyn Any` is the culprit, because there isn't a `'static` bound on the pretokenizer.

Because of the ugly restrictions on `Any`, it might be better to introduce default trait methods to `Trainer` and `PreTokenizer` like this
```rust
trait Trainer {
    // ... previous methods
    fn enforce_utf8_boundaries(&self) -> Option<bool> {
        None
    }
}

impl Trainer for BpeTrainer {
    // .. previous methods
    fn enforce_utf8_boundaries(&self) -> Option<bool> {
        Some(self.enforce_utf8_boundaries)
    }
}

trait PreTokenizer {
    // ... previous methods
    fn is_byte_level(&self) -> true {
        false
    }
}

impl PreTokenizer for ByteLevel {
    // .. previous methods
    fn is_byte_level(&self) -> bool {
        true
    }
}
```

This is a bit more type-safe, will catch more errors at compile time, and is a bit easier to understand imo, not like `'static` bounds. But this probably should get the input of the maintainers.

tokenizers/src/pre_tokenizers/byte_level.rs
sanderland: the function is used in normalizers/byte_level.rs, but will make it pub(crate) again and use CHAR_BYTES

tokenizers/src/models/bpe/trainer.rs
sanderland: sure

tokenizers/src/models/bpe/trainer.rs
sanderland: will do, just missed them I think :)

tokenizers/src/tokenizer/mod.rs
sanderland: have replaced this with a placeholder awaiting comments by maintainers.

tokenizers/src/pre_tokenizers/byte_level.rs
sftse: Yes, but `normalizers/byte_level.rs` only needs the function to build an exact copy of the static `BYTE_TO_CHAR`. So as a drive-by cleanup, we could make just the statics `pub(crate)` and reuse them across both `pre_tokenizers/byte_level.rs` and `normalizers/byte_level.rs`. We can then make this function private.

ArthurZucker:
Sorry for being super late here and thanks a lot for the PR, will have a look in a bit!

PR #1832, opened 2025-07-23
Unused Unicode Character Filter

sanderland:
This PR adds a Unicode normalizer to the tokenizers library, enabling filtering of unused and private use code points based on Unicode properties. These characters are often artifacts from editing text with proprietary programs and are not useful for NLP tasks. Filtering them out can improve tokenizer quality.

The implementation covers the Rust core, Python bindings, and Node.js bindings, along with corresponding tests.

### DIFF
--- a/bindings/node/index.d.ts
+++ b/bindings/node/index.d.ts
@@ -77,12 +77,7 @@ export function lowercase(): Normalizer
 export function replace(pattern: string, content: string): Normalizer
 export function nmt(): Normalizer
 export function precompiled(bytes: Array<number>): Normalizer
-export interface UnicodeFilterOptions {
-  filterUnassigned?: boolean
-  filterPrivateUse?: boolean
-  filterSurrogate?: boolean
-}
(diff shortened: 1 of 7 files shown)


Narsil:
Thanks for the great PR !! 

You created 3 booleans. Shouldn't we maybe use a flag array so users can filter in/out any particular categories maybe ?

Are there no way to reuse any of the pre-existing dependencies maybe ?

https://en.wikipedia.org/wiki/UTF-8#Surrogates
https://en.wikipedia.org/wiki/Private_Use_Areas

sanderland:
> Thanks for the great PR !!
> 
> You created 3 booleans. Shouldn't we maybe use a flag array so users can filter in/out any particular categories maybe ?
> 
> Are there no way to reuse any of the pre-existing dependencies maybe ?
> 
> https://en.wikipedia.org/wiki/UTF-8#Surrogates https://en.wikipedia.org/wiki/Private_Use_Areas

* Simplified a bit and removed the surrogates option which is not used in utf8 anyway
* Looked into using the existing dependencies, but could not get the unassigned filter working there

PR #1841, opened 2025-08-06
feat: whitespace optimize

### LINKED ISSUE #1825: Proposal: Replace regex in `whitespace.rs` with manual code for speed improvements
Hi Hugging Face Tokenizers Team,

I’ve been exploring the `whitespace.rs` pre-tokenizer code and noticed it relies on regex for splitting tokens by whitespace.

I wanted to propose replacing this regex-based approach with a manual implementation using byte-level scanning (e.g., `memchr` or custom iterator), which could significantly improve performance by reducing overhead and allocations.

**Key points:**

* The manual approach is expected to bring **substantial speed improvements**, especially on large volumes of text.
* However, this change **may not always produce exactly the same tokenization results** in edge cases compared to the current regex (e.g., certain Unicode whitespace or unusual characters).
* This trade-off might be acceptable depending on the use case, or it could be offered as an optional alternative pre-tokenizer mode.
* I’m happy to prototype and benchmark this approach to demonstrate the performance benefits.

Would you be open to discussing this optimization and its potential impact? I’d love to hear your thoughts on whether this would align with the goals of the `tokenizers` repo and the importance of output consistency vs speed in this context.

Thanks for your time and for creating such an amazing project!

Best regards,
AndriaK

b00f:
This PR fixes #1825 by introducing `WhitespaceOptimized` struct.

### DIFF
--- a/tokenizers/benches/whitespace_benchmark.rs
+++ b/tokenizers/benches/whitespace_benchmark.rs
@@ -3,7 +3,7 @@ extern crate criterion;
 
 use criterion::{Criterion, Throughput};
 use tokenizers::pre_tokenizers::whitespace::{Whitespace, WhitespaceOptimized};
-use tokenizers::{OffsetReferential, OffsetType, PreTokenizer, PreTokenizedString};
+use tokenizers::{OffsetReferential, OffsetType, PreTokenizedString, PreTokenizer};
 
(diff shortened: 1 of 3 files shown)


ArthurZucker (commented):
If it is 10-30% speed boost very nice but needs to be tested for exactitude -> xnli is a good dataset for that!

PR #1842, opened 2025-08-06
feat: add cli for tokenizer and training

b00f:
This PR adds CLI with two subcommands:
1. `Tokenize`: To tokenize a text using a given model
2. `Train`: To train a new tokenize model

### DIFF
--- a/tokenizers/bin/tokenize.rs
+++ b/tokenizers/bin/tokenize.rs
@@ -1,0 +1,77 @@
+use clap::{Parser, Subcommand};
+use tokenizers::tokenizer::Tokenizer;
+
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+struct Cli {
+    #[command(subcommand)]
+    command: Commands,
+}
+
+#[derive(Subcommand, Debug)]
+enum Commands {
+    /// Tokenize input text using a model file
+    Tokenize {
+        /// Path to the tokenizer model file (e.g., tokenizer.json)
+        #[arg(long)]
+        model: String,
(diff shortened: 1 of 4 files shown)


tokenizers/bin/tokenize.rs
```diff
@@ -0,0 +1,77 @@
...
+                eprintln!("Failed to load tokenizer model: {}", e);
+                std::process::exit(1);
+            }
+        },
+        Commands::Train {
+            files,
+            vocab_size,
+            output,
+        } => {
+            use tokenizers::models::bpe::{BpeTrainer, BPE};
```
ArthurZucker, line 59: probably want to add all flavors here no? Unigram etc

ArthurZucker (commented):
Nice! we probably need to update the CI build release to make sure it exposes the bin as well!

tokenizers/bin/tokenize.rs
b00f: Not sure what do you mean, could you please explain more?

Narsil:
If it's that small, and adds new dependencies, I feel like it should be it's own crate. tokenizers is a library, it shouldn't be a CLI as well.

b00f:
> If it's that small, and adds new dependencies, I feel like it should be it's own crate. tokenizers is a library, it shouldn't be a CLI as well.

This is a valid point... I am not sure if we can define `bin` dependencies separately.

ArthurZucker:
We can do a new crate just fort the CLI, it does make sense instead. If this is highly requested happy to do!

Issue #1797, opened 2025-06-11
py03 async bindings for encode/decode in rust


michaelfeil:
Py03 unlocks the GIL, which is great. Most LLM inference servers have many cores (>200), but are blocked by the GIL. 
Also, most servers are async, by nature that python thread based parallelism isn't great. 

As most tokenization is looking something like this:

```
    def _encode(self, prompt: str) -> List[int]:
        """Encode using the rust tokenizer directly, while relaizing gil"""
        return self.tokenizer.encode(prompt, add_special_tokens=True)

    async def encode_prompt(self, prompt: str) -> List[int]:
        f len(prompt) > 2_000:
            # offload to thread to avoid blocking the event loop
            loop = asyncio.get_running_loop()
            tokenized = await loop.run_in_executor(self._threadpool, self._encode, prompt)
        else:
            tokenized = self._encode(
                prompt,
            )
        return tokenized[1:]
```

Proposal: Adding Py03-async-runtimes, a async runtime option for encode/decode. 
Its potentially worth it for every operation that takes >1ms, or every encode step. 

Similar async vs sync usage:
https://github.com/basetenlabs/truss/blob/0816876a474b0c4910eaa3f869ed4c685f7a7570/baseten-performance-client/src/lib.rs#L659C1-L760C20 
also, e.g. sglang uses the primitive and could be directly pluged in there. https://github.com/sgl-project/sglang/blob/777688b8929c877e4e28c2eac208d776abe4c3af/python/sglang/srt/managers/tokenizer_manager.py#L454

michaelfeil:
cc @Narsil @ArthurZucker

ArthurZucker:
Personnaly happy to have if it helps the community!

davidhewitt:
Was this closed in #1843?

ArthurZucker:
Yes let’s close!

michaelfeil:
@davidhewitt if you find issues with the async or sync interfaces, feel free to optimize or flag.

PR #1843, opened 2025-08-10
Tokenizer: Add native async bindings, via py03-async-runtimes.

michaelfeil:
This PR adds native support via py03-async bindings. https://github.com/PyO3/pyo3-async-runtimes as an extra: 
```
pip install tokenizers[async]
```
gives access to:
```python
tokenizer = Tokenizer.from_pretrained("gpt2") 

out = await tokenizer.async_encode("Hey how are you?")
```

### Why is this relevant:
This is mostly relevant for online-inference (vllm, sglang, trt-llm, ..) that have only 1 python thread. 
A common scenario, is that e.g. few users will request very long input (e.g 160k tokens), which typically will take >0.5s to process. 

 A solution is to use the `batch_encode()` py03 api, which releases GIL.  Since the operation is still blocking, you would still starve the asyncio python runtime, which has only 1 thread, for a single task. 
 Relive would be the use of e.g. ray workers, or threadpools.

Quote: vLLM docs: https://docs.vllm.ai/en/v0.8.3/serving/openai_compatible_server.html
```
--tokenizer-pool-size
Size of tokenizer pool to use for asynchronous tokenization. If 0, will use synchronous tokenization.

Default: 0

--tokenizer-pool-type
Type of tokenizer pool to use for asynchronous tokenization. Ignored if tokenizer_pool_size is 0.

Default: “ray”

--tokenizer-pool-extra-config
Extra config for tokenizer pool. This should be a JSON string that will be parsed into a dictionary. Ignored if tokenizer_pool_size is 0.
```

The dependency of ray is much heavier than e.g. py03. For e.g. a small project of mine (github.com/michaelfeil/infinity), it seems overkill.

Summary:
- benefits: async training
- benefits: inference server
- not interesting for: pretraining, dataset preprocessing, torch dataloader etc.

### How is it implemented
I use the same approach that has been tested in py03 here: https://github.com/basetenlabs/truss/tree/main/baseten-performance-client/python_bindings. 
The py03_async runtime requires a non-main thread rust runtime to be initialized. This is done via LazyInit. 
https://github.com/PyO3/pyo3-async-runtimes 

### Performance:
Performance is okay ish. Probably better than the current pools, but worse than the sync bindings. If you have many threads available, and can contend your gil without running the gpu in the same pid, its propably faster

Below a example of what you need to do in async inference endinges;

```
await loop.run_in_executor(
                    executor, 
                    lambda: self.tokenizer.encode_batch_fast(large_batch)
                )
```

```
try:
            executor = concurrent.futures.ThreadPoolExecutor(
                max_workers=2048
            )
            loop = asyncio.get_running_loop()

            async def encode_sync_with_executor(_):
                # Use the pre-initialized executor
                return await loop.run_in_executor(
                    executor, 
                    lambda: self.tokenizer.encode_batch_fast(large_batch)
                )
                
            async def encode_to_thread_sync(_):
                return await asyncio.to_thread(
                    self.tokenizer.encode_batch_fast, large_batch
                )
            
            async def encode_async(_):
                return await self.tokenizer.async_encode_batch_fast(large_batch)
            
            await asyncio.gather(*[encode_sync_with_executor(i) for i in range(2048)])
            await asyncio.gather(*[encode_async(i) for i in range(2048)])
            
            for n_tasks in [1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048]:
                # Measure sync performance with pre-initialized executor
                # Warm up
                await asyncio.gather(*[encode_sync_with_executor(i) for i in range(10)])
                time.sleep(0.03)
                # Actual measurement
                start = time.perf_counter()
                await asyncio.gather(*[encode_sync_with_executor(i) for i in range(n_tasks)])
                sync_time = time.perf_counter() - start

                # Measure async performance
                # Warm up
                await asyncio.gather(*[encode_async(i) for i in range(10)])
                
                # Actual measurement
                time.sleep(0.03)
                start = time.perf_counter()
                await asyncio.gather(*[encode_async(i) for i in range(n_tasks)])
                async_time = time.perf_counter() - start
                
                # Log times
                print(f"sync vs async processing times: {sync_time:.4f}s vs {async_time:.4f}s for {n_tasks} tasks")
                results_sync.append(sync_time)
                results_async.append(async_time)
        finally:
            # Make sure we shut down the executor properly
            executor.shutdown(wait=False)
```

```
sync vs async processing times: 0.0003s vs 0.0004s for 1 tasks
sync vs async processing times: 0.0004s vs 0.0005s for 2 tasks
sync vs async processing times: 0.0004s vs 0.0007s for 4 tasks
sync vs async processing times: 0.0014s vs 0.0011s for 8 tasks
sync vs async processing times: 0.0028s vs 0.0022s for 16 tasks
sync vs async processing times: 0.0057s vs 0.0040s for 32 tasks
sync vs async processing times: 0.0097s vs 0.0065s for 64 tasks
sync vs async processing times: 0.0172s vs 0.0127s for 128 tasks
sync vs async processing times: 0.0324s vs 0.0293s for 256 tasks
sync vs async processing times: 0.0528s vs 0.0487s for 512 tasks
sync vs async processing times: 0.1549s vs 0.0960s for 1024 tasks
sync vs async processing times: 0.3044s vs 0.1846s for 2048 tasks
```

#1797 cc @ArthurZucker

### DIFF
--- a/bindings/python/Cargo.toml
+++ b/bindings/python/Cargo.toml
@@ -15,6 +15,9 @@ serde_json = "1.0"
 libc = "0.2"
 env_logger = "0.11"
 pyo3 = { version = "0.25", features = ["abi3", "abi3-py39", "py-clone"] }
+pyo3-async-runtimes = { version = "0.25", features = ["tokio-runtime"] }
+tokio = { version = "1.47.1", features = ["rt", "rt-multi-thread", "macros", "signal"] }
+once_cell = "1.19.0"
 numpy = "0.25"
 ndarray = "0.16"
 itertools = "0.14"
--- a/bindings/python/py_src/tokenizers/__init__.pyi
+++ b/bindings/python/py_src/tokenizers/__init__.pyi
@@ -929,6 +929,56 @@ class Tokenizer:
 
         """
         pass
+    
+    async def async_encode_batch(self, input, is_pretokenized=False, add_special_tokens=True):
+        """
+        Asynchronously encode the given batch of inputs. This method can be used to
+        encode multiple sequences in parallel.
+
+        Args:
+            input (A :obj:`List`/:obj:`Tuple` of :obj:`~tokenizers.EncodeInput`):
+                A list of single sequences or pair sequences to encode. Each sequence
+                can be either raw text or pre-tokenized, according to the ``is_pretokenized``
+                argument:
+
+                - If ``is_pretokenized=False``: :class:`~tokenizers.TextEncodeInput`
+                - If ``is_pretokenized=True``: :class:`~tokenizers.PreTokenizedEncodeInput`
+
+            is_pretokenized (:obj:`bool`, defaults to :obj:`False`):
+                Whether the input is already pre-tokenized
+
+            add_special_tokens (:obj:`bool`, defaults to :obj:`True`):
+                Whether to add the special tokens
+
+        Returns:
+            A :obj:`List` of :class:`~tokenizers.Encoding`: The encoded batch
+        """
+        pass
+    
+    async def async_encode_batch_fast(self, input, is_pretokenized=False, add_special_tokens=True):
+        """
+        Asynchronously encode the given batch of inputs. This method is faster than
+        `async_encode_batch` because it doesn't keep track of offsets, they will be all zeros.
+
+        Args:
+            input (A :obj:`List`/:obj:`Tuple` of :obj:`~tokenizers.EncodeInput`):
+                A list of single sequences or pair sequences to encode. Each sequence
+                can be either raw text or pre-tokenized, according to the ``is_pretokenized``
+                argument:
+
+                - If ``is_pretokenized=False``: :class:`~tokenizers.TextEncodeInput`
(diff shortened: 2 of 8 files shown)


ArthurZucker:
Thanks for the PR, will have a look!

ArthurZucker:
@michaelfeil I took the liberty to commit my updates as I want to release today!

bindings/python/Cargo.toml
```diff
@@ -15,6 +15,9 @@ serde_json = "1.0"
 libc = "0.2"
 env_logger = "0.11"
 pyo3 = { version = "0.25", features = ["abi3", "abi3-py39", "py-clone"] }
+pyo3-async-runtimes = { version = "0.25", features = ["tokio-runtime"] }
+tokio = { version = "1.47.1", features = ["rt", "rt-multi-thread", "macros", "signal"] }
+once_cell = "1.19.0"
```
ArthurZucker, line 20: I think we can make these optional with "async" feature!

ArthurZucker: Actually not really possible so leaving it as is

ArthurZucker:
thanks @michaelfeil

PR #1844, opened 2025-08-12
mismatch in wordpiece vocab to use AHashMap instead of HashMap

rinechran:
The wordpiece module’s vocab variable is currently a std::collections::HashMap, but the function expects an ahash::AHashMap. This mismatch causes a compilation error E0308.

Fixes the compilation error


ArthurZucker:
Thanks! #1840 was the first PR so I went ahead and merged ti

Issue #1845, opened 2025-08-13
Trouble installing tokenizers (python) on AWS lambda arm64 image.


crawftv:
It seems that a tokenizers installation pulls in 'huggingface_hub' which has a 'hf-xet' dependency which has some interesting platform specific rules which fails my build on install. a dependency resolution issue.
```txt
    "hf-xet>=1.1.3,<2.0.0; platform_machine=='x86_64' or platform_machine=='amd64' or platform_machine=='arm64' or platform_machine=='aarch64'",
```
I noticed that hugging face_hub is only used in 2 places in the python bindings repo. 
Could this dependency moved to the testing extras, `extras["testing"] ` ?

ArthurZucker:
Thanks for sharing, will consider, tho this would be quite breaking for us!

ArthurZucker:
hf-xet should have full support soon I think

ArthurZucker:
(Yep internally asked and latest version should work!)

jbbqqf:
Hi! Triaging older issues — this looks like a duplicate of #1973, which is the more recent and more actively discussed version of the same feature ask (make `huggingface_hub` / `hf-xet` optional so installs on slim runtimes like AWS Lambda arm64 don't fail on transitive deps).

Evidence:

- Same proposal: #1973 ("Make huggingface-hub an optional dependency (extras)", 2026-03-20) by @clemlesne quantifies the cost ("`huggingface-hub` 2.2 MB, `hf-xet` 7.2 MB … users who load tokenizers from local files pay the cost of 12 MB of unused transitive deps") and proposes the same `[hub]` extras split @crawftv suggests here.
- Active maintainer engagement on #1973: @ArthurZucker replied "Thanks for sharing, will consider, tho this would be quite breaking for us!" and "hf-xet should have full support soon I think" — concentrating discussion there avoids splitting the same conversation across two threads.
- Other commenters (@davidmezzetti, @yuchsiao) have piled on to #1973 with the LiteRT / production-slim use cases that match this issue's AWS Lambda arm64 context.

In the meantime, the working hack is `pip install tokenizers --no-deps` (as @davidmezzetti confirmed in #1973). If @crawftv would like to consolidate, you could add your Lambda-arm64 traceback as a comment on #1973 so the maintainer has the full failure mode in one place. @ArthurZucker, would you mind closing this one in favor of #1973?

---
[...]

PR #1848, opened 2025-08-19
Move huggingface-hub dependency to optional test dependencies.

crawftv:
References the issue I created #1845 
https://github.com/huggingface/tokenizers/issues/1845

I propose moving the huggingface_hub dependency to the optional testing dependencies to make installs easier and cleaner.

### DIFF
--- a/bindings/python/pyproject.toml
+++ b/bindings/python/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 ]
 keywords = ["NLP", "tokenizer", "BPE", "transformer", "deep learning"]
 dynamic = ["description", "license", "readme", "version"]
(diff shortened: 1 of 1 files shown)


Narsil:
That's not possible.

tokenizers HARD depends on huggingface_hub for the "from_pretrained" function.
It seems the issue you had was with hf-xet, and should be fixed soon. I'll close this PR for now, but let's reopen if there's anything that actually makes this http dependency problematic for other reasons. (I'll admit this dependency is only for a single function, which should be replaceable in 2 lines, but still it would be a huge breaking change to remove it since basically most HF code uses it.)

clemlesne:
_Deleted — moved to a dedicated issue._

PR #1852, opened 2025-08-25
fix(wordpiece):  use AHashMap instead of HashMap for vocab construction

joshirugved11:
### Description
This PR updates the WordPiece model’s vocabulary construction to use
`ahash::AHashMap` instead of `std::collections::HashMap`. The `tokenizers` crate
consistently relies on `AHashMap` for performance and uniformity, but the code
in `wordpiece/mod.rs` was returning a `HashMap`. This caused the following
compilation error:

### Changes Made
- Updated `wordpiece/mod.rs` to declare and return `AHashMap<String, u32>`
- Added the required `use ahash::AHashMap;` import

### Rationale
- Fixes the type mismatch and restores successful compilation
- Keeps the implementation consistent with the rest of the crate
- Avoids unnecessary conversions between `HashMap` and `AHashMap`

### Impact
- No functional changes beyond correcting the map type
- All tests pass after the change

### ✅ PR Checklist
- [x] Code compiles without errors (`cargo build`)
- [x] All tests pass locally (`cargo test`)
- [x] Code is formatted (`cargo fmt --all`)
- [x] No lints/warnings (`cargo clippy --all-targets --all-features -- -D warnings`)
- [x] Commit message and PR description are clear

fix(wordpiece): use AHashMap instead of HashMap for vocab construction

The WordPiece model was returning a std::collections::HashMap, but the crate
expects an ahash::AHashMap for vocab storage. This caused a type mismatch
during compilation. Updated the vocab initialization to use AHashMap to align
with the rest of the codebase and resolve the build error.

---

### Note to Maintainers
I did not open a separate issue for this since it is a straightforward
compilation fix. Happy to create one if preferred.

### DIFF
--- a/tokenizers/src/models/wordpiece/mod.rs
+++ b/tokenizers/src/models/wordpiece/mod.rs
@@ -175,7 +175,7 @@ impl WordPiece {
     pub fn read_bytes(vocab: &[u8]) -> Result<Vocab> {
         let file = BufReader::new(vocab);
 
-        let mut vocab = HashMap::new();
+        let mut vocab = AHashMap::new();
         for (index, line) in file.lines().enumerate() {
             let line = line?;
             vocab.insert(line.trim_end().to_owned(), index as u32);


ArthurZucker (commented):
Thanks! #1840 came in first so merged it!

joshirugved11:
> Thanks! #1840 came in first so merged it!

No issues, i work on this repo again, thanks for the reply

PR #1861, opened 2025-09-05
Mark immutable pyclasses as frozen

ngoldbaum:
The `Regex` and `Token` class are both immutable and can be trivially marked as frozen. The avoids the cost of atomic reference counting for any reads on borrowed data.

It's likely that in the future `frozen` will become the default for pyclasses.

See https://github.com/huggingface/tokenizers/pull/1860#issuecomment-3253764308 and reply for more details.


ArthurZucker (approved):
Okay ty !

PR #1865, opened 2025-09-13
Add setter to mutable properties

cyyever:


cyyever:
@ArthurZucker Where is the source file?

ArthurZucker:
Actually I'll handle it don't worry!

ArthurZucker:
Thanks tho!

PR #1869, opened 2025-09-29
ci: add support for building Win-ARM64 wheels

MugundanMCW:
- The adoption of Windows on ARM (WoA) devices is steadily increasing, yet many Python wheels are still not available for this platform.
- GitHub Actions now offer native CI runners for Windows on ARM devices (windows-11-arm), enabling automated builds and testing.
- Currently, official tokenizers Python wheels are not provided for Windows ARM64 and thus users/developers were facing difficulties using popular tokenizers library natively.
- This PR introduces support for building tokenizers wheels on Windows ARM64, improving accessibility for developers and end users on this emerging platform.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -94,12 +94,14 @@ jobs:
             target: x64
           - runner: windows-latest
             target: x86
+          - runner: windows-11-arm
+            target: aarch64
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
           python-version: 3.x
-          architecture: ${{ matrix.platform.target }}
+          architecture: ${{ matrix.platform.target == 'aarch64' && 'arm64' || matrix.platform.target }}
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:


.github/workflows/CI.yml
```diff
@@ -94,12 +94,14 @@ jobs:
...
             target: x86
+          - runner: windows-11-arm
+            target: aarch64
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
           python-version: 3.x
-          architecture: ${{ matrix.platform.target }}
+          architecture: ${{ matrix.platform.target == 'aarch64' && 'arm64' || matrix.platform.target }}
```
Narsil: Can we remove this weird condition, and have 2 target names if it's really needed ? Why isn't the target named aarch64 ?

MugundanMCW: Hi @Narsil,
The reason for using a separate Python target name for Windows on ARM is that the official Python GitHub Action only supports the architectures x86, x64, and arm64 for Windows [[Reference](https://github.com/actions/setup-python/tree/2e3e4b15a884dc73a63f962bff250a855150a234?tab=readme-ov-file#supported-architectures)]. Specifying aarch64 for Windows on ARM64 results in an invalid configuration, which causes the CI to fail.

Narsil: I know, but then maybe we could have "platform.target" and "platform.architecture" be defined in the matrix, which should make the flow easier to read.

MugundanMCW: @Narsil thanks for the suggestions, made the relevant changes

PR #1872, opened 2025-10-02
Mark Python tests that need network access

gordonmessmer:
I would like to mark tests that need network access in order to more easily disable them when building the software in an environment that restricts network access.

I'll mark this PR a draft while I try to determine which other tests require network access.

### DIFF
--- a/bindings/python/pytest.ini
+++ b/bindings/python/pytest.ini
@@ -1,0 +1,3 @@
+[pytest]
+markers =
+   network: mark a test that requires network access.
--- a/bindings/python/tests/bindings/test_encoding.py
+++ b/bindings/python/tests/bindings/test_encoding.py
@@ -5,6 +5,7 @@ from tokenizers import BertWordPieceTokenizer
 from ..utils import bert_files, data_dir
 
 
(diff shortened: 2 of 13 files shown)


ArthurZucker (commented):
Hey! Not super sure this is very important, will leave it as is

gordonmessmer:
Please consider marking tests that need network access so that they can be disabled easily.

In Fedora and related distributions, packages are built without network access to ensure that the source code provided is the only information required to produce the binary build artifact. 

SLSA Build Level 4 is likely to specify hermetic builds as a secure build practice, in the future (as early drafts did).  https://slsa.dev/spec/v1.2/future-directions

jpodivin:
I would also like to ask for this change to be implemented. It would simplify CI/CD in cases when network access is limited, for example for security reasons. Furthermore, it wouldn't impact experience of developers in situations when that is not the case.

ArthurZucker:
sur, feel free to rebase fix conflicts!

gordonmessmer:
I've rebased on main and tested a variant of this patch against 0.22.2 for a networkless build.

PR #1878, opened 2025-10-21
DOCS: add `add_prefix_space` to `processors.ByteLevel` 

### LINKED ISSUE #1819: Bug with `add_prefix_space` parameter for ByteLevel post-processor
Hello, I've noticed a bug in the implementation of the `ByteLevel` **post-processor**.

The documentation of this post-processor only lists `trim_offsets` as a configurable parameter: https://huggingface.co/docs/tokenizers/en/api/post-processors#tokenizers.processors.ByteLevel

However, the Rust source code shows that the post-processor also uses an `add_prefix_space` parameter under the hood. The default value for this param is `true`: https://github.com/huggingface/tokenizers/blob/dd4fc3df1a8a7cd135eecca2158db018d85f94f1/tokenizers/src/pre_tokenizers/byte_level.rs#L187

This creates a problem when the `ByteLevel` **pre-tokenizer** is configured with `add_prefix_space=false` ([in this case this param is correctly listed in the documentation](https://huggingface.co/docs/tokenizers/en/api/pre-tokenizers#tokenizers.pre_tokenizers.ByteLevel)). In this scenario, the pre-tokenizer and post-processor are misaligned, as the post-processor incorrectly uses the default `add_prefix_space=true`. This can lead to incorrect offset mappings when post-processor is used with `trim_offsets=true` and when input text starts from a whitespace.

The easiest fix for this is to allow users to configure the `add_prefix_space` parameter of the `ByteLevel` post-processor and describe this option accordingly in the documentation.

CloseChoice:
closes #1819

We already support `add_prefix_space` for `processors.ByteLevel`, it is just not documented since this class is defined via pyo3's `pyclass`, see [here](https://github.com/huggingface/tokenizers/blob/main/bindings/python/src/processors.rs#L487). See further down the signature for the `new` function, which mirrors python's `__new__` method: https://github.com/huggingface/tokenizers/blob/main/bindings/python/src/processors.rs#L492.

### DIFF
--- a/bindings/python/py_src/tokenizers/processors/__init__.pyi
+++ b/bindings/python/py_src/tokenizers/processors/__init__.pyi
@@ -100,8 +100,9 @@ class ByteLevel(PostProcessor):
             Whether to trim the whitespaces from the produced offsets.
 
         add_prefix_space (:obj:`bool`, `optional`, defaults to :obj:`True`):
-            Whether the add_prefix_space option was enabled during pre-tokenization. This
-            is relevant because it defines the way the offsets are trimmed out.
+            If :obj:`True`, keeps the first token's offset as is. If :obj:`False`, increments
+            the start of the first token's offset by 1. Only has an effect if :obj:`trim_offsets`
+            is set to :obj:`True`.
     """
     def __init__(self, trim_offsets=True, add_prefix_state=True):
         pass


bindings/python/src/processors.rs
```diff
@@ -484,12 +484,16 @@ impl PyRobertaProcessing {
 /// Args:
 ///     trim_offsets (:obj:`bool`):
 ///         Whether to trim the whitespaces from the produced offsets.
+///
+///     add_prefix_space (:obj:`bool`, `optional`, defaults to :obj:`True`):
+///         Whether the add_prefix_space option was enabled during pre-tokenization. This
```
ArthurZucker: the doc is wrong in the sense that this does not remove the space in the post processor nor does it add it. It just accounts for it in the offset computation if `trim_offsets`. 

```rust
                    // If we are processing the first pair of offsets, with `add_prefix_space`,
                    // then we shouldn't remove anything we added. If there are more than one
                    // leading spaces though, it means we didn't add them, and they should be
                    // removed.
```
try to incorporate this please!

ArthurZucker (approved):
LGTM but lets update to proper doc

bindings/python/src/processors.rs
CloseChoice: Thanks for the review, I updated the docs. I somehow misunderstood the meaning of `add_prefix_space` for the postprocessor, but it simply refers to the preprocessor and has more of a technical meaning here. To actually understand the meaning I added some code to the tests and thought that might be helpful to keep. Let me know if you want further changes.

ArthurZucker:
just missing make style

CloseChoice:
> just missing make style

done

ArthurZucker:
```
AssertionError: The content of py_src/tokenizers/processors/__init__.pyi seems outdated, please run `python stub.py`
```
😉

CloseChoice:
That was weird, nothing happened when I ran `python stub.py`, just once I merged main it had an effect. So I hope I didn't miss something again, sorry for the inconvenience.

ArthurZucker:
you need to cargo build or `maturin dev` then run `python stub`

ArthurZucker:
Ty!

CloseChoice:
> you need to cargo build or `maturin dev` then run `python stub`

Ah right, thanks!

PR #1880, opened 2025-10-30
Providing byte level offsets for effective alignment in Cross-Tokenizer On-Policy Distillation

JqzChandler:
issue
https://github.com/huggingface/tokenizers/issues/1881

Our team tried different aligning solutions when implementing our self-developed On-Policy distillation method, including trl's existing implementation (which repeatedly calls the decode() method but encounters correctness issues in special cases and has high computational overhead).

Later we found a better method: by making one call to the encode() method to get byte-level offsets for all tokens, we can effectively avoid BPE's complexity, and byte level offsets are also compatible with all other types of tokenizers. Additionally, for distillation between two BPE tokenizers, we can get more accurate alignment by skipping string as an intermediate modality.

Therefore, we hope to merge this simple patch to expose the byte-level offset calculation already supported in the Rust code for use by Python classes.

More description at:
https://github.com/huggingface/trl/issues/4393

### DIFF
--- a/bindings/python/src/tokenizer.rs
+++ b/bindings/python/src/tokenizer.rs
@@ -1255,7 +1255,6 @@ impl PyTokenizer {
                     .into_pyerr::<exceptions::PyValueError>())
                 }
             };
-            
             ToPyResult(result.map(|encodings| encodings.into_iter().map(|e| e.into()).collect())).into()
         })
     }


bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
```diff
@@ -195,6 +195,7 @@ def encode(
         pair: Optional[InputSequence] = None,
         is_pretokenized: bool = False,
         add_special_tokens: bool = True,
+        offset_type: str = "char",
```
ArthurZucker, line 198: I'd rather we make it optional!

bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
ArthurZucker, line 1: btw this file is not really something I thought had a lot of usage 😄 

WDYT about rather adding some doc in a md file or something? As this already exists?

PR #1888, opened 2025-11-21
C and C++ bindings to Tokenizers

thammegowda:
Adding in bindings for two more languages!
* `bindings/cpp` 
* `bindings/c`

C is an intermediate step to bind C++ and Rust:  i.e., C++ <--> C <--> Rust. 

--
* Added tests to c++ 
* Added benchmarks for my sanity checks and the results are as expected.

### DIFF
--- a/tokenizers/src/tokenizer/serialization.rs
+++ b/tokenizers/src/tokenizer/serialization.rs
@@ -143,7 +143,11 @@ where
                 "post_processor" => {
                     builder = builder.with_post_processor(map.next_value()?);
                 }
-                _ => {}
+                _ => {
+                    // Skip unknown fields (consume the value); useful behavior for experimental 
+                    // wrappers who may store their own configs
+                    map.next_value::<serde::de::IgnoredAny>()?;
+                }
             };
         }
         let mut tokenizer = builder


bindings/c/src/lib.rs
```diff
@@ -0,0 +1,133 @@
+use std::ffi::{CStr, CString};
```
ArthurZucker, line 1: ok ffi perfect

bindings/c/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+#[no_mangle]
+pub extern "C" fn tokenizers_free(tokenizer: *mut c_void) {
+    if tokenizer.is_null() { return; }
+    unsafe { drop(Box::from_raw(tokenizer as *mut CTokenizer)); }
+}
+
+#[no_mangle]
+pub extern "C" fn tokenizers_encode(
```
ArthurZucker, line 190: think we need decode, batch encode batch encode fast etc and asyc? I don't know how best to define these all but basically the same surface as python!

thammegowda: Thanks for the quick review! Happy to know you're interested/open to this effort. 
I will get back after surfacing more functionality to C++.

thammegowda:
Hi @ArthurZucker I've made some progress with c++ adding more APIs. 
But... templating like jinaj2 is giving me a bit of trouble.  Wondering how your team is handling templates, e..g. chat_template -- is there a native support in the rust code for chat_template in jinja2 format?

I tried integrating https://github.com/jinja2cpp/Jinja2Cpp/ at c++ bindings layer, but some features are limited  and lead to crash (IIRC, negative index like `messages[-1]`)

Any tips/recommendation for handling templates natively would be great.

thammegowda:
Upon digging https://github.com/huggingface/text-generation-inference codebase, I found it's using minijina in Rust to render chat templates (with some workarounds for unsupported features).  So, I've removed jinja rendering at c++ bindings layer, instead moved it to tokenizer core in Rust, using minijinja2. This way, all bindings can access the functionality. 

Disclaimer: I'm not proficient in Rust, and most of the code is done by AI agents (though I've tried to closely supervise it/them). Based on my testing, everything seems to work (at least for my usecase and its tests pass).

IvanIsCoding:
I will drop this more as a FYI: check https://github.com/mlc-ai/tokenizers-cpp

There's an existing C++ bindings from Tokenizers that goes through the same Rust -> C -> C++ path. The C++ code also binds more than tokenizers because it includes sentencepiece, but that can be cut off.

I think if you want a more battle tested code, fork [tokenizers-cpp/rust](https://github.com/mlc-ai/tokenizers-cpp/tree/main/rust) into your current `bindings/c` folder and ship [tokenizers_c.h](https://github.com/mlc-ai/tokenizers-cpp/blob/main/include/tokenizers_c.h).

Once that is done, fork [tokenizers_cpp.h](https://github.com/mlc-ai/tokenizers-cpp/blob/main/include/tokenizers_cpp.h) and [huggingface_tokenizer.cc](https://github.com/mlc-ai/tokenizers-cpp/blob/main/src/huggingface_tokenizer.cc) into `bindings/cpp`.

The LICENSE shouldn't be an issue. And I think https://chat.webllm.ai/ has a live deployment of the tokenizer with WASM.

johnmai-dev:
Hi! Is there any plan to merge this PR? I’m currently looking for tokenizers implemented in C.

ArthurZucker:
Actually yeah but I need to take over a little and check a few things but planned for sure

benchmarks/README.md
```diff
@@ -0,0 +1,84 @@
...
+   - Expected overhead from Python interpreter
+   - Produces identical results to Rust
+
+### Key Findings
+
+#### Speed Comparison (All Implementations)
+- **Rust** (baseline): 100%
+- **C Bindings**: ~100% (essentially identical to Rust)
+- **C++ Bindings**: 97.6% (only 2.4% slower)
+- **Python**: 66.5% (33.5% slower)
```
ArthurZucker, line 53: cc @McPatate there's stuff we can do to improve this i'm sure 👁️ 👁️

bindings/c/src/lib.rs
```diff
@@ -0,0 +1,795 @@
...
+#[derive(Default, Clone)]
+struct TokenizerConfig {
+    bos_token: Option<String>,
+    eos_token: Option<String>,
+    pad_token: Option<String>,
+    unk_token: Option<String>,
+    chat_template: Option<String>,
+    add_bos_token: bool,
+    add_eos_token: bool,
+}
```
ArthurZucker, line 29: We partially have a PR for #1942 + this is not in the core `tokenizers` we should not add it !

ArthurZucker: `chat_template` potentially we'll add support for it, never `add_bos_token` as `tokenizer.json` has that info

bindings/c/src/lib.rs
```diff
@@ -0,0 +1,795 @@
...
+
+/// Opaque tokenizer type exposed as void* on the C side.
+/// Contains tokenizer + optional config (auto-loaded from same directory)
+struct CTokenizer {
+    tokenizer: Tokenizer,
+    config: Option<TokenizerConfig>,
+}
+
+impl CTokenizer {
+    fn new_from_file(path: &str, config_path: Option<&str>) -> Option<Self> {
```
ArthurZucker, line 88: the C binding will be at worst more minimal than the `python` one, at best the same. So we should not have anything extra like `TokenizerConfig`

bindings/c/src/lib.rs
ArthurZucker: Yeah I am interested! I'd actually want to define explicitly the surface we actually want to maintain! 
I don't have a good solution will talk about it with @McPatate but the idea is to reduce it as much as possible!

bindings/c/src/lib.rs
```diff
@@ -0,0 +1,795 @@
...
+///   - template: Jinja2 template string
+///   - messages_json: JSON array of messages with "role" and "content" fields
+///   - add_generation_prompt: whether to append generation prompt
+///   - bos_token: optional BOS token string
+///   - eos_token: optional EOS token string
+///   - error_out: pointer to error string (caller must free with tokenizers_string_free)
+///
+/// Returns: rendered template string (caller must free with tokenizers_string_free), or null on error
+#[no_mangle]
+pub extern "C" fn tokenizers_apply_chat_template(
```
ArthurZucker, line 688: for now this should never go here unless its in rust core!

bindings/cpp/example/main.cpp
ArthurZucker, line 1: I like the idea of having examples for each bidnings!

tokenizers/src/chat_template.rs
ArthurZucker, line 1: let's split this in another PR! 🤗

ArthurZucker: not 100% sure this makes sense to have in `tokenizers`

PR #1782, opened 2025-05-27
Add benchmark for deserializing large added vocab +  optimizations

ArthurZucker:

PR #1891, opened 2025-11-27
Update deserialize of added tokens

### LINKED ISSUE #1635: Adding many AddedTokens makes loading a tokenizer extremely slow.
Hi!

I'm not sure if this is a problem that can be solved, or needs to be solved. Basically, we want to make a kind of hybrid tokenizer, in which we add a whole bunch of whole words to a tokenizer, and select these words instead of the subwords if they appear. 

For example: if we pass the pretokenized string `["dog", "walks", "around", "Paris"]`, and "Paris" is a whole token, we want to select it instead of decomposing it into subtokens. I think that adding `Paris` as an `AddedToken` is the right approach for this (but please correct me if I'm wrong.)

So, we added many of these tokens (about 400k), but this makes loading a tokenizer extremely slow, like, it takes 15-30 minutes to load. We now add them as regular tokens, which works fine, but which has the downside of also finding these whole word tokens as part of other words. For example `Parisians` will now be turned into `["Paris", "##ians"]`, which might have a different meaning.

So my main question is: is there a reason why adding many `AddedToken`s is slow? Or is this just a path that hasn't been fully optimized yet? 

Is using `AddedToken`s in this way simply wrong? Should we be trying something else?

Thanks!
Stéphan

ArthurZucker:
8-9x speedup on special tokens and about 4x on non-special cases for deserialization

Fixes #1635 and superseeds #1782

### DIFF
--- a/bindings/node/yarn.lock
+++ b/bindings/node/yarn.lock
@@ -4405,14 +4405,14 @@ __metadata:
   linkType: hard
 
 "js-yaml@npm:^3.13.1, js-yaml@npm:^3.14.1":
-  version: 3.14.1
-  resolution: "js-yaml@npm:3.14.1"
+  version: 3.14.2
+  resolution: "js-yaml@npm:3.14.2"
   dependencies:
     argparse: ^1.0.7
     esprima: ^4.0.0
   bin:
     js-yaml: bin/js-yaml.js
-  checksum: bef146085f472d44dee30ec34e5cf36bf89164f5d585435a3d3da89e52622dff0b188a580e4ad091c3341889e14cb88cac6e4deb16dc5b1e9623bb0601fc255c
+  checksum: 626fc207734a3452d6ba84e1c8c226240e6d431426ed94d0ab043c50926d97c509629c08b1d636f5d27815833b7cfd225865631da9fb33cb957374490bf3e90b
   languageName: node
   linkType: hard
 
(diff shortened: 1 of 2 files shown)


ArthurZucker:
@jannisborn ty for testing 😉

jannisborn:
Amazing @ArthurZucker, it's still lightspeed fast. Loading a tokenizer with 150K tokens, 140K of which are special tokens, takes 3.5sec
I re-tokenized 200K samples and compared to previous tip of main and it's still identical to the last token 🚀

ArthurZucker:
Damnnn amazing! 🤗

.github/workflows/rust.yml
```diff
@@ -94,7 +94,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: audit
-          args: -D warnings -f ./tokenizers/Cargo.lock --ignore RUSTSEC-2024-0436 --ignore RUSTSEC-2025-0014
+          args: -D warnings -f ./tokenizers/Cargo.lock --ignore RUSTSEC-2024-0436 --ignore RUSTSEC-2025-0014 --ignore RUSTSEC-2025-0119
```
McPatate, line 97: Could be the opportunity to do some cleaning within our deps, but don't know the context that much.

ArthurZucker: yeah just unmaintained...

PR #1894, opened 2025-11-28
Python: Add ruff rules for asyncio and performance

cclauss:
```toml
lint.extend-select = [
  "ASYNC",
  "C4",
  "PERF",
]
```
https://docs.astral.sh/ruff
* https://docs.astral.sh/ruff/rules/#flake8-async-async
* https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4
* https://docs.astral.sh/ruff/rules/#perflint-perf

Extra review please on the `time.sleep() --> asyncio.sleep()` in `bindings/python/tests/bindings/test_tokenizer.py` because these are tests, not production code.

---
% `cd bindings/python`
% `ruff check --extend-select=ASYNC,C4,PERF --output-format=concise`
```
benches/test_tiktoken.py:32:11: F541 [*] f-string without any placeholders
examples/using_the_visualizer.ipynb:cell 13:1:21: C408 Unnecessary `dict()` call (rewrite as a literal)
py_src/tokenizers/tools/visualizer.py:167:18: C417 Unnecessary `map()` usage (rewrite using a set comprehension)
scripts/sentencepiece_extractor.py:141:37: C417 Unnecessary `map()` usage (rewrite using a generator expression)
tests/bindings/test_tokenizer.py:344:20: C419 Unnecessary list comprehension
tests/bindings/test_tokenizer.py:953:17: ASYNC251 Async functions should not call `time.sleep`
tests/bindings/test_tokenizer.py:964:17: ASYNC251 Async functions should not call `time.sleep`
Found 7 errors.
```
% `ruff rule ASYNC251`
# blocking-sleep-in-async-function (ASYNC251)

Derived from the **flake8-async** linter.

## What it does
Checks that async functions do not call `time.sleep`.

## Why is this bad?
Blocking an async function via a `time.sleep` call will block the entire
event loop, preventing it from executing other tasks while waiting for the
`time.sleep`, negating the benefits of asynchronous programming.

Instead of `time.sleep`, use `asyncio.sleep`.

## Example
```python
import time


async def fetch():
    time.sleep(1)
```

Use instead:
```python
import asyncio


async def fetch():
    await asyncio.sleep(1)
```

### DIFF
--- a/bindings/python/pyproject.toml
+++ b/bindings/python/pyproject.toml
@@ -50,6 +50,11 @@ features = ["pyo3/extension-module", "abi3"]
 [tool.ruff]
 line-length = 119
 target-version = "py311"
+lint.extend-select = [
+  "ASYNC",
+  "C4",
+  "PERF",
+]
 lint.ignore = [
   # a == None in tests vs is None.
   "E711",
--- a/bindings/python/benches/test_tiktoken.py
+++ b/bindings/python/benches/test_tiktoken.py
@@ -29,7 +29,7 @@ def benchmark_batch(model: str, documents: list[str], num_threads: int, document
     os.environ["RAYON_NUM_THREADS"] = str(num_threads)
     num_bytes = sum(map(len, map(str.encode, documents)))
     readable_size, unit = format_byte_size(num_bytes)
-    print(f"==============")
+    print("==============")
     print(
         f"num_threads: {num_threads}, data size: {readable_size}, documents: {len(documents)} Avg Length: {document_length:.0f}"
     )
--- a/bindings/python/examples/using_the_visualizer.ipynb
+++ b/bindings/python/examples/using_the_visualizer.ipynb
@@ -552,7 +552,7 @@
     }
    ],
    "source": [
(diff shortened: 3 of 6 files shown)


bindings/python/pyproject.toml
```diff
@@ -53,6 +53,11 @@ target-version = ["py35"]
 [tool.ruff]
 line-length = 119
 target-version = "py311"
+lint.extend-select = [
+  "ASYNC",
+  "C4",
+  "PERF",
```
ArthurZucker, line 56: can you explain this?

ArthurZucker (commented):
sure why not!

bindings/python/pyproject.toml
cclauss: We are adding these three families of rules to the ruff linter.  The rules are documented in the URLs in the commit message.

https://docs.astral.sh/ruff/settings/#lint_extend-select

ArthurZucker (approved):
can you resolve conflicts please?

PR #1901, opened 2025-11-30
bump PyO3 to 0.26

davidhewitt:
First step towards getting tokenizers working on Python 3.14 free-threading.

This PR bumps PyO3 to 0.26. There are three major deprecations in this release, which cleaned up a bunch of PyO3 APIs to remove the "gil" association (which free-threading doesn't have). Each is carried out as a separate commit; in the final commit removing the `PyObject` type alias for `Py<PyAny>` I happened upon a number of bits of code which had obvious simplifications. I applied those at the same time.

### DIFF
--- a/bindings/python/src/models.rs
+++ b/bindings/python/src/models.rs
@@ -33,7 +33,7 @@ pub struct PyModel {
 }
 
 impl PyModel {
-    pub(crate) fn get_as_subtype<'py>(&self, py: Python<'_>) -> PyResult<Py<PyAny>> {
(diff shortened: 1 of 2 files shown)


davidhewitt:
> LGTM Is my understanding correct that with pyo3 0.26, building for py<3.14 will still work?

Yes, we still support python 3.7+.

I'll fix the lint failures ASAP, probably later today.

davidhewitt:
Lints fixed in 78af19177669b11abf81d837662eca052d6fcf94

PR #1912, opened 2025-12-14
Use `unicode-normalization` instead of `unicode-normalization-alignments`

IvanIsCoding:
Builds upon https://github.com/huggingface/tokenizers/pull/19, 6 years after the original PR

This brings the updates of `unicode-normalization` to Tokenizers. It turns out that the changes made in the original work do not require a full fork of  `unicode-normalization`.

During the time, `unicode-normalization` went from `0.1.12` to `0.1.25` with 100+ commits. Unicode went from version 12.1 to 17.0.

Instead of having to update the fork (which has never happened), reusing the original dependency will bring updates like the Unicode version bumps ([example](https://github.com/unicode-rs/unicode-normalization/pull/113)). Those should be trivial to handle, at max we'll need to change `Cargo.toml` instead of updating the fork.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -122,7 +122,7 @@ jobs:
     strategy:
       matrix:
         platform:
-          - runner: macos-13
+          - runner: macos-15-intel
             target: x86_64
           - runner: macos-14
             target: aarch64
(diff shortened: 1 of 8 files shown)


IvanIsCoding:
@ArthurZucker can you also take a look at this when updating `spm_precompiled` for #1917? Thanks.

PR #1931, opened 2026-01-20
Include license file into python wheels

justeph:
python wheels distributed on PyPI don't include the license. Fix it, by symlinking the LICENSe file into bindings/python the same way it's done for rust crate in https://github.com/huggingface/tokenizers/commit/67080e163add50e278dce1407538e5cd03ef9d6d


ArthurZucker (approved):
Thanks

ArthurZucker:
Tho I do see on `pypi`: License: Apache Software License 
What is missing for you? the full file?

justeph:
> Tho I do see on `pypi`: License: Apache Software License What is missing for you? the full file?

Yes. When you distribute a wheel you should add the license file. Right now, if you download and unzip a wheel from pypi, it will not contain the License file. Only the name of the license is included in the METADATA file

Issue #1927, opened 2026-01-11
[FR] Add type hint for python


xcpky:
Could you please add type hint support for python bindings, even ?

my language server(basedpyright) complains everywhere tokenziers package is used

<img width="486" height="196" alt="Image" src="https://github.com/user-attachments/assets/ca6f5996-b43e-460f-8bf5-90628019c708" />

```
/home/ykpcx/euclid/main.py
  /home/ykpcx/euclid/main.py:7:1 - error: Type of "tokenizer" is unknown (reportUnknownVariableType)
  /home/ykpcx/euclid/main.py:7:13 - error: Type of "from_str" is partially unknown
    Type of "from_str" is "(json: Unknown) -> Unknown" (reportUnknownMemberType)
  /home/ykpcx/euclid/main.py:10:5 - error: Type of "output" is unknown (reportUnknownVariableType)
  /home/ykpcx/euclid/main.py:10:14 - error: Type of "encode_batch" is unknown (reportUnknownMemberType)
  /home/ykpcx/euclid/main.py:11:15 - error: Type of "ids" is unknown (reportUnknownMemberType)
  /home/ykpcx/euclid/main.py:11:15 - error: Argument type is unknown
    Argument corresponds to parameter "obj" in function "len" (reportUnknownArgumentType)
  /home/ykpcx/euclid/main.py:12:11 - error: Type of "tokens" is unknown (reportUnknownMemberType)
  /home/ykpcx/euclid/main.py:12:11 - error: Argument type is unknown
    Argument corresponds to parameter "values" in function "print" (reportUnknownArgumentType)
8 errors, 0 warnings, 0 notes
```

ArthurZucker:
Hey! Pretty sure they were added in `0.22.2`! Can you make sure you have that version? 
EDIT: actually we added type check but maybe not type hint!

zhaob1n:
> Hey! Pretty sure they were added in `0.22.2`! Can you make sure you have that version? EDIT: actually we added type check but maybe not type hint!

I am sure I am using 0.22.2, there is no type hints.

ArthurZucker:
Yeah I'll work on that, its a bit experimental TBH https://pyo3.rs/v0.27.1/python-typing-hints is a bit new

ArthurZucker:
Closed in #1928

PR #1932, opened 2026-01-20
fix: added type hints in .py files

ashmi8:
added all the missing type hints for the funciton in python files solution for issue no #1927


ArthurZucker (commented):
copilot' s review is quite in depth 😉

PR #1940, opened 2026-02-03
Update fancy-regex dependency to 0.17

musicinmybrain:
https://github.com/fancy-regex/fancy-regex/blob/0.17.0/CHANGELOG.md

I tried `make test` and `cargo test -F fancy-regex` and didn’t observe any regressions.


ArthurZucker (approved):
ty

PR #1942, opened 2026-02-05
🚨 feat: add role_to_token field for special token metadata

ArthurZucker:
## What this does:

- Adds a new `role_to_token` field to the `Tokenizer` class that maps token roles (`eos_token`, `bos_token`, etc.) to their token strings
- Exposes and getters and setters for any role that is in the `role_to_token`:
```python
tokenizer = Tokenizer.from_pretrained("gpt2")
tokenizer.role_to_token = {"eos_token": "<|endoftext|>"}
print(tokenizer.eos_token)
print(tokenizer.eos_token_id)
```

```
<|endoftext|>
50256
```

You can now do:
```python 
tokenizer.bos_token = "<|bos_token|>"
```
which will add the token if its not already in the added vocab. 

```python 
>>> tokenizer.bos_token_id
50257
>>> tokenizer.decode([50258], False)
'<|bos_token|>'
```

Also fixed this:
```diff
>>> tokenizer.add_token([AddedToken("special", special=True)]
>>> tokenizer.decode(["special"], True)
''
>>> tokenizer.add_token([AddedToken("special", special=False)]
>>> tokenizer.decode(["special"], True)
- ''
+ "special"
```

This however breaks forward compatibility for `tokenizer.json`


bindings/python/py_src/tokenizers/__init__.pyi
```diff
@@ -1195,6 +1199,24 @@ class Tokenizer:
...
+        Get the role to token mapping
+
+        Returns:
+            :obj:`Dict[str, str]` or :obj:`None`: The role to token mapping if set
+        """
+        ...
+    @role_to_token.setter
+    def role_to_token(self, /, role_to_token: typing.Any | None) -> None:
+        """
+        Get the role to token mapping
```
itazap:
```suggestion
        Set the role to token mapping
```

itazap (approved):
awesome 🚀

.github/workflows/CI.yml
```diff
@@ -61,43 +61,43 @@ jobs:
...
+  #       with:
+  #         target: ${{ matrix.platform.target }}
+  #         args: --release --out dist --manifest-path bindings/python/Cargo.toml -i python3.13t
+  #         sccache: ${{ !startsWith(github.ref, 'refs/tags/') }}
+  #         manylinux: musllinux_1_2
+  #     - name: Upload wheels
+  #       uses: actions/upload-artifact@v4
+  #       with:
+  #         name: wheels-musllinux-${{ matrix.platform.target }}
+  #         path: dist
```
ngoldbaum: This should all be fixed with the latest maturin release.

ArthurZucker:
Just have to overwrite the save order to put metadatas first now

.github/workflows/CI.yml
```diff
@@ -24,17 +24,17 @@ jobs:
     strategy:
       matrix:
         platform:
-          - runner: ubuntu-latest
+          - runner: ubuntu-22.04
```
McPatate: are we sure we want to pin to this version of ubuntu?

.github/workflows/CI.yml
```diff
@@ -122,7 +141,7 @@ jobs:
     strategy:
       matrix:
         platform:
-          - runner: macos-15-intel
+          - runner: macos-13
```
McPatate, line 153: 13? why?

.github/workflows/CI.yml
```diff
@@ -92,25 +106,30 @@ jobs:
         platform:
           - runner: windows-latest
             target: x64
-            architecture: x64
           - runner: windows-latest
             target: x86
-            architecture: x86
-          - runner: windows-11-arm
-            target: aarch64
-            architecture: arm64
```
McPatate, line 110: why are we removing windows-11-arm?

.github/workflows/CI.yml
```diff
@@ -92,25 +106,30 @@ jobs:
...
-          architecture: ${{ matrix.platform.architecture }}
+          architecture: ${{ matrix.platform.target }}
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
           target: ${{ matrix.platform.target }}
           args: --release --out dist --manifest-path bindings/python/Cargo.toml
-          sccache: 'true'
+          sccache: ${{ !startsWith(github.ref, 'refs/tags/') }}
+      - uses: actions/setup-python@v5
```
McPatate, line 132: while you're at it, you might want to update all versions of your action dependencies 
```suggestion
      - uses: actions/setup-python@v6
```

.github/workflows/CI.yml
```diff
@@ -92,25 +106,30 @@ jobs:
...
             target: x64
-            architecture: x64
           - runner: windows-latest
             target: x86
-            architecture: x86
-          - runner: windows-11-arm
-            target: aarch64
-            architecture: arm64
     steps:
       - uses: actions/checkout@v4
```
McPatate:
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/CI.yml
```diff
@@ -92,25 +106,30 @@ jobs:
...
       - uses: actions/setup-python@v5
```
McPatate:
```suggestion
      - uses: actions/setup-python@v6
```
won't do them all, I'll let you figure that out!

bindings/python/py_src/tokenizers/__init__.pyi
```diff
@@ -606,6 +607,9 @@ class Tokenizer:
     def __repr__(self, /) -> str:
         """Return repr(self)."""
         ...
+    def __setattr__(self, /, attr: str, value: typing.Any) -> typing.Any:
```
McPatate: are we sure we want this to be the final API? Shouldn't we be exposing a dedicated method instead of raw `setattr`?

bindings/python/py_src/tokenizers/__init__.pyi
```diff
@@ -1195,6 +1199,24 @@ class Tokenizer:
     def pre_tokenizer(self, /, pretok: tokenizers.pre_tokenizers.PreTokenizer | None) -> None:
         """The `optional` :class:`~tokenizers.pre_tokenizers.PreTokenizer` in use by the Tokenizer"""
         ...
+    @property
+    def role_to_token(self, /) -> typing.Any:
```
McPatate:
```suggestion
    def role_to_token(self, /) -> typing.Optional[typing.Dict[str, str]]:
```
no?

bindings/python/py_src/tokenizers/__init__.pyi
```diff
@@ -1195,6 +1199,24 @@ class Tokenizer:
...
+            :obj:`Dict[str, str]` or :obj:`None`: The role to token mapping if set
+        """
+        ...
+    @role_to_token.setter
+    def role_to_token(self, /, role_to_token: typing.Any | None) -> None:
+        """
+        Get the role to token mapping
+
+        Returns:
+            :obj:`Dict[str, str]` or :obj:`None`: The role to token mapping if set
```
McPatate: Return None here no?

bindings/python/src/tokenizer.rs
```diff
@@ -612,6 +612,115 @@ impl PyTokenizer {
         PyTokenizer::from_model(model.clone())
     }
 
+    fn __getattr__(&self, py: Python<'_>, attr: &str) -> PyResult<Py<PyAny>> {
```
McPatate, line 645: not a fan of this, why not go through the dedicated method? is it standard to do `tokenizer.{token_name} = value`?

McPatate: also, could we not type the result correctly here instead of PyAny?

bindings/python/tests/bindings/test_tokenizer.py
```diff
@@ -755,14 +755,103 @@ def test_setting_to_none(self):
...
+        tokenizer.eos_token = "[|NEW_EOS_TOKEN|]"
+        assert tokenizer.eos_token == "[|NEW_EOS_TOKEN|]"
+        assert tokenizer.eos_token_id == 5
+        assert "[|NEW_EOS_TOKEN|]" in tokenizer.get_vocab(with_added_tokens=True)
+
+        assert tokenizer.decode([5], False) == "[|NEW_EOS_TOKEN|]"
+        tokenizer.unk_token = None
+        assert tokenizer.unk_token is None
+        assert tokenizer.unk_token_id is None
+        assert tokenizer.decode([3], False) == "<unk>"  # we don't delete tokens.
```
McPatate, line 850: why not?

tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -312,6 +312,9 @@ impl AddedVocabulary {
             // Finally add the token to the classic set if special
             if !self.special_tokens_set.contains(&token.content) {
```
McPatate, line 316: what happens if `token.special == true` and `!self.special_tokens_set.contains(&token.content) == true`?

bindings/python/tests/bindings/test_tokenizer.py
```diff
@@ -755,14 +755,103 @@ def test_setting_to_none(self):
...
+        }
+
+        # Test individual token properties
+        assert tokenizer.eos_token == "</s>"
+        assert tokenizer.bos_token == "<s>"
+        assert tokenizer.pad_token == "<pad>"
+        assert tokenizer.unk_token == "<unk>"
+
+        # Test token ID properties (tokens are added in order, so ids are 0, 1, 2, 3)
+        assert tokenizer.eos_token_id == 0
```
McPatate, line 825: why not:
```suggestion
        assert tokenizer.eos_token.id() == 0
```
instead of special string parsing logic that is prone to error and imo kind of confusing?

Issue #1784, opened 2025-05-27
Add cp314t tests, mark extension modules as compatible, and ship free-threaded wheels


ngoldbaum:
Now that https://github.com/huggingface/tokenizers/issues/1767 is closed, tokenizers builds under the free-threaded build.

`aiohttp` is an (indirect) dev dependency, and they're still working on support, so that might be a complication.

The extension module also still needs to be marked as compatible with the free-threaded build:

```
<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'tokenizers.tokenizers', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```

If I do `PYTHON_GIL=0 make test`, the tests all pass. That said, I don't see any explicitly multithreaded tests, so it might be a good idea to extend the multiprocessing tests to also use thread pools, to the extent that's feasible. It might also be a good idea to try [pytest-run-parallel](https://github.com/Quansight-Labs/pytest-run-parallel) to shake out any use of global state in the tests or implementation.

ngoldbaum:
I looked more at free-threaded support today. I looked a little at making the pyclasses in the python bindings `frozen` to opt out of runtime borrow-checking, but that is currently blocked on the extensive use of `__setstate__` internally. Since `__setstate__` takes `&mut self`, it's incompatible with a frozen pyclass. I think tokenizers would need to migrate to using e.g. `__reduce__` or `__getnewargs_eg__` to avoid that.

There are a few thread-unsafe tests with `pytest-run-parallel` due to the way the multiprocessing tests manipulate `os.environ`, some use of the filesystem that assumes no other instances of a test are running, and  a couple cases where a fixture that gets shared between threads and mutated. If I fix these issues or mark tests that have to use global state as thread-unsafe, then `pytest-run-parallel` passes.

I'm planning to open a PR shortly that adds 3.13t CI and runs the tests under pytest-run-parallel with PYTHON_GIL=0, along with a few minor fixes. We can then declare free-threaded support in the pymodules and add free-threaded wheel builds as a next step.

ngoldbaum:
> I'm planning to open a PR shortly

see #1809

ngoldbaum:
[...]

PR #1948, opened 2026-02-11
Update to PyO3 0.28 to automatically disable GIL

ngoldbaum:
Towards fixing #1784.

For the most part, tokenizers relies on PyO3 for thread safety. There are some constructs that aren't allowed for pymodules that set `gil_used = false`, but tokenizers doesn't use any of them.

`RefMutContainer` does have some `unsafe` in its implementation, but @davidhewitt looked at it and doesn't see issues (see e.g https://github.com/PyO3/pyo3/issues/5664).

PyO3 0.28 made `gil_used = false` the default for PyModules, so there's no need to explicitly mark them as not relying on the GIL.

The only nontrivial changes in this update were around the migration described here: https://pyo3.rs/v0.28.0/migration.html#deprecation-of-automatic-frompyobject-for-pyclass-types-which-implement-clone. I added `skip_from_py_object` where the current implementation wasn't relying on the automatically derived `FromPyObject` implementation and otherwise added `from_py_object` to all the pyclasses that implement `Clone`. Let me know if you'd prefer to keep the old behavior for all the pyclasses instead.

See passing CI run from my fork: https://github.com/ngoldbaum/tokenizers/actions/runs/21910929364?pr=5. I don't see any spots where the GIL is getting enabled at runtime in the 3.14t jobs.

### DIFF
--- a/bindings/python/Cargo.toml
+++ b/bindings/python/Cargo.toml
@@ -29,7 +29,7 @@ path = "../../tokenizers"
 
 [dev-dependencies]
 tempfile = "3.10"
-pyo3 = { version = "0.28.0", features = ["auto-initialize"] }
+pyo3 = { version = "0.28.2", features = ["auto-initialize"] }
 
 
 [features]


ngoldbaum:
> I'd rather we keep from_py_object for all types

No problem, see latest push.

ngoldbaum:
PyO3 0.28.1 came out over the weekend, so I just pushed a commit to use that instead of 0.28.0.

ngoldbaum:
@ArthurZucker can you re-trigger CI over here? I don't think the failures are related.

ngoldbaum:
Here's a passing CI run I just triggered on my fork, FWIW: https://github.com/ngoldbaum/tokenizers/actions/runs/21913362339

ngoldbaum:
I just pushed another commit to move to PyO3 0.28.2, which came out this morning. The musllinux builds were failing on my fork but the CI affected here all passed.

ArthurZucker:
I'll try to fix the musllinux on my side and push to main

ArthurZucker:
before I can release I just need to merge #1942 !

ngoldbaum:
Maturin just did a release that should fix this.

I also still need to add 3.14t wheels to the wheel-building workflow. I'll try to get a PR open for that so it doesn't block a release. It'd be really nice to have 3.14t wheels on PyPI!

ngoldbaum:
Looks like it is indeed fixed: https://github.com/ngoldbaum/tokenizers/actions/runs/22159130634?pr=8

ArthurZucker:
Thanks again for the PR!

PR #1949, opened 2026-02-13
Fix identity comparison in toctree_tags.py

llukito:
Replaced `!= None` with `is not None` in `docs/source/_ext/toctree_tags.py`.

Comparisons to singletons like `None` should always be done with `is` or `is not`, never the equality operators (PEP 8 recommendations).


llukito:
Hi, just checking in on this, it’s ready for review whenever someone has a moment. Thanks!

ArthurZucker (approved):
Sure

PR #1951, opened 2026-02-19
Add riscv64 build, make Linux wheel build matrix more explicit

threexc:
maturin supports riscv64 cross build, but the arch string and the target toolchain for rustc don't exactly match up (this is also the case for other architectures like ppc64le and x86). To make the riscv64 addition consistent with the others, change the platform list so that arch and target are distinct fields.

Note that this is being done on behalf of the [RISE Project](https://riseproject.dev/) to improve Python ecosystem support on riscv64 platforms.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -42,6 +42,9 @@ jobs:
           - runner: ubuntu-latest
             arch: ppc64le
             target: powerpc64le-unknown-linux-gnu
+          - runner: ubuntu-latest
+            arch: riscv64
+            target: riscv64gc-unknown-linux-gnu
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5


threexc:
Not sure who to ping for some attention on this PR. Maybe @ArthurZucker ? I also have one for [hf_transfer](https://github.com/huggingface/hf_transfer/pull/77) and another for [safetensors](https://github.com/huggingface/safetensors/pull/708)

McPatate:
There's an issue with the CI, fix should be coming soon

#1978

threexc:
Rebased on latest main.

ArthurZucker:
Yeah sorry for the delay

threexc:
> Yeah sorry for the delay

Not a problem, I waited for the fix in #1978. Thanks for looking at it!

threexc:
> LGTM happy to have coverage for riscv64!

Thanks!

PR #1952, opened 2026-02-19
🚨 Update release workflow for Python 3.14, dropping 3.9

ngoldbaum:
See https://github.com/ngoldbaum/tokenizers/pull/9 where I hacked this together on my fork to test.

See https://github.com/ngoldbaum/tokenizers/actions/runs/22193203602/job/64186704987?pr=9 for the python-release workflow. I verified that all workflows are generating an abi3 and cp314t wheel.

maturin doesn't support Python 3.9 or pypy 3.9 and pypy 3.10 anymore. They *do* support pypy3.11, so I could add those builds if you want.

Pypy is not actively developed anymore, so from my perspective it doesn't make sense to start shipping pypy3.11 wheels only to need to stop shipping them next year when maturin drops support for pypy entirely.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -146,6 +146,11 @@ jobs:
           target: ${{ matrix.platform.target }}
           args: --release --out dist --manifest-path bindings/python/Cargo.toml
           sccache: 'true'
+        env:
+          # abi3 extension modules resolve Python symbols at runtime;
+          # explicit --target triggers cross-compilation mode which
+          # doesn't pass this flag automatically on macOS.
+          RUSTFLAGS: "-C link-arg=-undefined -C link-arg=dynamic_lookup"
       - name: Upload wheels
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02  # v4
         with:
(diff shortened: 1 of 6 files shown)


ngoldbaum:
> we are dropping support for py39
> but we still have support for 310->314 since we use abi310
?

Exactly! See https://pyo3.rs/v0.28.2/building-and-distribution.html#minimum-python-version-for-abi3

ngoldbaum:
Also, for what it's worth, I'm looking at adding a section to [PEP 803](https://peps.python.org/pep-0803/) with quotes from library maintainers. 

If you wouldn't mind, I'd appreciate it if I could include your opinion about the need for a stable ABI that supports the free-threaded build. Accepting PEP 803 will ensure that such a thing exists for Python 3.15 and I'd like to make that case as strong as possible.

ngoldbaum:
It looks like the tests failed due to an issue conntecting with Peter Norvig's blog.

@ArthurZucker let me know when you're ready to prepare a new tokenizers release and I'll update this PR.

ArthurZucker:
I am hoping for EOW

ArthurZucker:
Update sorry will do one tomorrow / today but want to drop py3.9 in the next

ngoldbaum:
Let me know if you need me to do anything here!

ArthurZucker:
Actually we'll drop it for 0.23!

PR #1965, opened 2026-03-17
Fix type_ids not applied to overflow encodings

joaquinhuigomez:
When using TemplateProcessing with type_id mappings, type_ids were only applied to the main encoding but not to overflow encodings from truncation. This fix iterates over overflow encodings in apply_template and sets both type_ids and sequence_ids consistently.


joaquinhuigomez:
Hi — just checking if this is still on the review queue. Happy to address any feedback.

joaquinhuigomez:
Closing — no review activity in a few weeks and CI doesn't run on forks here, so signal is limited. The fix is small if anyone wants to pick it up.

PR #1966, opened 2026-03-19
feat: add async decode batch type annotations in decoding

michaelfeil:
Something I forgot to add to https://github.com/huggingface/tokenizers/pull/1843. 
Reminder to myself to also wire up native asynchronous support.

### DIFF
--- a/bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
+++ b/bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
@@ -336,6 +336,24 @@ class BaseTokenizer:
 
         return self._tokenizer.decode_batch(sequences, skip_special_tokens=skip_special_tokens)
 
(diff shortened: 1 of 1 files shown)


michaelfeil:
@ArthurZucker this should be an easy one to review. TY.

michaelfeil:
Seems like a unrelated failure

ArthurZucker:
Yep!

Qubitium:
Hi @ArthurZucker @michaelfeil I just ran into this PR and am curious at the structure: why is BaseTokenizer another wrapper which holds the real tokenzer in `_tokenizer`? Since the naming and the code is confusing me a little. It's been a long time since I checked the internals of hf tokenizers so I am really clueless and want to start.

ArthurZucker:
its not an internal its more like an example of how you can use. I don't think many people use it tho

PR #1967, opened 2026-03-19
feat: Optimize BPE tokenization: sharded cache, packed merge keys, FxHash (10-15% speedup)

michaelfeil:
Tips and tricks from a Blog post describing how this lib is 4-10x slower than it needs to be. https://www.crusoe.ai/resources/blog/reducing-ttft-by-cpumaxxing-tokenization

## Performance: sharded cache, packed merge keys, FxHash, byte table

Benchmarked on LLaMA 3 tokenizer (`data/llama-3-tokenizer.json`) with `data/big.txt` (6.5 MB, 128K lines).

### Setup

- **Baseline:** `main` branch (onig regex, single `RwLock<AHashMap>` cache, `AHashMap<(u32,u32)>` merge map)
- **Test:** `mf/performance-improvements` branch
- **Hardware:** Linux, Intel Xeon
- **Benchmark:** `cargo bench --bench llama3_benchmark`

### Results

| Benchmark | main | optimized | Change |
|---|---|---|---|
| Offsets (batch + char offsets) | 298 ms | 226 ms | **-24.1%** |
| Batch encode (1000 items) | 737 ms | 594 ms | **-19.3%** |
| Sequential encode (full file) | 1.82 s | 1.80 s | -1.0% (noise) |
| BPE Train | ~10 s | ~11 s | no change (noise) |

### What changed

1. **Sharded cache** (`cache.rs`): Replaced single `RwLock<AHashMap>` with 64-shard `RwLock<FxHashMap>`. Eliminates lock contention in parallel `encode_batch` / rayon workloads.

2. **Packed u64 MergeMap** (`mod.rs`): BPE merge lookup keys packed from `(u32, u32)` into single `u64`. FxHash on one u64 is a single multiply vs hashing two fields separately.

3. **FxHash** (`Cargo.toml`, `word.rs`): `rustc-hash` for merge map and cache — faster non-cryptographic hashing for small integer and string keys.

4. **Flat byte-to-char table** (`byte_level.rs`): Pre-computed `[char; 256]` lookup replaces `AHashMap<u8, char>` in the byte-level encoding hot path.

### Why sequential encode barely moved

The sequential path is dominated by the regex step (~40-50% of time) and per-split `NormalizedString` allocations (~20-30%). These changes target the merge lookup and cache layers, which matter most in parallel/batch workloads where lock contention was the bottleneck.

### Files changed

| File | Change |
|---|---|
| `Cargo.toml` | Added `rustc-hash = "2"` |
[...]

PR #1968, opened 2026-03-19
feat: performance, adding pcre2 backend + regex-shards (5-15% speedup)

michaelfeil:
Tips and tricks from a Blog post describing how this lib is 4-10x slower than it needs to be. https://www.crusoe.ai/resources/blog/reducing-ttft-by-cpumaxxing-tokenization

## Performance: PCRE2 JIT regex backend

Benchmarked on LLaMA 3 tokenizer (`data/llama-3-tokenizer.json`) with `data/big.txt` (6.5 MB, 128K lines).

### Setup

- **Baseline:** `main` branch with default `onig` (Oniguruma) regex backend
- **Test:** same branch with `--no-default-features --features pcre2`
- **Hardware:** Linux, Intel Xeon
- **Benchmark:** `cargo bench --bench llama3_benchmark`

### Results

| Benchmark | onig (main) | pcre2 JIT | Change |
|---|---|---|---|
| Sequential encode (full file) | 1.82 s | 1.63 s | **-10 to -14%** |
| Batch encode (1000 items) | 737 ms | 658 ms | **~-5 to -11%** |
| Offsets (batch + char offsets) | 295 ms | 288 ms | ~-3% |
| BPE Train | ~10 s | ~10 s | no change |

Sequential encode is the most stable measurement. Batch/offsets show improvement but have higher run-to-run variance due to thread scheduling.

### Why

PCRE2 with JIT compiles regex patterns to native machine code at initialization. The GPT-2 byte-level pre-tokenization regex (`'s|'t|'re|'ve|'m|'ll|'d| ?\p{L}+| ?\p{N}+| ?[^\s\p{L}\p{N}]+|\s+(?!\S)|\s+`) is the hot path for every `encode()` call and benefits directly from JIT compilation vs Oniguruma's interpreted matching.

### Usage

```toml
# Cargo.toml
[dependencies]
tokenizers = { version = "0.22", features = ["pcre2"] }
```

The `pcre2` feature is opt-in. When enabled, it takes priority over `onig` and `fancy-regex`. Cross-compilation works the same as `onig` — `pcre2-sys` bundles and compiles PCRE2 from C source via the `cc` crate. Not available on WASM (use `fancy-regex` there, as before).

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,4 @@
+[env]
+# Only takes effect when the optional `pcre2` feature is enabled and `pcre2-sys`
+# is built. Keeps runtime free of dynamic libpcre2 dependency.
+PCRE2_SYS_STATIC = { value = "1", force = false }
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -81,6 +81,7 @@ compact_str = { version = "0.9", features = ["serde"] }
 
 [features]
 default = ["progressbar", "onig", "esaxx_fast"]
+pcre2 = ["dep:pcre2", "fancy-regex"]
 esaxx_fast = ["esaxx-rs/cpp"]
 progressbar = ["indicatif"]
 http = ["hf-hub"]
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -75,6 +75,7 @@ getrandom = { version = "0.3" }
 esaxx-rs = { version = "0.1.10", default-features = false, features=[]}
 monostate = "0.1.12"
 ahash = { version = "0.8.11", features = ["serde"] }
+pcre2 = { version = "0.2", optional = true }
 dary_heap = { version = "0.3.6", features = ["serde"] }
 compact_str = { version = "0.9", features = ["serde"] }
 
--- a/tokenizers/src/utils/fancy.rs
+++ b/tokenizers/src/utils/fancy.rs
@@ -18,6 +18,30 @@ impl SysRegex {
             regex: Regex::new(regex_str)?,
         })
     }
+
+    pub fn find_matches(
+        &self,
+        inside: &str,
+    ) -> Result<Vec<(Offsets, bool)>, Box<dyn Error + Send + Sync + 'static>> {
+        if inside.is_empty() {
+            return Ok(vec![((0, 0), false)]);
+        }
+
+        let mut prev = 0;
+        let mut splits = Vec::with_capacity(inside.len());
+        for matched in self.regex.find_iter(inside) {
+            let matched = matched?;
+            if prev != matched.start() {
+                splits.push(((prev, matched.start()), false));
+            }
+            splits.push(((matched.start(), matched.end()), true));
+            prev = matched.end();
+        }
+        if prev != inside.len() {
+            splits.push(((prev, inside.len()), false));
+        }
+        Ok(splits)
+    }
 }
 
 pub struct Matches<'r, 't>(fancy_regex::Matches<'r, 't>);
--- a/tokenizers/src/utils/onig.rs
+++ b/tokenizers/src/utils/onig.rs
@@ -20,6 +20,26 @@ impl SysRegex {
             regex: Regex::new(regex_str)?,
         })
     }
+
+    pub fn find_matches(&self, inside: &str) -> Result<Vec<(Offsets, bool)>> {
+        if inside.is_empty() {
+            return Ok(vec![((0, 0), false)]);
+        }
+
+        let mut prev = 0;
+        let mut splits = Vec::with_capacity(inside.len());
+        for (start, end) in self.regex.find_iter(inside) {
+            if prev != start {
+                splits.push(((prev, start), false));
+            }
+            splits.push(((start, end), true));
+            prev = end;
+        }
+        if prev != inside.len() {
+            splits.push(((prev, inside.len()), false));
+        }
+        Ok(splits)
+    }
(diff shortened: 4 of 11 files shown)


wheynelau:
@michaelfeil Thanks for providing more details! I didn't know that fasttokens used pcre2 as well, just learnt about it yesterday.

wheynelau:
Curious as you are active on this, have you tried thread local caches to see if they help? Or other distributed caches like dashmap and parking lot? Thank you!

michaelfeil:
@wheynelau thats the other PR. https://github.com/huggingface/tokenizers/pull/1967 

I think thread_local caches are most useful if re-used in a rayon pool, or if the threads are in some kind of pool. 
For the L1 in the blog post, I tried to implement something, and gave it to claude, but both did not work. Implementing the shared lookup is a bit better IMO, since hashing is fast + you only need to loopup by hash, so the only place that uses RwLock is guarded.

tokenizers/src/utils/fancy.rs
```diff
@@ -18,6 +18,30 @@ impl SysRegex {
             regex: Regex::new(regex_str)?,
         })
     }
+
+    pub fn find_matches(
+        &self,
+        inside: &str,
+    ) -> Result<Vec<(Offsets, bool)>, Box<dyn Error + Send + Sync + 'static>> {
```
McPatate, line 25: this doesn't need to be a `Result` or am I missing something?

.cargo/config.toml
```diff
@@ -0,0 +1,4 @@
+[env]
+# Only takes effect when the optional `pcre2` feature is enabled and `pcre2-sys`
+# is built. Keeps runtime free of dynamic libpcre2 dependency.
```
McPatate, line 3: > and `pcre2-sys` is built

which is always?

tokenizers/src/utils/onig.rs
```diff
@@ -20,6 +20,26 @@ impl SysRegex {
             regex: Regex::new(regex_str)?,
         })
     }
+
+    pub fn find_matches(&self, inside: &str) -> Result<Vec<(Offsets, bool)>> {
```
McPatate, line 24: same here for the `Result` ?

tokenizers/src/utils/pcre2_backend.rs
```diff
@@ -0,0 +1,507 @@
...
+}
+
+/// A single PCRE2 regex instance. Each instance maintains its own DFA cache,
+/// so sharing across threads causes cache thrashing.
+struct Pcre2Regex {
+    inner: pcre2::bytes::Regex,
+}
+
+impl Pcre2Regex {
+    fn compile(pattern: &str) -> Result<Self, Box<dyn Error + Send + Sync + 'static>> {
```
McPatate, line 54: I think we have the dependency on `thiserror`, would much rather have a proper defined error type

tokenizers/src/utils/pcre2_backend.rs
```diff
@@ -0,0 +1,507 @@
...
+    fn find_at(&self, text: &[u8], offset: usize) -> Option<(usize, usize)> {
+        match self.inner.find_at(text, offset) {
+            Ok(Some(m)) => Some((m.start(), m.end())),
+            _ => None,
+        }
+    }
+}
+
+// Safety: pcre2::bytes::Regex is Send+Sync. Each Pcre2Regex instance has its
+// own match context, so concurrent find_at calls on *different* instances are safe.
```
McPatate, line 72: But compile is not thread safe I assume?

tokenizers/src/utils/pcre2_backend.rs
```diff
@@ -0,0 +1,507 @@
...
+unsafe impl Sync for Pcre2Regex {}
+
+/// PCRE2-backed regex with JIT compilation and per-thread copies.
+///
+/// Pre-compiles a pool of independent PCRE2 regex instances at construction time
+/// (capped at 32). Each thread picks its own copy via a stable thread-local index,
+/// avoiding DFA cache contention under concurrent use.
+///
+/// Falls back to `fancy_regex` at runtime if PCRE2 compilation fails for a
+/// particular pattern.
```
McPatate:
```suggestion
/// particular regex pattern.
```

tokenizers/src/utils/pcre2_backend.rs
McPatate: or perhaps at least `use crate::Result` instead

tokenizers/Cargo.toml
```diff
@@ -75,11 +75,13 @@ getrandom = { version = "0.3" }
 esaxx-rs = { version = "0.1.10", default-features = false, features=[]}
 monostate = "0.1.12"
 ahash = { version = "0.8.11", features = ["serde"] }
+pcre2 = { version = "0.2", optional = true }
 dary_heap = { version = "0.3.6", features = ["serde"] }
 compact_str = { version = "0.9", features = ["serde"] }
 
 [features]
 default = ["progressbar", "onig", "esaxx_fast"]
+pcre2 = ["dep:pcre2", "fancy-regex"]
```
McPatate, line 84: why is one declared with the `dep:` prefix and not the other?

tokenizers/src/utils/pcre2_backend.rs
McPatate, line 1: I'd add at least one extra test to make sure the fallback logic is sane here.

McPatate (commented):
Good stuff! Didn't repro the benchmarks yet, but looking good so far! Thank you for taking the time.

Issue #1573, opened 2024-07-19
Truncation works slowly. Tokenizer firstly encodes long sequence and then truncate it.


galtimur:
I work with very long sequences, which I want to truncate to a shorter one in the pre-processing stage.
I found that the Tokenizer first encodes the whole sequence whole, after that truncates it. Is there a way to truncate it on the fly, to save time on tokenization of unused part of the sequence?

I use transformers 4.42.4.
tokenizer class is LlamaTokenizerFast

The code and the output:

```
from transformers import AutoTokenizer
from time import time

tokenizer = AutoTokenizer.from_pretrained("deepseek-ai/deepseek-coder-1.3b-base")
# tokenizer class is LlamaTokenizerFast
def encode_long(meta_len, batch_size=200):
    text = 500*"jibber-jabber"
    text_batch = batch_size*[meta_len*text]

    start_time = time()
    batch = tokenizer.batch_encode_plus(text_batch, truncation=True, max_length=1000)
    print(f"Seq encoded len = {len(batch['input_ids'][0])}; Time used = {time() - start_time:.2f}, Symbols = {meta_len*len(text)}")

encode_long(meta_len=1)
encode_long(meta_len=10)
encode_long(meta_len=100)
```

Results of the code on my machine:

Seq encoded len = 1000; Time used = 0.07, Symbols = 6500
Seq encoded len = 1000; Time used = 0.59, Symbols = 65000
Seq encoded len = 1000; Time used = 8.98, Symbols = 650000

[Colab ](https://colab.research.google.com/drive/1HkEBto8wgL4MbBp4dqZJ0hH_6KNUS1Ka?usp=sharing) for reproduction.

ArthurZucker:
Hey! There is currently no way to do that because what you want sounds more like truncate the text directly no? 
We are gonna improve the throughput in general, but you need to encode first, otherwise what you need is just cutting your texts to the correct length no?

galtimur:
> Hey! There is currently no way to do that because what you want sounds more like truncate the text directly no?

Yes, that's true. I overcome this issue by first truncating the text directly. To do so I have to assess the average length of the token with some extra margin.
I thought that it would be natural to truncate sequences right in the tokenization loop, breaking it when the desired number of tokens is reached.

komitydev:
[...]

PR #1990, opened 2026-03-26
feat(tokenizer): add early exit for truncation

McPatate:
Start tackling #1573.

Here we skip tokenizing remaining splits after the `max_length` of the `TruncationParams` has been reached in the case `TruncationDirection::Right`.

This leaves further questions as to how we tackle shortening the pre-tokenization pipeline as well as the tokenization. Since here we introduce the bound check on the token count, we still do the full context's pretok before the token count skip.
In the benchmark I've left for reproduction in case that is of interest to anyone, there's a `pre-truncated-input-512` bench that is orders of magnitude faster, because it cuts the actual input text before passing it to the tokenizer.

I also only handled right truncation as it's the simplest, need to figure out how to do left tokenization. I think I'll tackle both topics in subsequent PRs.

Benchmark improvement results (if unspecified, `direction == right`):

| Benchmark | Before | After | Change |
  |---|---|---|---|
  | **truncate-to-512 (full corpus)** | 1.01 s | 732 ms | **-28%** |
  | **scaling: 10K chars** | 1.26 ms | 1.06 ms | **-16%** |
  | **scaling: 100K chars** | 13.2 ms | 10.5 ms | **-21%** |
  | **scaling: 500K chars** | 71.3 ms | 56.6 ms | **-21%** |
  | **max_length=128** | 6.63 ms | 5.21 ms | **-22%** |
  | **max_length=512** | 6.62 ms | 5.30 ms | **-20%** |
  | **max_length=2048** | 6.77 ms | 5.46 ms | **-20%** |
  | **max_length=8192** | 6.74 ms | 6.00 ms | **-11%** |
  | **right direction** | 1.27 ms | 1.06 ms | **-16%** |
  | **left direction** | 1.27 ms | 1.27 ms | no change (expected) |
  | **no truncation** | 975 ms | 971 ms | no change (expected) |

### DIFF
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -1674,4 +1674,104 @@ mod tests {
             "Left-truncated should match last 3 tokens of full encoding"
         );
     }
+
+    #[test]
+    fn pair_right_truncation_longest_first() {
+        // Seq A: "a b c d e f g h i j" → 10 tokens
+        // Seq B: "a b c d e"           → 5 tokens
+        // Total: 15, max_length: 6, LongestFirst → should trim A more than B
+        let seq_a = "a b c d e f g h i j";
+        let seq_b = "a b c d e";
+
+        let mut tok = test_tokenizer();
+        tok.with_truncation(Some(TruncationParams {
+            max_length: 6,
+            strategy: TruncationStrategy::LongestFirst,
+            stride: 0,
+            direction: TruncationDirection::Right,
+        }))
+        .unwrap();
+        let truncated = tok.encode((seq_a, seq_b), false).unwrap();
+
+        assert_eq!(
+            truncated.get_ids().len(),
+            6,
+            "Pair encoding should have exactly max_length tokens"
+        );
+        // First N tokens should match the start of the full encoding
+        assert_eq!(
+            &truncated.get_ids()[..truncated.get_ids().len()],
+            &[0, 1, 2, 0, 1, 2],
+            "Truncated pair should match prefix of full pair encoding"
+        );
+    }
+
+    #[test]
+    fn pair_only_second_does_not_truncate_first() {
+        // Seq A: "a b c d e" → 5 tokens
+        // Seq B: "a b c d e f g h i j" → 10 tokens
+        // max_length: 8, OnlySecond → A stays at 5, B truncated to 3
+        let seq_a = "a b c d e";
+        let seq_b = "a b c d e f g h i j";
+
+        let mut tok = test_tokenizer();
+        tok.with_truncation(Some(TruncationParams {
+            max_length: 8,
+            strategy: TruncationStrategy::OnlySecond,
+            stride: 0,
+            direction: TruncationDirection::Right,
+        }))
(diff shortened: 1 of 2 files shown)


McPatate:
So actually implementing left truncation support was pretty easy, yields a perf boost consitent with the right truncation of ~17%:

| Benchmark | Before | After | Change |
|---|---|---|---|
| **left direction** | 1.27 ms | 1.06 ms | **-17%** |

tokenizers/src/tokenizer/pre_tokenizer.rs
```diff
@@ -126,6 +127,66 @@ impl PreTokenizedString {
...
+                let mut first_tokenized_idx = self.splits.len();
+
+                for (i, split) in self.splits.iter_mut().enumerate().rev() {
+                    if let Some(tokens) = &split.tokens {
+                        total_tokens += tokens.len();
+                        first_tokenized_idx = i;
+                        continue;
+                    }
+
+                    let tokens = tokenize(&split.normalized)?;
```
ArthurZucker, line 152: If you have 1 split, HUGE ass one -> slow -> chinese don't benefit AFAIK. 
If you have many small splits -> normal
If you have many big splits -> huge wins (after 2 splits boom your done).

tokenizers/Cargo.toml
ArthurZucker, line 1: remove unrelated styling my lord

tokenizers/src/tokenizer/pre_tokenizer.rs
```diff
@@ -126,6 +127,66 @@ impl PreTokenizedString {
         Ok(())
     }
 
+    pub fn tokenize_with_limit<F>(
```
ArthurZucker, line 131:
```suggestion
    /// Early quit if we have reach the token limit!
    pub fn tokenize_with_limit<F>(
```

tokenizers/src/tokenizer/mod.rs
```diff
@@ -1579,3 +1592,86 @@ where
...
+            .build()
+            .unwrap();
+
+        let mut tokenizer = Tokenizer::new(model);
+        tokenizer.with_pre_tokenizer(Some(WhitespaceSplit));
+        tokenizer
+    }
+
+    #[test]
+    fn right_truncation_early_exit_matches_full_encode() {
```
ArthurZucker, line 1624: missing test with 2 inputs (encodings) as in https://github.com/huggingface/tokenizers/blob/af00f329aa499b3efa02d44c217c7a74023dc6c2/tokenizers/src/utils/truncation.rs#L165

tokenizers/Cargo.toml
McPatate: Could we just leave it? imo it's cleaner this way and I will have to revert fmt every time on my end which is annoying 🤓

tokenizers/benches/truncation_benchmark.rs
McPatate, line 1: @ArthurZucker should I keep this bench file or no?

ArthurZucker: yes pls

Issue #849, opened 2021-12-16
Sampling alternative tokenizations


david-waterworth:
In [this](https://colab.research.google.com/github/huggingface/notebooks/blob/master/transformers_doc/tokenizer_summary.ipynb#scrollTo=QCkb-TFs1Wdu) colab it mentions:

> Unigram saves the probability of each token in the training corpus on top of saving the vocabulary so that the probability of each possible tokenization can be computed after training. The algorithm simply picks the most likely tokenization in practice, but also offers the possibility to sample a possible tokenization according to their probabilities.

I'm sure I saw another post that described how to do this but I cannot find it any more - is this possible? i.e. instead of simply calling encode and getting the `best` tokenization I want to select the top-k (and their scores)

Narsil:
hi @david-waterworth, this isn't enabled within this library (it is only possible to do with `Unigram` models anyway).

In `transformers` if you use the `slow` tokenizer, you can pass `sp_model_kwargs` to enable sampling:

```python
AlbertTokenizer(...., sp_model_kwargs={"enable_sampling": True})
```

https://github.com/huggingface/transformers/blob/master/src/transformers/models/albert/tokenization_albert.py#L105

That will use `sentencepiece` under the hood which will just sample from desired sampling policy `NBest` seems to also enable do `top_k`.

PRs welcome to add it to `tokenizers` (the naming is very close to `sentencepiece` so it should be doable).

david-waterworth:
Thanks @Narsil, I'll take a look at the `albert` example and see if I have any bandwidth to implement a unigram version.

RuiWang1998:
Hi, 

Sorry to bother you but if I have trained a Unigram tokenizer using its `UnigramTrainer`, am I correct to understand that the values following the token in the json file is its log-probability?

I followed https://huggingface.co/docs/tokenizers/training_from_memory

Or how best should I interpret the value?


Thanks in advance.

Narsil:
Yes that's something like log probability.

kellymarchisio:
Seconding this request for sampling with Unigram

Narsil:
Unfoprtunately, activity on this is issue seems to suggest this is a very low demand feature, very low probability to see the light anytime soon. 

I'm going to close it. If there's some interest we might reopen.

kellymarchisio:
> Unfoprtunately, activity on this is issue seems to suggest this is a very low demand feature, very low probability to see the light anytime soon.
> 
[...]

Issue #730, opened 2021-06-06
Subword regularization


nikitakit:
As far as I can tell, this library doesn't currently implement [subword regularization](https://arxiv.org/abs/1804.10959).

I've found BPE dropout to be quite helpful when using BPE-based models like GPT-2, and I'd like to have the option to try subword regularization for Unigram models (like T5) as well.

I should note that I'm using pre-trained models via the `transformers` package, and my code requires offset mappings for aligning subword tokens with characters in the input string. Even though the so-called "slow" tokenizers support subword regularization, their lack of offset mapping capabilities is a non-starter for my application.

Narsil:
Hi, 

`sample` algorithm is indeed not implemented, it should be however relatively simple to do (just look at sentencepiece code, we have a very similar structure here), PRs welcome.

PR #1994, opened 2026-03-27
Add unigram sampling (alpha, nbest_size)

kennethsible:
I noticed that `models.Unigram` doesn't support sampling, which enables subword regularization (arguably one of the main reasons to choose the Unigram model). I checked GitHub and there are multiple closed issues on this topic (#730, #849). In one of these issues, it was mentioned that the sampling code has already been implemented in `lattice.rs` and simply needs to be exposed through Python. I filled in the missing details and added a `sample_nbest` function for parity with Google's implementation. I also copied the interface for BPE dropout as closely as possible, including getters and setters for the new sampling parameters.

### DIFF
--- a/bindings/python/src/models.rs
+++ b/bindings/python/src/models.rs
@@ -932,7 +932,7 @@ impl PyUnigram {
 
     #[new]
     #[pyo3(
-        signature = (vocab=None, unk_id=None, byte_fallback=None, alpha=None, nbest_size=None), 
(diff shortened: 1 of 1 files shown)


kennethsible:
> Also update the doc of PyUnigram to mention these new args and the new behavior!

@ArthurZucker Is updating the docstring sufficient? I wasn't sure if the documentation is autogenerated.

PR #1999, opened 2026-03-30
perf: replace aho-corasick with daachorse for added vocabulary matching

ArthurZucker:
## Summary

Replaces the `aho-corasick` crate with [`daachorse`](https://github.com/daac-tools/daachorse) for multi-pattern string matching in `AddedVocabulary`.

- daachorse implements a **Double-Array Aho-Corasick** automaton, which is significantly faster and more memory-efficient than the standard linked-list-based implementation
- Token IDs are stored directly as values in the automaton via `build_with_values`, eliminating the separate `Vec<u32>` indirection on every match lookup
- Empty tries are represented as `None` (daachorse requires at least 1 pattern), which also fast-paths the common case of no added tokens

## Changes

- `tokenizers/Cargo.toml`: swap `aho-corasick = "1.1"` → `daachorse = "1.0.0"`
- `tokenizers/src/tokenizer/added_vocabulary.rs`: update import, `MatchingSet` type, trie construction, and match iteration to use daachorse API

### DIFF
--- a/tokenizers/src/tokenizer/added_vocabulary.rs
+++ b/tokenizers/src/tokenizer/added_vocabulary.rs
@@ -345,20 +345,31 @@ impl AddedVocabulary {
             .partition(|(token, _)| token.normalized);
 
         let (tokens, ids): (Vec<&AddedToken>, Vec<u32>) = non_normalized.into_iter().unzip();
-        self.split_trie = if tokens.is_empty() {
+        // Deduplicate patterns by content (keeping first occurrence) and filter empty strings
+        // to avoid DuplicatePattern / ZeroLengthPattern errors from daachorse.
+        let mut seen: AHashSet<&str> = AHashSet::new();
+        let (deduped_patterns, deduped_ids): (Vec<&str>, Vec<u32>) = tokens
+            .iter()
+            .map(|t| t.content.as_str())
+            .zip(ids.iter().copied())
+            .filter(|(content, _)| !content.is_empty() && seen.insert(content))
+            .unzip();
+        self.split_trie = if deduped_patterns.is_empty() {
             None
         } else {
-            Some(
-                DoubleArrayAhoCorasickBuilder::new()
-                    .match_kind(MatchKind::LeftmostLongest)
-                    .build_with_values(
-                        tokens
-                            .iter()
-                            .map(|token| &token.content)
-                            .zip(ids.iter().copied()),
-                    )
-                    .expect("Failed to build trie when refreshing tokens"),
-            )
+            match DoubleArrayAhoCorasickBuilder::new()
+                .match_kind(MatchKind::LeftmostLongest)
+                .build_with_values(deduped_patterns.into_iter().zip(deduped_ids))
+            {
+                Ok(trie) => Some(trie),
+                Err(e) => panic!(
+                    "Failed to build trie when refreshing tokens: {}. \
+                     This is likely because the total size of all patterns exceeded the \
+                     double-array capacity (~4 GB). Consider reducing the number or size \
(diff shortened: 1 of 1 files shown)


ArthurZucker:
## Benchmark Results: daachorse vs main

Measured on 100,000 added tokens, deserialization from file (`Tokenizer::from_file`).

### vs `main` baseline

| Benchmark | main | this PR (daachorse) | speedup |
|---|---|---|---|
| special tokens, no normalizer | 393 ms | 233 ms | **1.69×** |
| **non-special tokens, no normalizer** | **7,567 ms** | **257 ms** | **29.4×** |
| special tokens, NFKC | 392 ms | 232 ms | **1.69×** |
| **non-special tokens, NFKC** | **7,611 ms** | **267 ms** | **28.5×** |

Swapping aho-corasick for daachorse (Double-Array Aho-Corasick) yields a **29× speedup** on non-special token deserialization. The Double-Array construction is dramatically faster for large vocabularies, and storing token IDs directly as automaton values eliminates the `Vec<u32>` indirection on every match.

tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -359,11 +372,21 @@ impl AddedVocabulary {
...
+            Some(
+                DoubleArrayAhoCorasickBuilder::new()
+                    .match_kind(MatchKind::LeftmostLongest)
+                    .build_with_values(
+                        patterns
+                            .iter()
+                            .map(|content| content.get())
+                            .zip(nids.iter().copied()),
+                    )
+                    .expect("Failed to build trie when refreshing tokens (normalized)"),
```
McPatate: Could be nice to handle error here no?

tokenizers/src/tokenizer/added_vocabulary.rs
```diff
@@ -342,11 +345,21 @@ impl AddedVocabulary {
...
+            Some(
+                DoubleArrayAhoCorasickBuilder::new()
+                    .match_kind(MatchKind::LeftmostLongest)
+                    .build_with_values(
+                        tokens
+                            .iter()
+                            .map(|token| &token.content)
+                            .zip(ids.iter().copied()),
+                    )
+                    .expect("Failed to build trie when refreshing tokens"),
```
McPatate: Handle err properly?

tokenizers/src/tokenizer/added_vocabulary.rs
McPatate: and here

odashi:
Hey all, thank you for taking interests about daachorse!
daachorse authors (@vbkaisetsu, @kampersanda and me) are currently working in the same entity, and the libray is maintained as usual although there's no major update.
If there's anything we need to develop, we'd be happy for working with you.

tokenizers/src/tokenizer/added_vocabulary.rs
vbkaisetsu: Possible error cases here include: (1) zero-length patterns, (2) duplicate patterns, and (3) the double-array length exceeding the limit (approx. 4GB).
https://docs.rs/daachorse/latest/daachorse/bytewise/struct.DoubleArrayAhoCorasick.html#errors

Assuming cases (1) and (2) can be handled outside of daachorse, the remaining issue is case (3). However, considering the data compression performance of the compact double array and the intended use cases of tokenizers, the likelihood of such a pattern set being input is extremely low at present. Therefore, I believe there is little need to worry about error handling for this scenario.

ArthurZucker: ty for the hint!

PR #2003, opened 2026-03-31
feat(pattern): parallel regex `find_matches` for large inputs

McPatate:
- Parallel `find_matches` for `&Regex` and `&SysRegex` patterns using Rayon, based on the authority zone approach from Crusoe AI's fastokens
- Input is divided into chunks (8KB min) with 1KB forward overlap; each chunk is matched independently, then results are merged with boundary repair (ghost detection, truncation correction, gap filling)
- Activates automatically for inputs >16KB when `TOKENIZERS_PARALLELISM` is enabled; short inputs use the existing sequential path unchanged
- Correctness validated against sequential output with GPT-2 regex (including lookahead), multi-byte Unicode, and pathological inputs (long words spanning chunk boundaries)

Benchmark results from the provided bench file:

  | Input size | Sequential | Parallel | Speedup |
  |---|---|---|---|
  | 1K chars | 37 µs | 38 µs | ~1x (below // threshold) |
  | 10K chars | 397 µs | 400 µs | ~1x (below // threshold) |
  | 100K chars | 3.9 ms | 850 µs | **4.6x** |
  | 500K chars | 19.5 ms | 3.7 ms | **5.3x** |
  | Full corpus (~6MB) | 241 ms | 52 ms | **4.6x** |

### DIFF
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -1,5 +1,8 @@
 [package]
-authors = ["Anthony MOI <m.anthony.moi@gmail.com>", "Nicolas Patry <patry.nicolas@protonmail.com>"]
+authors = [
+  "Anthony MOI <m.anthony.moi@gmail.com>",
+  "Nicolas Patry <patry.nicolas@protonmail.com>",
+]
 edition = "2018"
 name = "tokenizers"
 version = "0.22.3-dev.0"
@@ -13,7 +16,14 @@ description = """
 Provides an implementation of today's most used tokenizers,
 with a focus on performances and versatility.
 """
-exclude = [ "rust-toolchain", "target/*", "Cargo.lock", "benches/*.txt", "benches/*.json", "data/*" ]
+exclude = [
+  "rust-toolchain",
+  "target/*",
+  "Cargo.lock",
+  "benches/*.txt",
+  "benches/*.json",
+  "data/*",
+]
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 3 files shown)


McPatate:
Note that I have validated correctness with the xnli dataset with the following test, I can upload it but doesn't feel necessary imo:
```rs
    #[test]
    fn parallel_correctness_xnli() {
        let Ok(data) = std::fs::read_to_string("data/xnli.txt") else {
            eprintln!("Skipping parallel_correctness_big_txt: data/xnli.txt not found");
            return;
        };

        let patterns = vec![
            SysRegex::new(r"\s+").unwrap(),
            SysRegex::new(r"\w+|[^\w\s]+").unwrap(),
            SysRegex::new(
                r"'s|'t|'re|'ve|'m|'ll|'d| ?\p{L}+| ?\p{N}+| ?[^\s\p{L}\p{N}]+|\s+(?!\S)|\s+",
            )
            .unwrap(),
        ];

        for pattern in &patterns {
            // Test with production constants (8KB chunks, 1KB overlap)
            let sequential = pattern.find_matches(&data).unwrap();
            set_parallelism(false);
            let seq_explicit = pattern.find_matches(&data).unwrap();
            set_parallelism(true);
            let parallel = pattern.find_matches(&data).unwrap();

            assert_eq!(
                sequential, seq_explicit,
                "Sequential results should be stable"
            );
            assert_eq!(
                sequential, parallel,
                "Parallel should match sequential on xnli.txt"
            );
        }
    }
```

ArthurZucker (commented):
As I commented offline we should really bench the entire encode function because it's already wrapped around rayon, well this is very promising I realize might not be the best direction

PR #2011, opened 2026-04-08
feat: compact vocabulary — single-allocation id→token store for BPE

ArthurZucker:
## Summary

- Replaces `vocab_r: AHashMap<u32, String>` in `BPE` with `CompactVocab`
- All token strings concatenated into one `Vec<u8>` buffer, indexed by a `Vec<u32>` offset array
- `id_to_token` is now two array reads instead of a hash-table probe
- Serializes/deserializes as the same `{"token": id}` JSON — no format change
- Sparse ids (gaps) handled via empty-slice sentinel

## Benchmark

*Results on compact-vocab branch — baseline (main) bench running, will update with delta.*

| Benchmark | Time | Throughput |
|---|---|---|
| GPT2 encode | 1.079 s | 5.74 MiB/s |
| GPT2 encode batch | 833 ms | 7.43 MiB/s |
| GPT2 encode, no cache | 1.277 s | 4.85 MiB/s |
| GPT2 encode batch, no cache | 234 ms | 26.4 MiB/s |
| Train small | 26.2 ms | 277 KiB/s |
| Train large | 791 ms | 7.82 MiB/s |

## Notes

- Draft — only BPE uses `CompactVocab`; WordLevel and WordPiece still use `AHashMap<u32, String>`
- Forward map (`vocab: AHashMap<String, u32>`) is unchanged; a follow-up could eliminate string duplication there too using a custom hasher over slices of the data buffer

### DIFF
--- a/tokenizers/src/models/bpe/model.rs
+++ b/tokenizers/src/models/bpe/model.rs
@@ -1,6 +1,7 @@
-use super::{super::OrderedVocabIter, trainer::BpeTrainer, Error, Pair, Word};
+use super::{trainer::BpeTrainer, Error, Pair, Word};
 use crate::tokenizer::{Model, Result, Token};
 use crate::utils::cache::{Cache, DEFAULT_CACHE_CAPACITY, MAX_LENGTH};
+use crate::utils::compact_vocab::CompactVocab;
 use crate::utils::iter::ResultShunt;
 use ahash::AHashMap;
 use serde_json::Value;
@@ -15,7 +16,6 @@ use std::{
 };
 
 pub type Vocab = AHashMap<String, u32>;
-type VocabR = AHashMap<u32, String>;
 pub type MergeMap = AHashMap<Pair, (u32, u32)>;
 pub type Merges = Vec<(String, String)>;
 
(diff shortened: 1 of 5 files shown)


ArthurZucker:
## Benchmark results — compact-vocab vs main

Ran `cargo bench --bench bpe_benchmark` independently on both branches (same machine, sequential runs).

| Benchmark | main | compact-vocab | Δ |
|---|---|---|---|
| GPT2 encode | 1.102 s | 1.079 s | **−2.1%** |
| GPT2 encode batch | 775 ms | 833 ms | +7.5% |
| GPT2 encode, no cache | 1.312 s | 1.277 s | **−2.7%** |
| GPT2 encode batch, no cache | 241 ms | 234 ms | **−2.9%** |
| Train small | 25.9 ms | 26.2 ms | +1.2% |
| Train large | 784 ms | 791 ms | +0.9% |

### Reading

Single-threaded encode and no-cache paths show a consistent 2–3% improvement — consistent with `vocab_r` lookups hitting the compact buffer instead of chasing hash-table pointers.

The cached batch result (+7.5%) is the outlier and is suspicious: the no-cache batch goes the other way (−2.9%), which is the path that calls `vocab_r` more heavily. Likely thermal noise between the two runs rather than a real regression; worth a second run with criterion's `--save-baseline` / `--load-baseline` flags on the same binary to be sure.

Train throughput is flat (±1%), as expected — `vocab_r` is rarely touched during training.

### Summary

The change is at worst neutral and shows a measurable win on the hot single-threaded decode path. The forward map (`vocab: AHashMap<String, u32>`) is the next candidate if we want bigger gains.

Issue #1900, opened 2025-11-30
Performance: batch_encode scales poorly on high-core Server CPUs compared to sharded tokenizer instances


stargazerZJ:
TLDR: In a 120-core k8s container (cpu request = cpu limit) on a 192-core system, sharding the tokenizer manually (letting 32 threads share one tokenizer instance) drastically outperforms the default `batch_encode` implementation on uniform-length, synthetic data.

I have also tested this on actual text datasets with non-uniform lengths, where my manual worker group implementation performs **7-9x faster**. This likely indicates lock contention, memory locality issues, or other bottlenecks in the current `batch_encode` implementation when scaling to high core counts.

**Benchmark Output:**

```text
❯ ./bench_data/target/release/tokenizer_bench -i bench_data -n 4 --compare-batch -b 1024 -w 120 --threads-per-tokenizer 32 --synthetic
[INFO] Tokenization Benchmark
[INFO] Workers: 120
[INFO] Batch size: 1024
[INFO] Loading tokenizer: Qwen/Qwen2.5-Coder-32B-Instruct
[INFO] Tokenizer loaded successfully
[INFO] Generating 10000 synthetic strings of 204800 bytes each
[00:00:08] ======================================== 10000/10000 strings
[INFO] Generated 10000 strings, total size: 1953.12 MB

=== Worker Pool Tokenization (120 workers, 32 threads per tokenizer) ===
[INFO] Creating 4 tokenizer groups
[00:00:12] ======================================== 10000/10000 Docs (798.3005/s)
[THROUGHPUT] Docs: 10000 (797.9/s) | Tokens: 1346513955 (107441384.6/s) | Data: 1.91 GB (155.84 MB/s) | Elapsed: 12.5s

=== Batch Tokenization (batch_size=1024) ===
[00:00:40] ======================================== 10000/10000 Docs (244.8588/s)
[THROUGHPUT] Docs: 10000 (244.9/s) | Tokens: 1346513955 (32970451.7/s) | Data: 1.91 GB (47.82 MB/s) | Elapsed: 40.8s

=== Performance Comparison ===
Worker Pool: 12.5s | Batch: 40.8s | Speedup: 0.31x

[INFO] Benchmark complete
❯ ./bench_data/target/release/tokenizer_bench -i bench_data -n 4 --compare-batch -b 10000 -w 120 --threads-per-tokenizer 32 --synthetic
[INFO] Tokenization Benchmark
[INFO] Workers: 120
[...]

PR #2029, opened 2026-04-23
Batch encode: lock-free work queue with dynamic window sizing

sebpop:
Replace `inputs.into_maybe_par_iter().map(...).collect()` in `encode_batch`, `encode_batch_char_offsets` and `encode_batch_fast` with a small helper `TokenizerImpl::run_batch` that:

- Dispatches to a plain `inputs.into_iter().map(...).collect()` serial loop when parallelism is disabled or only one thread is available, avoiding all rayon involvement for single-threaded callers.
- At higher thread counts, uses a lock-free atomic counter (`BatchWorkQueue`) inside one `rayon::scope` with one `s.spawn` per worker.  Each worker claims windows of item indices via a single `AtomicUsize::fetch_add`, takes inputs from per-slot `UnsafeCell<Option<EncodeInput>>`, and writes results into per-slot `UnsafeCell<Option<Result<Encoding>>>`.  No shared mutable state outside the counter; no final `collect()` on a parallel iterator.

The lock-free design is motivated by aarch64 LSE atomic cost: every mutex / condvar the previous parallel-iterator path took hit was a CAS / LDADD emitted by libpthread, and those dominate small-work parallel loops at high thread counts on arm64.  Replacing that with a single `fetch_add` per window removes the mutex-backed per-item signaling entirely.

## Cache-line / loop-tiling rationale

Shared-memory parallel loops are bottlenecked by the cache coherence protocol when two cores alternate writes to the same cache line: the line "ping-pongs" between their private L1d caches, each transfer costing dozens of cycles.  To avoid that, every line should be filled by one producer core, drained (or no longer needed), and only then touched by a different core.  This is the cache-aware equivalent of loop tiling / blocking: group the iteration space into chunks whose data footprint is a whole number of cache lines, and give each chunk to a single core.

The work queue enforces this three ways:

1. The counter itself lives on its own 64-byte cache line (`#[repr(C, align(64))]` on `AlignedCounter`).  A worker's `fetch_add` does not evict any neighbouring data, and reads of the counter do not pull input or result payloads into the core's L1d.

2. Each window is a contiguous run of `window_size` indices, so every worker owns a run of adjacent slots for the duration of one window.  With `MAX_WINDOW_SIZE = 8`, a window covers roughly `8 * sizeof(slot)` bytes -- for `Option<EncodeInput>` (~48 B) that is ~6 cache lines; for `Option<Result<Encoding>>` (multi-line per slot) it is even more.  Within one window, a worker writes several whole cache lines before any other worker comes near them.

3. Each slot has its own `UnsafeCell` (`Vec<UnsafeCell<Option<T>>>`).  `UnsafeCell<T>` is `#[repr(transparent)]`, so the heap layout is byte-identical to a plain `Vec<Option<T>>` (no padding, same alignment, same contiguous packing -- zero runtime overhead vs. the "unsafe fast" version that reborrows the whole `Vec`).  What the per-slot cell buys is that `self.0[i].get()` returns `*mut Option<T>` pointing straight at slot `i`, without ever materialising a `&mut Vec<Option<T>>` that would alias the enclosing container (which is UB when two threads touch any distinct indices concurrently).

At window boundaries a single cache line can be shared between two successive windows when the slot size does not divide 64 bytes.  That is a sequential handoff (window N finishes writes; window N+1 then reads/writes), not a concurrent ping-pong, so the cost is at most one coherence transfer per window-pair.

## Window sizing

`window_size = ceil(total / (num_threads * WINDOWS_PER_THREAD))`, clamped to `[1, MAX_WINDOW_SIZE]`.

- `WINDOWS_PER_THREAD = 4` keeps several windows per thread so a slow worker on its last item does not stall the whole batch.
- `MAX_WINDOW_SIZE = 8` caps per-claim atomic latency and keeps the per-window memory footprint small enough to fit in L1d.

Examples: 100 items / 16 threads yields `window_size = 2` (50 windows); 10 000 items / 16 threads yields `window_size = 8` (1250 windows).

## Tests

7 new unit tests in `utils::batch::tests` cover window sizing, `TakeVec` and `ResultVec` round-trip, and `test_parallel_distribution` (4 threads concurrently claiming and writing 100 slots, exercising the Sync bounds under real contention).

cargo test --lib --features http: 208 passed, 0 failed.

## Perf evidence

On Vera (88-core Olympus, 176 logical),
`bpe_benchmark`/`bpe-encode/BPE GPT2 encode batch` at 88T, `perf record -g --call-graph fp -F 4999`.

LSE atomic instructions (the direct motivation for the lock-free counter):
```
  instruction                    before    after
  __aarch64_cas4_acq              3.57%   0.61%   (-5.9x)
  __aarch64_ldadd8_acq_rel        1.05%   0.08%   (-13x)
  __aarch64_swp4_rel              0.21%   0.05%
  __aarch64_ldadd8_relax          0.17%   0.24%
  __aarch64_swp4_acq              0.12%   0.00%
  __aarch64_swp8_acq_rel          0.06%   0.00%
  __aarch64_cas8_acq_rel          0.01%   0.01%
  total LSE                       ~5.2%   ~1.0%   (-4.2x)
```
Rayon / crossbeam-epoch:
```
  symbol                                                     before    after
  rayon_core::sleep::Sleep::wake_specific_thread              0.57%   0.06%   (-10x)
  crossbeam_epoch::internal::Global::try_advance             25.93%  28.38%
  crossbeam_epoch::default::with_handle                      21.41%  23.12%
  rayon_core::registry::WorkerThread::wait_until_cold         8.40%  10.72%
  rayon::iter::plumbing::bridge_producer_consumer::helper     0.20%   0.24%
```
`bridge_producer_consumer::helper` was not a hotspot on this workload before the change (0.20%) and does not move; the observable rayon-side change is `Sleep::wake_specific_thread` dropping ~10x because `rayon::scope` issues one wake per worker per batch call rather than streaming wakes per parallel-iterator split.  The three remaining rayon/crossbeam ceiling symbols (`try_advance` + `with_handle` + `wait_until_cold` = ~62% of cycles) stay similar in percentage because total cycles decrease; absolute wall-clock per benchmark iteration drops 35 ms (295 ms -> 260 ms at 88T).  Removing that rayon ceiling is a separate change.

Throughput on Vera, `bpe-encode/BPE GPT2 encode batch` (data/big.txt, encode_batch through the full post-processor):
```
  threads  before       after        change
  -------  ------       ------       ------
  1T       3.98 MiB/s   4.46 MiB/s   +12%
  88T      20.97 MiB/s  23.76 MiB/s  +13%
  176T     18.83 MiB/s  21.58 MiB/s  +15%
```

### DIFF
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -20,6 +20,7 @@ use std::{
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 
+use crate::utils::batch::{BatchWorkQueue, ResultVec, TakeVec};
 use crate::utils::iter::ResultShunt;
 use crate::utils::parallelism::*;
 use crate::utils::progress::{ProgressBar, ProgressStyle};
@@ -1300,7 +1301,11 @@ where
     PP: PostProcessor + Send + Sync,
     D: Decoder + Send + Sync,
 {
-    /// Encode all the sentences in parallel, using multiple threads
+    /// Encode all the sentences in parallel, using multiple threads.
+    ///
+    /// Uses a lock-free work queue with cache-line-sized windows instead of
+    /// rayon's `bridge_producer_consumer`, eliminating its synchronization
+    /// overhead at higher thread counts.
     pub fn encode_batch<'s, E>(
         &self,
         inputs: Vec<E>,
@@ -1309,13 +1314,10 @@ where
     where
         E: Into<EncodeInput<'s>> + Send,
     {
-        let mut encodings = inputs
-            .into_maybe_par_iter()
-            .map(|input| self.encode(input, add_special_tokens))
-            .collect::<Result<Vec<Encoding>>>()?;
+        let mut encodings =
+            self.run_batch(inputs, |this, input| this.encode(input, add_special_tokens))?;
 
         if let Some(params) = &self.padding {
-            // We do the padding here to make sure we handle the batch padding
             pad_encodings(&mut encodings, params)?;
         }
 
@@ -1332,20 +1334,22 @@ where
     where
         E: Into<EncodeInput<'s>> + Send,
     {
-        let mut encodings = inputs
-            .into_maybe_par_iter()
-            .map(|input| self.encode_char_offsets(input, add_special_tokens))
-            .collect::<Result<Vec<Encoding>>>()?;
+        let mut encodings = self.run_batch(inputs, |this, input| {
+            this.encode_char_offsets(input, add_special_tokens)
+        })?;
 
         if let Some(params) = &self.padding {
-            // We do the padding here to make sure we handle the batch padding
             pad_encodings(&mut encodings, params)?;
         }
 
         Ok(encodings)
     }
 
-    /// Encode all the sentences in parallel, using multiple threads
+    /// Encode all the sentences in parallel, using multiple threads.
+    ///
+    /// Uses a lock-free work queue with cache-line-sized windows instead of
+    /// rayon's `bridge_producer_consumer`, eliminating its synchronization
+    /// overhead at higher thread counts.
     pub fn encode_batch_fast<'s, E>(
         &self,
         inputs: Vec<E>,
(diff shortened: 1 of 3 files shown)


sebpop:
Small semantic note, since @codex flagged it: 
rayon's `collect::<Result<_, _>>()` is best-effort fail-fast: other workers are asked to stop after an Err is observed, but in-flight items may still finish. The new implementation always processes every item, then returns the Err at the lowest failing index deterministically. For all-Ok batches the two are identical; for failing batches, the new implementation does bounded extra work, at most `batch_size - first_error_index - 1` additional encode calls after the first failing item, in exchange for deterministic error selection.
That determinism comes from collecting the per-slot results back in input order after all workers finish: i.e., `into_vec().into_iter().collect()`.

I chose not to add a shared stop flag because it would add hot-path polling to every batch to improve a rare cold path, and because deterministic-error-at-lowest-index is arguably the better user contract. This tradeoff follows Brendan Gregg's [Utilization Saturation and Errors (USE) Method](https://www.brendangregg.com/usemethod.html): optimize the common Utilization/Saturation path while keeping the rare Error path correct and bounded.
Happy to add it if a reviewer thinks the behavior change warrants a follow-up.

vyalamar:
@sebpop  Great timing. I was also looking at #1900 myself. 
Since you bench-marked on Vera CPU, I can help out by running a validation on some  cloud hardware to see the scaling improvements generalize. I am curios to how it handles x86 hardware (Intel Sapphire) 
Plan is to spin up hire Graviton , AMD Chiplet, and AWS x86 ( c7i.metal-24xl) and test v0.22.2 vs. main (post-#2028) vs. this PR.

If there is a preferred benchmark command etc for this test, I am happy to use that @sebpop  Else I’ll drop a compact summary table and the raw data here once I have the machines spun up and the runs finished.

sebpop:
Thanks @vyalamar, very welcome. See the recipe below — tested on Vera and Grace (Nvidia arm64); should reproduce on x86_64 (Intel and AMD) and on arm64 Graviton with no fiddling.

Build the benchmark:
```
git clone https://github.com/huggingface/tokenizers.git
cd tokenizers/tokenizers
cargo bench --bench bpe_benchmark --no-run
```

Run:
```
BPE=$(ls -t target/release/deps/bpe_benchmark-* | grep -v '\.d$' | head -1)
THREADS="1 64 96 128 192"
for T in $THREADS; do
    echo "=== threads=$T ==="
    RAYON_NUM_THREADS=$T $BPE --bench '^bpe-encode/BPE GPT2 encode batch$'
done
```
You can also run the benchmark under https://github.com/aws/aperf and check flamegraphs and other PMU metrics.

I expect AMD and Intel CPUs to behave similarly to arm64. Let me know if you hit any roadblocks.

vyalamar:
@sebpop  keep you posted today with the results. Sorry  there was some Git outage https://www.githubstatus.com/

sebpop:
https://github.com/huggingface/tokenizers/pull/2044 (branch p2-prime) has a simpler implementation of the lock-free scheduler. #2044 has about the same performance as the implementation in the current PR #2029 (branch p2):
- throughput measured at main 20.67 MiB/s, p2 24.53 MiB/s, p2-prime 23.55 MiB/s
- reduction in LSE atomic operations: main 5.5%, p2 0.59%, p2-prime 0.9%.

Diff for p2-prime is smaller than p2:
- p2: 3 files, 381 insertions, 17 deletions, net +364.
- p2-prime: 2 files, 208 insertions, 18 deletions, net +190.

vyalamar:
@sebpop  : https://github.com/huggingface/tokenizers/pull/2044#issuecomment-4339407947

sebpop:
Closing this in favor of p2-prime branch in #2044, which lands the same idea in a smaller and simpler form. p2-prime drops the standalone tokenizers::utils::batch module (no new public surface) and folds the work-queue, slot, and cache-padded counter directly into a private helper inside tokenizer/mod.rs. The fork-safety USED_PARALLELISM marker is preserved via a small get_parallelism_with_use() helper in utils/parallelism.rs, and three correctness tests were added that this PR didn't have: encode_batch_variants_keep_order_in_parallel (covers all three batch entry points under a 4-thread pool), encode_batch_propagates_parallel_errors, and test_get_parallelism_marks_parallelism_used. Side-by-side measurement on Vera at 88 threads, bpe-encode/BPE GPT2 encode batch:

 Desc | this PR #2029 (p2) | PR #2044 (p2-prime)
-- | -- | --
throughput @88T | 24.53 MiB/s | 24.25 MiB/s (+4%)
total LSE atomic share | ~0.57% | ~0.97% (both well below the 5.2% baseline)

Both reduce LSE traffic by ~5× vs origin/main; the residual ~1% is in rayon_core::sleep::Sleep::wake_specific_thread's mutex, not in either implementation of the work queue. That, together with crossbeam_epoch::try_advance + with_handle + WorkerThread::wait_until_cold accounting for ~67% of cycles at 88T, is the next bottleneck — addressed by a follow-up that replaces rayon::scope with a barrier-based persistent worker pool, which I'll send as a separate PR on top of #2044. Continuing the work in #2044 (p2-prime branch.)

PR #2036, opened 2026-04-26
feat: SIMD ASCII fast path for Lowercase normalizer (~30-49x)

KimYannn:
## Summary

- Adds `utils::simd::ascii_lower`, a runtime-dispatched in-place ASCII lowercaser with AVX-512BW, AVX2, SSE2, NEON, and scalar back-ends.
- Gates `NormalizedString::lowercase` on `s.is_ascii()` so all-ASCII inputs skip the per-`char` Unicode case-folding loop and the `transform()` alignments rebuild, mutating bytes in place instead.
- Microbench shows ~30-49× over the previous Unicode `chars()` path on Apple/NEON, and **up to ~117×** on x86_64 AVX2 (Zen 4) at 16 KiB.

## Why this is safe

For any ASCII char, `char::to_lowercase()` returns exactly one ASCII char equal to `b | 0x20` for `A`..=`Z` (Unicode `SpecialCasing.txt` has no ASCII entries), so 1 byte → 1 byte and `transform()` would have rebuilt `alignments` to the *same values* it already had. Skipping `transform()` is therefore observationally equivalent for ASCII inputs.

Two new unit tests pin this down:

- `lowercase_ascii_fast_path_preserves_alignments` — after a non-trivial `nfkd()` transform, asserts `normalized`, `alignments`, `original`, and `original_shift` are byte-identical between fast-path output and the expected values.
- `lowercase_ascii_matches_unicode_path_byte_for_byte` — for every printable ASCII byte, checks the fast path output equals `chars().flat_map(to_lowercase)`.

Non-ASCII inputs hit the original code path unchanged.

## Implementation notes

- Runtime dispatch via `is_x86_feature_detected!("avx512fp16")`, `"avx2"`. SSE2 is always available on the x86_64 baseline; aarch64 stable ABI mandates NEON, so no detect there.
- AVX-512BW path is gated on `avx512fp16` (not `avx512bw`) intentionally as a frequency-safety proxy: `avx512fp16` lights up only on Sapphire Rapids+ / Granite Rapids / Zen 5, where heavy 512-bit integer ops do not meaningfully downclock the core. Pre-SPR Intel parts (SKX/CLX/ICX) and Zen 4 — which all *have* AVX-512BW — fall back to AVX2 to avoid the historical 512-bit clock penalty.
- Each `unsafe fn ascii_lower_*` is `#[target_feature(enable = "...")]`-gated and has a scalar tail. The SSE2/AVX2 paths use signed `cmpgt_epi8`, which excludes bytes ≥ 0x80 naturally — defensive even though the gate filters them out.
- Equivalence-tested across critical lengths (0, 1, 7, 15, 16, 17, 31, 32, 33, ...) and against high-bit bytes mixed in.

## Microbench

`cargo bench --bench ascii_lower_benchmark`.

The "unicode chars" stand-in mirrors what `NormalizedString::lowercase` did before this change (per-`char` UTF-8 decode + `to_lowercase()` + `Vec<(char, isize)>` accumulation, alignment bookkeeping omitted).

### Apple Silicon (NEON)

|   size  |   SIMD     |  scalar (auto-vec) | unicode chars (old) | SIMD vs old |
|---------|------------|--------------------|---------------------|-------------|
|   64 B  | 30.2 GiB/s |     5.9 GiB/s      |     1.0 GiB/s       |   ~30×      |
|  1 KiB  | 55.4 GiB/s |     6.3 GiB/s      |     1.2 GiB/s       |   ~48×      |
| 16 KiB  | 58.4 GiB/s |     6.3 GiB/s      |     1.2 GiB/s       |   ~49×      |
| 256 KiB | 57.9 GiB/s |     6.3 GiB/s      |     1.2 GiB/s       |   ~48×      |

### x86_64 AVX2 — AMD Ryzen 7 7800X3D (Zen 4, Ubuntu 26.04, rustc 1.95 stable, criterion `sample_size=50`)

|   size  |   SIMD      |  scalar (auto-vec) | unicode chars (old) | SIMD vs scalar | SIMD vs old |
|---------|-------------|--------------------|---------------------|----------------|-------------|
|   64 B  | 17.5 GiB/s  |     3.51 GiB/s     |     591 MiB/s       |   ~5.0×        |   ~30×      |
|  1 KiB  | 70.5 GiB/s  |     3.61 GiB/s     |     723 MiB/s       |   ~19.5×       |  ~100×      |
| 16 KiB  | 83.0 GiB/s  |     3.63 GiB/s     |     728 MiB/s       |   ~22.9×       |  ~117×      |
| 256 KiB | 65.6 GiB/s  |     3.63 GiB/s     |     738 MiB/s       |   ~18.1×       |   ~91×      |

Notes for the AVX2 row:

- AVX-512BW path is *not* taken on Zen 4 because `avx512fp16` is absent — by design (see Implementation notes). On a Zen 5 / Sapphire Rapids+ host this path will engage and should improve the 1–16 KiB range further.
- `scalar` (the simple `is_ascii_uppercase() | 0x20` byte loop) tops out at ~3.6 GiB/s on this host because the autovectorizer does not lift it; SIMD wins ~5–23× even against this baseline.
- 16 KiB peak (~83 GiB/s) is bounded by L1/L2; 256 KiB tail reflects working-set leaving L1.

## Test plan

- [x] `cargo test --lib` (207 tests pass)
- [x] `cargo build --release --lib`
- [x] New unit tests in `utils::simd::tests` (5 tests covering empty input, critical lengths, mixed-case random, idempotence, and high-byte safety)
- [x] New equivalence tests in `tokenizer::normalizer::tests`
- [x] `cargo bench --bench ascii_lower_benchmark` (Apple/NEON + x86_64 AVX2 / Zen 4)
- [x] `cargo fmt --check`
- [ ] CI: clippy + tests on x86_64 (running via repo CI)

## Follow-ups (not in this PR)

- Same pattern can extend to `Strip`, `BertPreTokenizer` byte classification, and `ByteLevel` ASCII pass-through.


KimYannn:
Hi @ArthurZucker, could you please take a look when you have a moment? This adds a SIMD ASCII fast path for the Lowercase normalizer with benchmarks included. Happy to adjust the approach if needed. Thanks!

tokenizers/src/tokenizer/normalizer.rs
```diff
@@ -544,6 +544,16 @@ impl NormalizedString {
 
     /// Lowercase
     pub fn lowercase(&mut self) -> &mut Self {
+        // ASCII fast path: each `A`..=`Z` becomes a single-byte `a`..=`z`,
+        // so byte length and per-byte alignments are unchanged. We can mutate
+        // bytes in place and skip the Unicode-aware `transform` rebuild.
+        if self.normalized.is_ascii() {
```
ArthurZucker, line 550: https://doc.rust-lang.org/src/core/str/mod.rs.html#2782 if this is using this, the check itself is not + does sequentially read the stream no? (https://doc.rust-lang.org/src/core/slice/ascii.rs.html#45) 
Since the feature is gated, you could have this run at least on chunks that are ascii entirely!

ArthurZucker (approved):
Hey! sorry for the late review, this looks great actually! not only there is a scallar fallback, but its ported cleanly and simply. 0
We are in the middle of pushing v1 for tokenizers which is based on #2119 if you we can rebase there directly, or just merge it as is?

ArthurZucker:
See #2150 were we are starting to leverage similar ideas but for pretokenizers

PR #2037, opened 2026-04-26
feat(NFC): skip Unicode pass for all-ASCII inputs

KimYannn:
## Summary

Adds a single-line ASCII gate to the `NFC` normalizer: when `NormalizedString::get().is_ascii()` is true the normalizer returns immediately, skipping the per-`char` `unicode-normalization-alignments` pass and the `transform()` rebuild it triggers.

## Why this is safe

ASCII (U+0000..=U+007F) is NFC by Unicode invariant:

- no ASCII code point has a `Decomposition_Mapping`
- no ASCII code point combines with adjacent characters
- no ASCII code point is the target of any composition

Running the existing NFC code on an all-ASCII `NormalizedString` therefore rebuilds `normalized` and `alignments` to the exact same bytes and tuples it already had — the gate just elides that work.

The gate is strictly conservative: any non-ASCII byte in the input falls through to the original code path with zero changes. Combining-mark sequences (e.g. `e` + U+0301 → `é`), CJK, Arabic, Cyrillic, Vietnamese, Hindi, Thai, etc. all keep their existing behavior.

## Tests

Two unit tests in `normalizers::unicode::tests` pin this:

- `nfc_ascii_fast_path_is_no_op` — runs `NFKD` on `ﬀ` first to produce all-ASCII text with non-trivial alignments (each output byte still maps back to the 3-byte ligature), then asserts `NFC.normalize` leaves the entire `NormalizedString` byte-identical.
- `nfc_non_ascii_still_runs_unicode_path` — checks that `e` + combining acute is still composed to `é`.

## Test plan

- [x] `cargo test --lib` (202 tests pass)
- [x] New equivalence + non-ASCII regression tests
- [ ] CI on x86_64 / Linux / macOS

## Notes

This is a minimal-risk performance gate — no new dependencies, no `unsafe`, no SIMD. Companion to #2036 (SIMD ASCII fast path for `Lowercase`), but fully independent of it.

Same-shape gates may apply to `NFD`, `NFKC`, `NFKD` (ASCII is also already NFD/NFKD/NFKC by the same invariant); happy to extend in a follow-up if maintainers prefer. Kept minimal here for review clarity.


KimYannn:
Hi @ArthurZucker, could you please take a look when you have a moment? This skips the Unicode NFC path for all-ASCII inputs and should be a small safe perf win. Happy to update anything if needed. Thanks!

PR #2041, opened 2026-04-27
bindings/python: free-threaded Python (3.14t) support

ArthurZucker:
Adds dedicated 3.14t support to the python bindings without breaking the regular CPython API surface.
The release workflow was updated but we did not specify "use_gil" -> it would be a pointless release. 

Key changes:

- Wrap PyTokenizer's inner Tokenizer in std::sync::RwLock<Tokenizer>. 

- Each #[pymodule] is now declared as `#[pymodule(gil_used = false))]` 

- Promote `abi3` from a hardcoded pyo3 dep-feature to a project-level cargo feature (default on). Allows building without abi3 on free-threaded Python (`maturin develop --no-default-features --features ext-module`) — abi3 / limited API is not available under free-threading.

- Add bindings/python/tests/test_freethreaded.py: stress tests racing N encoders against M setters on the same Tokenizer. All pass on 3.14t (4/4) and regular 3.14 (3 pass + 1 skip for the 3.14t-specific GIL check).

Building 3.14t wheels:

  maturin develop --release --no-default-features --features ext-module


bindings/python/src/decoders.rs
```diff
@@ -846,7 +846,7 @@ impl PyDecodeStream {
             StreamInput::Ids(ids) => ids,
         };
         ToPyResult(tk::tokenizer::step_decode_stream(
-            &tokenizer.tokenizer,
+            &tokenizer.tokenizer.read().unwrap(),
```
McPatate: could be worth erroring out cleanly instead of hard panic here

bindings/python/src/tokenizer.rs
```diff
@@ -513,15 +514,31 @@ type Tokenizer = TokenizerImpl<PyModel, PyNormalizer, PyPreTokenizer, PyPostProc
...
+            tokenizer: RwLock::new(self.tokenizer.read().unwrap().clone()),
+        }
+    }
+}
+
+impl Serialize for PyTokenizer {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        // Equivalent to the previous `#[serde(transparent)]` derive — forward
+        // through to the inner Tokenizer.
+        self.tokenizer.read().unwrap().serialize(serializer)
```
McPatate: unwrap again?

bindings/python/src/tokenizer.rs
```diff
@@ -513,15 +514,31 @@ type Tokenizer = TokenizerImpl<PyModel, PyNormalizer, PyPreTokenizer, PyPostProc
...
-#[serde(transparent)]
 pub struct PyTokenizer {
-    pub(crate) tokenizer: Tokenizer,
+    pub(crate) tokenizer: RwLock<Tokenizer>,
+}
+
+impl Clone for PyTokenizer {
+    fn clone(&self) -> Self {
+        PyTokenizer {
+            tokenizer: RwLock::new(self.tokenizer.read().unwrap().clone()),
```
McPatate: wrap the lock in `Arc` for cheap cloning?

bindings/python/src/tokenizer.rs
```diff
@@ -513,15 +514,31 @@ type Tokenizer = TokenizerImpl<PyModel, PyNormalizer, PyPreTokenizer, PyPostProc
     name = "Tokenizer",
     from_py_object
 )]
-#[derive(Clone, Serialize)]
-#[serde(transparent)]
 pub struct PyTokenizer {
-    pub(crate) tokenizer: Tokenizer,
+    pub(crate) tokenizer: RwLock<Tokenizer>,
```
McPatate:
```suggestion
    pub(crate) tokenizer: Arc<RwLock<Tokenizer>>,
```

McPatate (commented):
the only thing I'm mad about is that you `unwrap` everywhere, other than that lgtm! 🔥

PR #2044, opened 2026-04-28
Batch encode: simple lock-free scheduler

sebpop:
Replace the batch encode parallel-iterator collect path with a compact scoped scheduler that uses one cache-line-padded atomic counter to hand out contiguous windows of work. This keeps the throughput win with less code than the standalone batch utility while preserving the fork-safety parallelism marker for direct Rayon use.

The speedup comes from avoiding Rayon per-item iterator plumbing and mutex-backed worker wakeups in this hot loop. On aarch64 those wakeups show up as LSE atomic helper calls such as __aarch64_cas4_acq and __aarch64_ldadd8_acq_rel.

The scheduler claims contiguous windows so one producer thread fills a run of adjacent input and result slots before another thread touches the next run. Completed windows flow through shared L3 cache as whole runs, avoiding producer and consumer ping-pong where two cores repeatedly alternate ownership of the same L1 cache line.

Tests cover the window sizing policy, ordered output for all three batch encode entry points, parallel error propagation, and the parallelism-used marker.

Verification:
- cargo fmt --manifest-path tokenizers/Cargo.toml.
- cargo test --manifest-path tokenizers/Cargo.toml --lib --features http: 205 passed.
- on Vera, 88 Rayon threads, bpe-encode/BPE GPT2 encode batch: LSE samples dropped from ~5.5% on main to ~0.9% on p2-prime; throughput measured at main 20.67 MiB/s, p2 24.53 MiB/s, p2-prime 23.55 MiB/s.

### DIFF
--- a/tokenizers/src/tokenizer/mod.rs
+++ b/tokenizers/src/tokenizer/mod.rs
@@ -11,10 +11,12 @@
 
 use ahash::AHashMap;
 use std::{
+    cell::UnsafeCell,
     fs::{read_to_string, File},
     io::{prelude::*, BufReader},
     ops::{Deref, DerefMut},
     path::{Path, PathBuf},
+    sync::atomic::{AtomicUsize, Ordering},
 };
 
 use serde::de::DeserializeOwned;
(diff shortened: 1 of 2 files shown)


vyalamar:
@sebpop will benchmark this one as well along with #2029 . You already gave me the commands.

vyalamar:
Preliminary Graviton datapoint.

  On AWS `c8g.metal-48xl` / Graviton4 (`aarch64`, 192 cores, no SMT), I see the PRs helping in the mid/high thread range,  but not yet at  full-machine win. I will dig into the 192T behavior next.

  I used the requested benchmark target:

  ```
  RAYON_NUM_THREADS=$T $BPE --bench '^bpe-encode/BPE GPT2 encode batch$'
  ```
  
  Final pass cleared target/criterion between thread-count runs and parsed Criterion estimates.json.

  | threads | main MiB/s | p2 / #2029 MiB/s | p2-prime / #2044 MiB/s | p2/main | p2-prime/main |
  |---:|---:|---:|---:|---:|---:|
  | 1 | 3.05 | 3.12 | 3.15 | 1.02x | 1.03x |
  | 64 | 25.04 | 29.53 | 21.76 | 1.18x | 0.87x |
  | 96 | 25.48 | 29.56 | 28.41 | 1.16x | 1.11x |
  | 128 | 16.41 | 19.96 | 20.07 | 1.22x | 1.22x |
  | 192 | 14.08 | 10.03 | 10.98 | 0.71x | 0.78x |


   Scheduler change helps around 96-128T, but at 192T both PR branches were below main on this host. Earlier repeats also showed some high-thread variability on the PR branches, so I would treat this as preliminary, not a conclusion.

  I’ll follow up with profiling on the suspicious high-thread points. I want to separate Rayon scheduling/parking overhead,  full-core frequency/power effects, and anything specific to the new scheduler path.

vyalamar:
@sebpop  I will re-run Graviton once again and check with aperf.  Currently trying the AMD and instances.

McPatate:
Have you tried: `rayon::iter::IndexedParallelIterator::with_min_len(window_size)`? How does it compare with your custom impl?

sebpop:
> Have you tried: `rayon::iter::IndexedParallelIterator::with_min_len(window_size)`? How does it compare with your custom impl?

Good suggestion — measured it on Vera (88-core, same bpe_benchmark, window_size held equal):
```
               1T    88T    176T         LSE atomics
custom queue 4.62  19.86   20.03 MiB/s    ~1.0%
with_min_len 4.58  19.36   18.70 MiB/s   ~0.91%
```
with_min_len gets the LSE-atomic reduction that motivates this PR in a single line change. 

Closing this PR.
The simpler with_min_len implementation is in p2-second branch: https://github.com/huggingface/tokenizers/pull/2077

PR #2049, opened 2026-05-02
Add encode_byte_offsets methods for byte-level alignment

### LINKED ISSUE #1881: Provide byte-level offsets for effective alignment in Cross-Tokenizer On-Policy Distillation
APOLOGIZE: Sorry for submitting a PR several hours ago without creating an issue firstly. I'm urgently coming back here to initiate an issue.

Motivation: While performing cross-tokenizer on-policy distillation, we need to align two different encoding results (by two tokenizers) of the text, which is rolled out by the student model. To efficiently implement this, we hope the encode() function can provide **byte-level offset** information, so that the alignment method only needs to call the encode() method once.

This scenario requires byte-level offsets rather than character-level offsets to achieve the most compatible cross-tokenizer alignment.

According to my understanding of current tokenizers project, the Rust code already supports this functionality but has not exposed it to the Python encode() method.

https://github.com/huggingface/tokenizers/blob/007fc767ac78fe4ec43c6790afbee5063157fc62/bindings/python/src/tokenizer.rs#L1070-L1076

More description at:
https://github.com/huggingface/trl/issues/4393

PR:
https://github.com/huggingface/tokenizers/pull/1880

kashif:
Add `encode_byte_offsets()` and `encode_batch_byte_offsets()` methods to enable byte-level offset computation during encoding. This is useful for cross-tokenizer alignment scenarios where byte positions provide more accurate alignment than character positions, especially for multilingual text.

Based on suggestion from @ArthurZucker in older PR #1880 to add a dedicated method instead of modifying the existing encode API.

Fixes #1881

### DIFF
--- a/bindings/python/CHANGELOG.md
+++ b/bindings/python/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Added `encode_byte_offsets()` method for single sequence encoding with byte-level offsets
+- Added `encode_batch_byte_offsets()` method for batch encoding with byte-level offsets
+- These methods enable byte-level offset computation useful for cross-tokenizer alignment
+
 ## [0.13.2] 
 
 - [#1096] Python 3.11 support
@@ -510,3 +518,4 @@ delimiter (Works like `.split(delimiter)`)
 [#134]: https://github.com/huggingface/tokenizers/issues/134
 [#131]: https://github.com/huggingface/tokenizers/issues/131
 [#99]: https://github.com/huggingface/tokenizers/pull/99
+[#XXXX]: https://github.com/huggingface/tokenizers/pull/XXXX
--- a/bindings/python/CHANGELOG.md
+++ b/bindings/python/CHANGELOG.md
@@ -6,8 +6,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
-
 - [#2049] Added `encode_byte_offsets()` and `encode_batch_byte_offsets()` methods returning byte-level offsets, enabling cross-tokenizer alignment without repeated `decode()` calls
 
 ## [0.13.2] 
--- a/bindings/python/CHANGELOG.md
+++ b/bindings/python/CHANGELOG.md
@@ -8,9 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added `encode_byte_offsets()` method for single sequence encoding with byte-level offsets
-- Added `encode_batch_byte_offsets()` method for batch encoding with byte-level offsets
-- These methods enable byte-level offset computation useful for cross-tokenizer alignment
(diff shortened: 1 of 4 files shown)


bindings/python/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
```
kashif:
```suggestion
```

bindings/python/src/tokenizer.rs
```diff
@@ -1213,6 +1213,74 @@ impl PyTokenizer {
...
 
+    /// Encode the given sequence and pair, using byte-level offsets.
+    ///
+    /// This method can process raw text sequences as well as already pre-tokenized sequences.
+    ///
+    /// Example:
+    ///     Here are some examples of the inputs that are accepted::
+    ///
+    ///         encode_byte_offsets("A single sequence")
+    ///         encode_byte_offsets("A sequence", "And its pair")
```
ArthurZucker: let's improve with proper example + in and out

bindings/python/tests/bindings/test_tokenizer.py
```diff
@@ -193,6 +193,64 @@ def test_encode(self):
         output = tokenizer.encode_batch(["my name is john", ("my name is john", "pair")])
         assert len(output) == 2
 
+    def test_encode_byte_offsets(self):
```
ArthurZucker, line 208: all of these test don't test a single proper tokenizer that would have bytlevel offset no?

ArthurZucker: I mean added tokens only != BPE

bindings/python/src/tokenizer.rs
```diff
@@ -1213,6 +1213,74 @@ impl PyTokenizer {
...
+                    pair.extract::<PreTokenizedInputSequence>()?.into()
+                } else {
+                    pair.extract::<TextInputSequence>()?.into()
+                };
+                tk::EncodeInput::Dual(sequence, pair)
+            }
+            None => tk::EncodeInput::Single(sequence),
+        };
+
+        ToPyResult(
```
ArthurZucker, line 1303: no super sure I see what's been changed?

ArthurZucker (commented):
I don't understand what feature is added, I don't see a diff

bindings/python/src/tokenizer.rs
kashif: Docstring now shows concrete `>>> tokenizer.encode("café").offsets → [(0, 4)]` vs `[(0, 5)]`, plus the byte-level-split case

bindings/python/tests/bindings/test_tokenizer.py
kashif: Test now trains a real ByteLevel BPE with the 256-byte initial alphabet; café splits into 5 byte-level tokens and the assertions verify char (3, 4) vs byte (3, 5) on the trailing é-tokens

bindings/python/src/tokenizer.rs
kashif: Refactored to share `extract_borrowed_encode_input{,s}` helpers so the diff between each pair is now a single tokenizer-method call, flagged with an inline comment

kashif:
closing as i can do the byte offsets via the existing API in the trainers

PR #2056, opened 2026-05-13
serialize tokenizer vocab and added_tokens compactly

ArthurZucker:
## Summary

- `OrderedVocabIter` (used by BPE / WordPiece / WordLevel) now serializes the whole vocab map as a single line.
- `AddedVocabulary` now emits each entry of `added_tokens` as a single line, one token per array element — the outer array still wraps so diffs stay clean.

`serde_json`'s `raw_value` feature is enabled to support `RawValue::from_string` / `RawValue::serialize`.

Result: pretty output now looks like
```json
{
  "version": "1.0",
  ...
  "added_tokens": [
    {"id":0,"content":"[SPECIAL_0]","single_word":false,"lstrip":false,"rstrip":false,"normalized":false,"special":true},
    ...
  ],
  ...
  "model": {
    "type": "BPE",
    "vocab": {"<unk>":0,"a":1,"b":2,...},
    "merges": [...]
  }
}
```

### DIFF
--- a/bindings/node/lib/bindings/post-processors.test.ts
+++ b/bindings/node/lib/bindings/post-processors.test.ts
@@ -9,7 +9,9 @@ describe('bertProcessing', () => {
   })
 
   it('throws if only one argument is provided', () => {
-    expect(() => (bertProcessing as any)(['sep', 1])).toThrow('Given napi value is not an array')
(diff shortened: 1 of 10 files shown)


ArthurZucker:
v1 will have that

PR #2068, opened 2026-05-22
Fix typo in EncodingVisualizer.annotation_converter attribute

c-tonneslan:
Tiny fix found while reading the visualizer code.

\`EncodingVisualizer.__init__\` stashes the converter under \`self.annotation_coverter\` (missing 'n'), and \`__call__\` reads from the same misspelled attribute. The keyword argument and the docstring both spell it correctly, so the public name and the attribute name don't match. Anyone trying to set or read \`viz.annotation_converter\` after instantiation would silently end up with a separate attribute, while the misspelled one keeps being used internally.

Renamed the attribute to match the public spelling. No behavior change otherwise, just makes \`self.annotation_converter\` actually do what its name says.


ArthurZucker:
I think the file was removed no?

PR #2074, opened 2026-05-27
Avoid full vocab clone in get_vocab_size()

eunseo9311:
`get_vocab_size(true)` was calling `get_vocab(true).len()`, which clones the entire model vocabulary into a new `HashMap` just to count entries.
For large vocabularies (e.g. LLaMA-3 128k tokens) this allocates ~10MB on every call.

Fix by computing `base + added.len() - overlapping` directly, where `overlapping` counts added tokens already present in the model via `token_to_id`. Zero allocation.

Applies the same fix to the Node binding, which had the identical pattern.

Adds a test covering the overlap scenario (token in both model vocab and added_vocabulary).

### DIFF
--- a/bindings/node/src/tokenizer.rs
+++ b/bindings/node/src/tokenizer.rs
@@ -458,7 +458,12 @@ impl Tokenizer {
 
   #[napi]
   pub fn get_vocab_size(&self, with_added_tokens: Option<bool>) -> u32 {
-    self.get_vocab(with_added_tokens).len() as u32
+    let with_added_tokens = with_added_tokens.unwrap_or(true);
+    self
+      .tokenizer
+      .read()
+      .unwrap()
+      .get_vocab_size(with_added_tokens) as u32
   }
 
   #[napi]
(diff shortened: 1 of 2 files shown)


ArthurZucker (approved):
Thanks for catching this! 

Benchmarked locally on the llama-3 tokenizer (vocab=128,256, added=256), PR cuts get_vocab_size(true) from ~5.70 ms → ~1.35 µs (≈4220×). 
Since non_overlapping = added.len() - overlapping is fully determined by what we've already inserted, we can cache it as a counter on AddedVocabulary
get_vocab_size(true) in O(1) without iterating added tokens at all. 
Will let add_tokens skip `added_tokens_map_r.keys().max()`.
Benched and: `get_vocab_size(true)` drops further to ~786 ps on llama-3.

The counter is model-dependent / stateful so we might want to do that another time as there are some edgecases.

ArthurZucker:
Can you pull latest from main BTW !

eunseo9311:
Thanks for the review and the benchmark numbers!

Rebased on latest main (1da56fa6) — already up to date. The caching idea with a counter on AddedVocabulary sounds great for a follow-up.

eunseo9311:
Hey @ArthurZucker,  thanks again for reviewing and merging this. I've been looking for more ways to contribute to tokenizers — are there any open issues or areas of the codebase that could use extra hands? Happy to pick something up.

PR #2079, opened 2026-05-29
Ci test fixes

sebpop:
Two pre-existing CI failures on main, neither related to any code change in tokenizers/ or bindings/{python,node}/src/:

tests/python: use canonical Salesforce/wikitext dataset id — huggingface_hub enforces namespace/name for HF URIs; the bare id "wikitext" (which was relocated on the Hub) now produces HfUriError: Invalid HF URI 'hf://datasets/wikitext@/.huggingface.yaml'. Updates both datasets.load_dataset(...) call sites in test_tutorial_train_from_iterators.py.

tests/node: update bertProcessing error message to match napi-rs — napi-rs now reports "Failed to get Array length" for invalid array args (was "Given napi value is not an array"). Updates the assertion in post-processors.test.ts.

Each is a one-line / two-line change in tests only; no production code touched.

### DIFF
--- a/bindings/node/lib/bindings/post-processors.test.ts
+++ b/bindings/node/lib/bindings/post-processors.test.ts
@@ -9,7 +9,7 @@ describe('bertProcessing', () => {
   })
 
   it('throws if only one argument is provided', () => {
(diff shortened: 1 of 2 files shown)


ArthurZucker (approved):
ty

PR #2081, opened 2026-06-01
Complete incomplete Viterbi lattice tests

eunseo9311:
The `test_viterbi` test in `lattice.rs` was marked `XXX: incomplete` —
it only checked `.len()` without verifying node IDs or token content.

This PR completes it and adds three new test cases:

- **`test_viterbi`**: verify node IDs and tokens returned by `viterbi()`,
  and confirm score-based path selection picks the higher-score path
- **`test_viterbi_multibyte`**: lattice over a multi-byte UTF-8 string
  (`"テスト"`), including a multi-character spanning node
- **`test_viterbi_empty`**: empty string edge case
- **`test_sample_smoke`**: structural validity check for `sample()` —
  sampled tokens must always concatenate back to the original input

### DIFF
--- a/tokenizers/src/models/unigram/lattice.rs
+++ b/tokenizers/src/models/unigram/lattice.rs
@@ -593,14 +593,77 @@ mod tests {
     #[test]
     fn test_viterbi() {
         let mut lattice = Lattice::from("ABC", 1, 2);
+
+        // Empty lattice → no path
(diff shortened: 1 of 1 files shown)


eunseo9311:
Friendly bump — happy to revise if anything needs adjustment!

PR #2089, opened 2026-06-07
fix(typing): correct encode() input typing (PreTokenizedInputSequence tuple + stub Any)

### LINKED ISSUE #2088: Two incorrect type annotations affect input typing to tokenizers.Tokenizer.encode()
# Summary

I found two issues with the type annotations. Both are easy to understand and can be fixed easily.

## The first type annotation issue

I make this mistake all the time, so I recognize it from my own experience. The intent of the `PreTokenizedInputSequence` type annotation is to describe the data type of a function/method parameter that must be a list or tuple of strings. The mistake is that the correct type annotation for an arbitrary length tuple containing only strings is actually `Tuple[str, ...]`, not `Tuple[str]`. You can read about this in the official documentation here: https://typing.python.org/en/latest/spec/tuples.html.

Here is the code as it is today:

https://github.com/huggingface/tokenizers/blob/2ab6430a316ce355cca4f5de881d1752aa796630/bindings/python/py_src/tokenizers/__init__.py#L21

And below is the surgical fix. This will allow correct programs to pass type checking when they pass tuples of strings into `tokenizers`:
```Python
PreTokenizedInputSequence = Union[List[str], Tuple[str, ...]]
```

If you don't care about the difference between list and tuple, then you can often accept any sequence type like this:
```Python
# At the top of the file:
from collections.abc import Sequence
# ... 
PreTokenizedInputSequence = Sequence[str]
```

### How this problem affects users

Say a user wants to:
 1. Tokenize pre-tokenized text.
 2. Pass in the pre-tokenized text as a tuple of strings.
 3. Use only code that passes Mypy type checking.

Then they can *only* do this because the typing issue above is masked by the second typing issue in `tokenizers/__init__.pyi`. See below.

## The second typing issue that masks the first

In this code, the `sequence` argument has the annotation `Any`, which is not correct and not consistent with the doc string below that on line 948 that indicates the type should be `tokenizers.InputSequence`. The value of `Any` is overly permissive and masks the issue above.

Here is the code as it is today:

https://github.com/huggingface/tokenizers/blob/2ab6430a316ce355cca4f5de881d1752aa796630/bindings/python/py_src/tokenizers/__init__.pyi#L930

And here is the fix:
```Python
        self, /, sequence: InputSequence, pair: Any | None = None, is_pretokenized: bool = False, add_special_tokens: bool = True
```

### How this second typing issue affects users

Here is an example script with an obvious type issue:
``` Python
# fail.py
from math import pi
tokenizer = Tokenizer.from_pretrained("bert-base-uncased")
tokenizer.encode(pi)
```

As expected, this fails with a TypeError:
```bash
❯ uvx --with tokenizers==0.23.1 python fail.py
  File "/home/james/temp.py", line 4, in <module>
    tokenizer.encode(pi)
    ~~~~~~~~~~~~~~~~^^^^
TypeError: TextInputSequence must be str
```

The second typing issue prevents Mypy from identifying this problem:
```bash
❯ uvx --with tokenizers==0.23.1 mypy temp.py
Success: no issues found in 1 source file
```

Anai-Guo:
## What

Fixes the two input-typing issues reported in #2088 so that type checkers correctly type `Tokenizer.encode()` (and `async_encode()`).

**1. `PreTokenizedInputSequence` (`__init__.py`)** — was `Union[List[str], Tuple[str]]`. `Tuple[str]` means a tuple of **exactly one** `str`, so type checkers reject correct programs that pass a pre-tokenized tuple of any other length. The intended meaning is an arbitrary-length tuple of strings, spelled `Tuple[str, ...]` ([typing spec](https://typing.python.org/en/latest/spec/tuples.html)). This also matches the docstring ("A `Tuple` of `str`") and the `List[str]` alternative in the same union.

```python
PreTokenizedInputSequence = Union[List[str], Tuple[str, ...]]
```

**2. `Tokenizer.encode` / `async_encode` stub (`__init__.pyi`)** — the stub annotated `sequence: Any` (and `pair: Any | None`). Because a `.pyi` stub takes precedence over `__init__.py` for type checkers, this `Any` **masked** issue 1 entirely and let invalid calls like `tokenizer.encode(3.14)` pass `mypy` even though they fail at runtime with `TypeError: TextInputSequence must be str`.

The stub now defines the input-sequence aliases (mirroring `__init__.py`) and types the `sequence`/`pair` parameters as `InputSequence`:

```python
TextInputSequence = str
PreTokenizedInputSequence = list[str] | tuple[str, ...]
InputSequence = TextInputSequence | PreTokenizedInputSequence
```

This is consistent with the existing docstrings, which already document `sequence (~tokenizers.InputSequence)`.

## Why

Without these, `mypy` cannot catch invalid `encode()` inputs, and correct programs that pass tuple-of-str pre-tokenized input are wrongly rejected.

Closes #2088.

### DIFF
--- a/bindings/python/py_src/tokenizers/__init__.py
+++ b/bindings/python/py_src/tokenizers/__init__.py
@@ -18,7 +18,7 @@ Offsets = Tuple[int, int]
 TextInputSequence = str
 """A :obj:`str` that represents an input sequence """
 
-PreTokenizedInputSequence = Union[List[str], Tuple[str]]
+PreTokenizedInputSequence = Union[List[str], Tuple[str, ...]]
 """A pre-tokenized input sequence. Can be one of:
 
     - A :obj:`List` of :obj:`str`
--- a/bindings/python/py_src/tokenizers/__init__.pyi
+++ b/bindings/python/py_src/tokenizers/__init__.pyi
@@ -12,6 +12,10 @@ from tokenizers.processors import PostProcessor
 from tokenizers.trainers import Trainer
 from typing import Any, Final, final
 
+TextInputSequence = str
+PreTokenizedInputSequence = list[str] | tuple[str, ...]
+InputSequence = TextInputSequence | PreTokenizedInputSequence
+
 __version__: Final[str]
 
 @final
@@ -737,7 +741,7 @@ class Tokenizer:
             :obj:`List[str]`: A list of decoded strings
         """
     def async_encode(
-        self, /, sequence: Any, pair: Any | None = None, is_pretokenized: bool = False, add_special_tokens: bool = True
+        self, /, sequence: InputSequence, pair: InputSequence | None = None, is_pretokenized: bool = False, add_special_tokens: bool = True
     ) -> Any:
         """
         Asynchronously encode the given input with character offsets.
@@ -927,7 +931,7 @@ class Tokenizer:
                 Truncate direction
         """
     def encode(
-        self, /, sequence: Any, pair: Any | None = None, is_pretokenized: bool = False, add_special_tokens: bool = True
+        self, /, sequence: InputSequence, pair: InputSequence | None = None, is_pretokenized: bool = False, add_special_tokens: bool = True
     ) -> "Encoding":
         """
         Encode the given sequence and pair. This method can process raw text sequences


JEHoctor:
Thank you @Anai-Guo for taking this up. This PR will surgically address the first type annotation issue I reported.

Maybe you can include your note on the second type annotation issue in the issue discussion as well?

Anai-Guo:
Done — I've posted the writeup on the second (stub-generator) issue in #2088 so it's captured in the discussion thread. Happy to follow up with a separate PR on the `tools/stub-gen` side if that direction is wanted.

qgallouedec:
This contribution comes from a profile that opens a massive number of pull requests generated purely by AI; I'm closing it.

PR #2098, opened 2026-06-12
Fix empty Encoding.overflowing when truncation is enabled

### LINKED ISSUE #2091: Release 0.23.1 introduces breaking change in Encoding.overflowing
Release 0.23.1 introduced a performance enhancement:

> Truncation early exit (https://github.com/huggingface/tokenizers/pull/1990)
> Right-direction truncation no longer pre-tokenizes past max_length. The new truncation_benchmark doesn't exist on v0.22.2 so there's no apples-to-apples here, but the PR's own measurements on the same machine showed −20–28% across a range of max_length values for right-truncation; left-truncation unchanged.

This change alters the return value of `overflowing` in the Encoding:

```python

vocab = {"[UNK]": 0, **{str(i): i + 1 for i in range(100)}}
tok = Tokenizer(WordLevel(vocab=vocab, unk_token="[UNK]"))
tok.pre_tokenizer = Whitespace()
tok.enable_truncation(max_length=10, stride=2)

enc = tok.encode(" ".join(str(i) for i in range(50)))
print(len(enc.overflowing)) # in v0.22.2 this is length 5. in v0.23.1, it is length 0.
```

We use the truncation feature and overflow attribute to conveniently tokenize full sequences. The new release results in a length-0 `enc.overflowing` even when the input text is of sufficient length to trigger overflow.

The docs still describe [Encoding.overflowing](https://huggingface.co/docs/tokenizers/main/en/api/encoding) as the place to find this data:

> A List of overflowing Encoding
> 
> When using truncation, the [Tokenizer](https://huggingface.co/docs/tokenizers/v0.23.0-rc0/en/api/tokenizer#tokenizers.Tokenizer) takes care of splitting the output into as many pieces as required to match the specified maximum length. This field lets you retrieve all the subsequent pieces.

At a minimum I think this should be documented as a breaking change in the release notes, rather than somewhat hidden under performance enhancements. However, if I may, I also wonder whether the behaviour should be reverted to retain Encoding.overflow as-is, and a user provided input argument is needed to opt in to the performance enhancement. I haven't looked at the implementation so I don't know how feasible that is.

discobot:
Fixes #2091.

Since 0.23.1, `do_tokenize` sends every truncating encode through the #1990 `tokenize_with_limit` early exit, which stops tokenizing once `max_length` tokens exist — but the dropped pre-token splits are exactly the tokens `Encoding::truncate` later chunks into `overflowing`. Because the encode pipeline always materializes overflow when it truncates, there is no configuration where the early exit preserves observable behavior.

The fix has the encode pipeline tokenize all pre-tokens again (0.22.2 behavior); `tokenize_with_limit` and the #2072 bindings overrides stay intact, so the #1990 perf win could return behind an explicit opt-in (e.g. a `TruncationParams` flag) — happy to do that as a follow-up if wanted.

New Rust tests for both truncation directions and the issue's repro as a Python binding test fail before the fix (`overflowing` empty) and pass after; the #1990 regression tests still pass.

### DIFF
--- a/bindings/python/tests/bindings/test_tokenizer.py
+++ b/bindings/python/tests/bindings/test_tokenizer.py
@@ -6,8 +6,8 @@ import numpy as np
 import asyncio
 from tokenizers import AddedToken, Encoding, Tokenizer, decoders
 from tokenizers.implementations import BertWordPieceTokenizer
-from tokenizers.models import BPE, Model, Unigram
-from tokenizers.pre_tokenizers import ByteLevel, Metaspace
+from tokenizers.models import BPE, Model, Unigram, WordLevel
+from tokenizers.pre_tokenizers import ByteLevel, Metaspace, Whitespace
 from tokenizers.processors import RobertaProcessing, TemplateProcessing
 from tokenizers.normalizers import Strip, Lowercase, Sequence
 from tokenizers.normalizers import ByteLevel as NormalizerByteLevel
@@ -368,6 +368,23 @@ class TestTokenizer:
 
         output = tokenizer.encode("my name is john", "pair")
         assert output.tokens == ["john", "pair"]
+
+    def test_truncation_populates_overflowing(self):
+        # https://github.com/huggingface/tokenizers/issues/2091
+        vocab = {"[UNK]": 0, **{str(i): i + 1 for i in range(50)}}
+        tokenizer = Tokenizer(WordLevel(vocab=vocab, unk_token="[UNK]"))
+        tokenizer.pre_tokenizer = Whitespace()
+        tokenizer.enable_truncation(max_length=10, stride=2)
+
+        output = tokenizer.encode(" ".join(str(i) for i in range(50)))
+        assert output.ids == [vocab[str(i)] for i in range(10)]
+
+        # The truncated tokens must remain available through `overflowing`,
+        # with consecutive pieces overlapping by `stride` tokens
+        assert len(output.overflowing) == 5
+        assert [piece.ids for piece in output.overflowing] == [
+            [vocab[str(i)] for i in range(start, start + 10)] for start in (8, 16, 24, 32, 40)
+        ]
 
     def test_padding(self):
         tokenizer = Tokenizer(BPE())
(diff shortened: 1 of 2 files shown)


lucadiliello:
Any update on this?

samlkrystof:
**Another downstream failure mode: non-empty but under-covering overflow** (transformers `return_overflowing_tokens`)

Confirming this from the transformers side (transformers 5.17.0, tokenizers 0.23.2). Beyond the empty-`overflowing` case in #2091, the early exit has a subtler failure mode: when the last pre-token split overshoots `max_length`, overflow windows *are* returned — but they contain only the overshoot, not the remaining input. The result looks plausible while silently dropping most of the tokens:

```python

tok = AutoTokenizer.from_pretrained("bert-base-uncased")
text = "this is my long setence " * 10  # 60 tokens

enc = tok(text, truncation=True, max_length=10, stride=3,
          return_overflowing_tokens=True)

len(enc["input_ids"])   # 2 chunks — looks fine
covered = [t for ids in enc["input_ids"] for t in ids if t not in (101, 102)]
len(covered)            # 13 (incl. stride duplicates); the other ~47 tokens
                        # were never tokenized
tok.decode(covered)     # 'this is my long setence this isence this is my long'
```

With tokenizers whose pre-tokenizer produces small splits (e.g. Qwen2's `Split` + `ByteLevel` chain) there is no overshoot at all, so `overflowing` comes back empty and every token past `max_length` is silently dropped:

```python
# self-contained, no downloads
import math
from tokenizers import Tokenizer, models, pre_tokenizers, trainers

words = [f"word{i}" for i in range(500)]
tok = Tokenizer(models.BPE())
tok.pre_tokenizer = pre_tokenizers.ByteLevel()
tok.train_from_iterator([" ".join(words)], trainer=trainers.BpeTrainer(vocab_size=600))

tok.enable_truncation(max_length=10, stride=3)
e = tok.encode(" ".join(words[:80]))  # 160 tokens

len(e.ids)          # 10
len(e.overflowing)  # 0 — expected ~22 windows tiling the remaining 150 tokens
```

So for any caller relying on overflow to *cover* the input (sliding-window inference over long documents — the whole point of `return_overflowing_tokens`), 0.23.1+ silently returns partial coverage whether or not `overflowing` happens to be empty. Would love to see this merged — happy to re-test both repros against this branch if useful.

PR #2099, opened 2026-06-12
perf: add benchmarks for the `decode` API

SBrandeis:
# TL;DR

`.decode()` variants were not monitored / benchmarked in the repo

Adds a new dedicated benchmark (`decode_benchmark`) to measure decode-specific performance

Adds a new criterion group to `ci_benchmark` to exercise `.decode()` variants


ArthurZucker (approved):
ty

PR #2102, opened 2026-06-12
ci: cache HF test data and authenticate Hub downloads to avoid rate limits

SBrandeis:
CI was hitting the Hub anonymously from GitHub-hosted runners on every run, getting rate-limited by per-IP anonymous quotas shared across the whole Actions IP pool.

- Cache test fixtures (and the huggingface_hub/datasets runtime cache in the python job) with actions/cache, keyed on the files that define the download list, so steady-state runs make no Hub requests at all. This also covers fork PRs, which can't read secrets.
- Pass HF_TOKEN (from the HF_HUB_READ_TOKEN secret) to make test, the pytest runtime downloader, and the benchmark `hf download` steps — authenticated requests are rate-limited per token, not per shared IP.
- Add retry flags to the wget/curl fixture downloads so a transient 429 doesn't fail the job.


SBrandeis:
Requires setting HF_HUB_READ_TOKEN as a secret

ArthurZucker (approved):
Nice ty, let's double check security / maybe switch our curl to hf download instead (with hf login if possible?)

ArthurZucker (approved):

PR #2103, opened 2026-06-12
ci: split audit in a separate job

SBrandeis:
Move th cargo audit step (for the python bindings) to a separate step:
- does not early stop build+test steps on different platforms
- easier to parse a failure (the job's name is more explicit about what went wrong)


McPatate (approved):
nice ty!

Issue #1908, opened 2025-12-09
`TemplateProcessing` does not apply `type_id` to overflow encodings


antoine-agthe-unity:
A piece is applied with a specific `type_id`, no matter if it is a `SpecialToken` or a `Sequence`.
But when the `piece` is a `sequence` with some overflows, it only applies the specified `type_id` to the main encoding.
Is that legit?

https://github.com/huggingface/tokenizers/blob/f7db48f532b3d4e3c65732cf745fe62863cbe5fa/tokenizers/src/processors/template.rs#L557

686f6c61:
Hi @antoine-agthe-unity, thanks for reporting this issue.

Indeed this is a bug. When a `sequence` with overflows is processed with a specific `type_id`, it should be applied to both the main encoding and the overflow encodings. Currently it's only being applied to the main encoding, which causes the overflows to have incorrect type_ids.

I've analyzed the code and the problem is in the `PostProcessor::process()` method in `src/tokenizer/mod.rs` (Line 113). When applying type_ids to encodings before post-processing, `encoding.set_sequence_id(i)` is done for the main encoding and its overflows, but `encoding.set_type_ids(vec![i as u32; encoding.len()])` (Line 113) is only applied to the main encoding.

The solution is to add the following block after (Line 113):

(lines 113, 115)

```rust
encoding
    .get_overflowing_mut()
    .iter_mut()
    .for_each(|encoding| encoding.set_type_ids(vec![i as u32; encoding.len()]));
```

I've tested this change with 48 different overflow encodings (4 for single sequence and 44 for pair sequences) and all maintain the correct type_ids. 

Can you also test this change to confirm it works for you?

antoine-agthe-unity:
Looks lgtm!

antoine-agthe-unity:
Oh I found this.
https://github.com/huggingface/tokenizers/blob/b874abe7316b5559e8483fb8e66272715e3024bf/tokenizers/src/processors/roberta.rs#L82-L84

Would it conflict with the fix?

686f6c61:
@antoine-agthe-unity Good catch, four eyes see more than two. You're right that `roberta.rs` also sets `type_ids`, but it has the same bug, it doesn't handle overflows either.

[...]

PR #2110, opened 2026-06-16
fix: propagate type_ids to overflow encodings in post-processors

JSap0914:
## Bug

When encoding a pair where one or both sequences produce overflow chunks (due to truncation), the overflow encodings do not receive the `type_id` declared in the template/post-processor — only the **main** encoding is updated.  This means pair-sequence overflows silently keep `type_id=0` instead of the template-declared value (e.g. `$B:1`).

Reported in #1908.

## Root cause

Three independent call sites call `set_type_ids` on the main encoding but not on `get_overflowing_mut()`:

| File | Location | Effect |
|------|----------|--------|
| `tokenizers/src/processors/template.rs` | `apply_template()`, Piece::Sequence arm | Overflow encodings ignore the template's `type_id` |
| `tokenizers/src/tokenizer/mod.rs` | `PostProcessor::process()` | Overflow encodings miss the sequence-index type_id (already fixed for `set_sequence_id` but not `set_type_ids`) |
| `tokenizers/src/processors/roberta.rs` | `process_encodings()`, all-zero override | Overflow encodings do not receive Roberta's type_id=0 invariant |

## Fix

After each `encoding.set_type_ids(…)` call, mirror the operation onto overflow encodings:

```rust
encoding
    .get_overflowing_mut()
    .iter_mut()
    .for_each(|overflow| overflow.set_type_ids(vec![type_id; overflow.len()]));
```

All three sites are patched consistently.

## Verification

```
cargo test processors -p tokenizers
```

> This PR was created with the assistance of an AI coding tool.

### DIFF
--- a/tokenizers/src/processors/roberta.rs
+++ b/tokenizers/src/processors/roberta.rs
@@ -79,9 +79,14 @@ impl PostProcessor for RobertaProcessing {
         }
 
         // Roberta is weird, and every encoding is type_id=0.
-        encodings
-            .iter_mut()
-            .for_each(|encoding| encoding.set_type_ids(vec![0; encoding.len()]));
+        // Roberta is weird, and every encoding is type_id=0.
+        encodings.iter_mut().for_each(|encoding| {
+            encoding.set_type_ids(vec![0; encoding.len()]);
+            encoding
+                .get_overflowing_mut()
+                .iter_mut()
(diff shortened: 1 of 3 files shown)


JSap0914:
Closing as a duplicate — #2055 already addresses propagating type_ids (and sequence_id) to overflow encodings for the same issue (#1908). Apologies for the overlap; deferring to that PR.

PR #2119, opened 2026-06-18
Tokenizers v1

McPatate:
This PR introduces `tokenizers v1`: 
TLDR:
- 40x faster
- linear scaling with threads
- 6x smaller crate
- more bindings (C, C++, etc)
- faster training
- less peak memory


tokenizers/Cargo.toml
```diff
@@ -1,7 +1,12 @@
+[workspace]
+members = ["tk-encode", "tk-train"]
```
SBrandeis: genuine question: do we want to also add the python bindings to the workspace?
could help with the DX

tokenizers/tk-train/Cargo.toml
SBrandeis, line 1: we commit the lockfile too?

tokenizers/Cargo.toml
McPatate: don't think so, I think it's best we leave the current split as such, but no strong opinion on this, python bindings are a special case binding

ArthurZucker (approved):

PR #2120, opened 2026-06-18
(wip) perf: Reduce allocations during normalization and pre-tokenization

SBrandeis:
BLOCKED: `NormalizedString` and `PreTokenizedString` are part of the Python API. `pyclass` objects can't hold references to external stuff because they can be GCed at anytime. This PR has `NormalizedString` and `PreTokenizedString` reference the input sequence (not own a copy of it), which is incompatible with the Python API as is.

### DIFF
--- a/tokenizers/src/tokenizer/normalizer.rs
+++ b/tokenizers/src/tokenizer/normalizer.rs
@@ -1,5 +1,6 @@
 use crate::pattern::Pattern;
 use crate::{Offsets, Result};
+use std::borrow::Cow;
 use std::ops::{Bound, RangeBounds};
 use unicode_normalization_alignments::UnicodeNormalization;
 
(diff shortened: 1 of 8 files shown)


tokenizers/src/tokenizer/normalizer.rs
```diff
@@ -102,11 +103,11 @@ impl std::fmt::Display for SplitDelimiterBehavior {
...
 /// possible to convert offsets from one referential to the other one easily.
 #[derive(Default, Debug, Clone, PartialEq, Eq)]
-pub struct NormalizedString {
+pub struct NormalizedString<'input> {
     /// The original version of the string, before any modification
-    original: String,
+    original: &'input str,
     /// The normalized version of the string, after all modifications
-    normalized: String,
+    normalized: Cow<'input, str>,
```
SBrandeis, line 110: `NormalizedString` does not clone the input sequence anymore

PR #2123, opened 2026-06-23
chore: refactor byte_level constants in utils

SBrandeis:
# TL;DR

Small cleanup PR, no behavior change.

The byte-to-unicode mapping for GPT-2 style byte-level tokenizers was duplicated in a few places (normalizer, pre-tokenizer, decoder, post-processor), each with its own copy of the lookup tables and the transform loop.

This PR factors it into one new util module (`utils/byte_level.rs`)

## Why bother

Prepares the ground for the byte-level fast path PR I'm working on, which lets most BPE tokenizers to skip the byte-to-unicode projection during encoding.


McPatate (approved):
lgtm!

PR #2127, opened 2026-06-24
perf(ByteLevel) bypass byte-to-unicode projection 

SBrandeis:
# TL;DR

ByteLevel tokenizers normally map every byte to a unicode char (GPT-2's
256-char alphabet) before BPE runs, then BPE looks each char back up in the
vocab. That round-trip is pure overhead at encode time.

This PR adds a fast path that skips it when the pipeline is a plain, well-behaved
byte-level setup:
- skip the byte-to-unicode projection in the pre-tokenizer (we still split with the regex, just keep the raw bytes)
- run BPE on raw bytes, using a precomputed byte -> token id map each

It also saves us from computing alignments

## Enablement guards

This fast path only triggers when the whole Tokenization pipeline is "well-behaved":
- The vocabulary maps all raw bytes (0 to 255) to a token
- No `continuing_subword_prefix` / `end_of_word_suffix` magic in the BPE model
- The pipeline is a byte-level pretokenizer with no-op normalization

The logic to check for eligibility and flip the flag to enable is factored in the Tokenizer struct

## Same output

This PR introduces extensive test suites to ensure the equivalence (same input = same output) of the bypass against the "slow" (legacy, correct) path

## TODO

- [ ] Test the bindings use the fast path too
- [ ] Store the vocab in the UTF-8 byte space to avoid the projection + lookup when ignore_merges is enabled
- [ ] ^ unlocks de-duplicating the cache static


tokenizers/tk-encode/src/tokenizer/mod.rs
```diff
@@ -534,6 +563,57 @@ pub struct TokenizerImpl<M, N, PT, PP, D> {
...
+    M: Model,
+    N: Normalizer,
+    PT: PreTokenizer,
+    PP: PostProcessor,
+    D: Decoder,
+{
+    fn drop(&mut self) {
+        self.tokenizer.refresh_byte_level_bypass();
+    }
+}
```
SBrandeis, line 615: Ensures that the fast path enablement is correctly re-computed after re-training an existing tokenizer

on drop of the smart pointer => call `refresh_byte_level_bypass`

tokenizers/tk-encode/src/tokenizer/mod.rs
```diff
@@ -110,6 +124,15 @@ pub trait Model {
...
     }
+
+    fn refresh_byte_level_bypass(&mut self) {}
+    fn byte_level_bypass_eligible(&self) -> bool {
+        false
+    }
+    fn set_byte_level_bypass_active(&mut self, _active: bool) {}
+    fn byte_level_bypass_active(&self) -> bool {
+        false
+    }
```
SBrandeis, line 135: adding all those blanket implemented trait fns is clunky af, but makes it straightforward to compute the eligibility from the generic TokenizerImpl

tokenizers/tk-encode/src/models/bpe/model.rs
```diff
@@ -585,6 +672,64 @@ impl BPE {
...
+            // "vocabulary space" so we can reuse `self.vocab` unchanged. Storing the vocab in
+            // byte space at load time would drop this alloc and the duplicate byte cache.
+            let mapped_string: String = bytes
+                .iter()
+                .map(|byte| BYTES_CHAR_LOOKUP[*byte as usize])
+                .collect();
+            if let Some(id) = self.vocab.get(&mapped_string) {
+                return Ok(vec![Token::new(*id, mapped_string, (0, bytes.len()))]);
+            }
+        }
```
SBrandeis, line 713: this sucks: we still project into gpt2's unicode space when looking up the vocab in ignore_merges mode

tokenizers/tk-encode/src/models/bpe/model.rs
```diff
@@ -87,6 +88,12 @@ thread_local! {
     /// same rayon worker thread never see each other's entries.
     static BPE_LOCAL_CACHE: RefCell<AHashMap<u64, AHashMap<String, Word>>> =
         RefCell::new(AHashMap::new());
+
+    /// Byte-space sibling of `BPE_LOCAL_CACHE`
+    /// Keyed by raw bytes instead of the byte-level-mapped string. Used by the byte-level bypass path.
+    /// Storing the vocab in byte space at load time allow merging the 2 statics.
+    static BPE_LOCAL_CACHE_BYTES: RefCell<AHashMap<u64, AHashMap<Vec<u8>, Word>>> =
+        RefCell::new(AHashMap::new());
```
SBrandeis, line 96: Annoying to have to duplicate the cache :(

PR #2129, opened 2026-06-30
Poc/fast encode

ArthurZucker:
# Faster added-token matching (inference encode path)

Replaces `tokenizers` main's added-token machinery (daachorse trie + two `AHashMap`s) with a single SIMD-scanned `Buckets`. Compared below vs daachorse (`main`), aho-corasick, and an IREE-style linear-scan matcher. Same matches, aarch64.

### Special-token matching (ns/byte, lower = faster)
TL;DR: ~2–4× faster than daachorse on dense input; the IREE matcher only wins on tiny vocabs like these (it's O(bucket) — see real models below).

| vocab (buckets) | input | POC | IREE | daachorse | aho-c |
|---|---|--:|--:|--:|--:|
| llama3 (1) | chat (sparse) | **0.40** | 0.18 | 1.16 | 0.53 |
| llama3 (1) | code (dense) | **0.54** | 1.06 | 0.94 | 0.53 |
| mixed (2) | chat (sparse) | **0.48** | 0.43 | 1.15 | 0.56 |
| mixed (2) | code (dense) | **0.62** | 2.41 | 1.03 | 1.11 |

### Real model added-token vocabs (long bucket; input = prose + the model's tokens)
TL;DR: on the long buckets real models actually have, the POC is **2.8–22× faster than IREE** (its per-candidate linear scan is O(bucket); the POC is O(1)).

| model | tokens | largest bucket | POC | IREE | daachorse | POC vs IREE |
|---|--:|--:|--:|--:|--:|--:|
| deepseek | 818 | 818 | **0.41** | 8.96 | 1.46 | **22.1×** |
| gemma | 218 | 156 | **0.86** | 2.39 | 1.24 | **2.79×** |
| qwen | 22 | 22 | 0.70 | 0.52 | 1.26 | 0.74× |

### Prefix-length rejection (ns/byte; near-miss input sharing the full prefix)
TL;DR: one `u64` compare rejects any ≤8-byte prefix, so the POC's lead grows with prefix length while IREE re-scans the whole bucket.

| shared prefix | POC | IREE | daachorse |
|---|--:|--:|--:|
| `<\|` (2 B) | **0.38** | 3.20 | 0.89 |
| `<\|fi` (4 B) | **0.28** | 2.67 | 0.89 |
| `<\|im_` (5 B) | **0.26** | 1.89 | 0.92 |

### Added-vocab memory (exact heap)
TL;DR: ~5× less memory — one structure stores the token bytes once, vs HF's three copies + two hash tables.

| tokens | POC | HF main | savings |
|--:|--:|--:|--:|
| 8 | 980 B | 4.75 KB | **4.85×** |
| 256 | 13.9 KB | 68.4 KB | **4.93×** |
| 1,000 | 52.1 KB | 260.5 KB | **5.00×** |
| 10,000 | 523 KB | 2.39 MB | **4.57×** |


tokenizers/tk-encode/src/tokenizer/vocab_store.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    pub fn len(&self) -> usize {
+        self.entries.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.entries.is_empty()
+    }
+
+    pub fn get_vocab(&self) -> Vec<(String, u32)> {
```
McPatate:
```suggestion
    pub fn content(&self) -> Vec<(String, u32)> {
```

tokenizers/tk-encode/src/tokenizer/vocab_store.rs
```diff
@@ -0,0 +1,301 @@
...
+        self.entries
+            .iter()
+            .filter_map(|m| self.id_to_token(m.id).map(|token| (token, m.id)))
+            .collect()
+    }
+
+    /// convenient when we want to re-build a vocab
+    pub fn get_vocab_bytes(&self) -> Vec<(Vec<u8>, u32)> {
```
McPatate:
```suggestion
    pub fn byte_content(&self) -> Vec<(Vec<u8>, u32)> {
```

for_the_blogpost.md
McPatate, line 1: perhaps let's not commit this and simply create a shared notion page or wtv?

tokenizers/tk-encode/src/tokenizer/whitespace.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+    i
+}
+
+/// `\w`-style word char (Unicode), matching HF's regex `\w`: alphanumeric, the connector `_`, or a
+/// combining/other mark. Marks matter — e.g. U+0330 is a word char, which `is_alphanumeric` misses.
+#[inline]
+fn is_word_char(c: char) -> bool {
+    c.is_alphanumeric() || c == '_' || c.is_mark()
+}
```
McPatate: incomplete no?
this is what I have in the `Whitespace` pretokenizer:
```rs
fn is_word_char(ch: char) -> bool {
    use unicode_categories::UnicodeCategories;

    ch.is_alphabetic()
        || ch.is_number_decimal_digit()
        || ch.is_punctuation_connector()
        || ch.is_mark()
        || ch == '\u{200c}' // Zero-Width Non-Joiner
        || ch == '\u{200d}' // Zero-Width Joiner
}
```

tokenizers/tk-encode/src/tokenizer/whitespace.rs
McPatate, line 1: I'm not sure the gains from doing things like this by hand are worth the correctness guarantees of using `char` or unicode_categories

tokenizers/tk-encode/src/added_vocabulary/vocab_store.rs
```diff
@@ -0,0 +1,301 @@
...
+
+type Mphf = PtrHash<u64, Linear>;
+
+// Fixed seeds so a given vocab always hashes identically (the hasher is also stored on the struct,
+// so build and query are guaranteed consistent regardless).
+const SEEDS: [u64; 4] = [
+    0x243F_6A88_85A3_08D3,
+    0x1319_8A2E_0370_7344,
+    0xA409_3822_299F_31D0,
+    0x082E_FA98_EC4E_6C89,
```
McPatate, line 16: where do these come from? Do we have `const fn` that could generate them for us instead of hardcoding them by hand?

tokenizers/tk-encode/src/added_vocabulary/vocab_store.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        // 2. A minimal perfect hash needs distinct keys. Collisions are astronomically unlikely
+        //    (~n^2/2^65); if one ever fires, switch the key type to u128. The byte check below makes
+        //    a collision a correct miss at query time, but it would drop a token at build, so guard.
+
+        let mut seen = HashSet::with_capacity(n);
+        for k in &keys {
+            let overlap = seen.insert(*k);
+            if !overlap {
+                println!(
```
McPatate, line 105: `println!` ? `eprintln!` perhaps? We could use `tracing` for logging

McPatate: shouldn't we error out instead of printing?

tokenizers/tk-encode/src/added_vocabulary/vocab_store.rs
```diff
@@ -0,0 +1,301 @@
...
+                id: 0
+            };
+            n
+        ];
+        let mut id_to_slot = vec![u32::MAX; max_id as usize + 1];
+        for (s, id) in &tokens {
+            assert!(
+                s.len() <= u16::MAX as usize,
+                "token longer than 65535 bytes"
+            );
```
McPatate, line 137: return an `Err` instead of assert?

tokenizers/tk-encode/src/added_vocabulary/vocab_store.rs
```diff
@@ -0,0 +1,301 @@
...
+    pub fn new() -> Self {
+        let empty: [u64; 0] = [];
+        Self {
+            mphf: PtrHash::<u64, Linear>::new(&empty, PtrHashParams::default_fast()),
+            hasher: RandomState::new(),
+            bytes: Box::new([]),
+            entries: Box::new([]),
+            id_to_slot: Box::new([]),
+        }
+    }
```
McPatate, line 166: Not sure we want to create empty `VocabStore`s with a `new`? what's the usecase?

tokenizers/tk-encode/src/tokenizer/vocab_store.rs
```diff
@@ -0,0 +1,301 @@
...
+    pub fn get_bytes(&self, q: &[u8]) -> Option<u32> {
+        if self.entries.len() == 0 {
+            return None;
+        }
+        let slot = self.mphf.index_single_part(&self.hasher.hash_one(q));
+
+        let e = self.entries[slot];
+        let (start, len) = (e.start as usize, e.len as usize);
+        // Byte equality: confirms `q` really is the token at this slot (perfect hashing only
+        // guarantees a valid slot for in-vocab keys; this rejects collisions and OOV queries).
```
McPatate: OOV == "out of vocab"?

tokenizers/tk-encode/src/added_vocabulary/vocab_store.rs
```diff
@@ -0,0 +1,301 @@
...
+    #[inline]
+    pub fn get_bytes(&self, q: &[u8]) -> Option<u32> {
```
McPatate, line 172: why do we need to wrap the result in an `Option` here?

McPatate: what does get_bytes do?

tokenizers/tk-encode/src/added_vocabulary/vocab_store.rs
```diff
@@ -0,0 +1,301 @@
...
+        }
+        let e = self.entries[slot as usize];
+        let start = e.start as usize;
+        self.bytes.get(start..start + e.len as usize)
+    }
+
+    #[inline]
+    pub fn id_to_token(&self, id: u32) -> Option<String> {
+        self.id_to_token_bytes(id)
+            .map(|b| String::from_utf8_lossy(b).into_owned())
```
McPatate, line 210: why `from_utf8_lossy`?

tokenizers/tk-encode/src/models/bpe/model.rs
```diff
@@ -610,14 +609,20 @@ impl Model for BPE {
...
+        self.vocab.token_to_id(token)
     }
 
     fn id_to_token(&self, id: u32) -> Option<String> {
-        self.vocab_r.get(&id).cloned()
+        self.vocab.id_to_token(id)
     }
 
     fn save(&self, folder: &Path, name: Option<&str>) -> Result<Vec<PathBuf>> {
+        let vocab_r: VocabR = self
```
McPatate, line 625: why do we still need this? thought the `VocabStore` handled both ways of converting?

tokenizers/tk-encode/src/added_vocabulary/added_vocabulary.rs
```diff
@@ -87,42 +85,25 @@ impl Default for AddedToken {
         }
     }
 }
+
+impl From<&AddedToken> for AddedTokenFlags {
```
McPatate, line 130: note for later, this seems redundant, might require clean up

tokenizers/tk-encode/src/tokenizer/added_vocabulary.rs
```diff
@@ -236,13 +209,8 @@ impl AddedVocabulary {
...
     /// `content` is returned.
-    pub fn simple_id_to_token(&self, id: u32) -> Option<String> {
-        self.added_tokens_map_r.get(&id).map(|t| {
-            self.normalized_cache
-                .get(&id)
-                .cloned()
-                .unwrap_or_else(|| t.content.clone())
-        })
+    pub fn simple_id_to_token(&self, _id: u32) -> Option<String> {
+        Some(String::new())
```
McPatate: no `TODO:` here?

tokenizers/tk-encode/src/tokenizer/added_vocabulary.rs
```diff
@@ -256,7 +224,10 @@ impl AddedVocabulary {
 
     /// Check if a token is a special token
     pub fn is_special_token(&self, token: &str) -> bool {
-        self.special_tokens_set.contains(token)
+        if let Some(tok) = self.vocab.token_to_id(token) {
+            return self.token_metadata[tok as usize].special;
+        }
+        return true;
```
McPatate: `true`? not `false`?

tokenizers/tk-encode/src/tokenizer/added_vocabulary.rs
```diff
@@ -273,294 +244,140 @@ impl AddedVocabulary {
     pub fn add_tokens<N: Normalizer>(
         &mut self,
         tokens: impl IntoIterator<Item = AddedToken>,
-        model: &impl Model,
+        _model: &impl Model,
```
McPatate: why is this unused now?

tokenizers/tk-encode/src/added_vocabulary/added_vocabulary.rs
```diff
@@ -273,294 +244,140 @@ impl AddedVocabulary {
...
-                if self.added_tokens_map_r.get(id) == Some(&token) {
-                    ignored += 1;
-                    continue;
+            let flags = AddedTokenFlags::from(&token);
+            // the form this token is keyed/matched by, in its matcher
+            let form: String = if flags.normalized {
+                if let Some(n) = normalizer {
+                    let mut s = NormalizedString::from(token.content.as_str());
+                    n.normalize(&mut s)?;
+                    s.get().to_string()
```
McPatate, line 328: 😢

tokenizers/tk-encode/src/added_vocabulary/added_vocabulary.rs
```diff
@@ -273,294 +244,140 @@ impl AddedVocabulary {
...
-    /// Extract the additional vocabulary from the given sentence, normalizing it along the way.
-    ///
-    /// Some tokens should match against their normalized representation, as well as the
-    /// non-normalized one. For example, when we expect to extract the token `yesterday` in the
-    /// input sentence `I read a book Yesterday`, if the normalizer is supposed to lowercase
-    /// everything, we expect a match.
     pub fn extract_and_normalize<N: Normalizer>(
         &self,
-        normalizer: Option<&N>,
+        _normalizer: Option<&N>,
```
McPatate, line 383: why is this unused?

tokenizers/tk-encode/src/added_vocabulary/added_vocabulary.rs
```diff
@@ -273,294 +244,140 @@ impl AddedVocabulary {
...
+                    search = match_end;
+                }
+                // since match_bytes goes to the end, this means we reach the end.
+                _ => break,
+            }
+        }
+        if emit < bytes.len() {
+            splits.push((emit, bytes.len(), None));
+        }
+        // FIXME: this will go away once we have the 0-allocation in the hot path :)
```
McPatate, line 429: 🥳

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+
+/// We implement a 2-level check:
+/// 1. on the next byte after the prefix.
+/// 2. on the lenght of the added tokens.
+/// 3. on the hashed bytes.
+#[derive(Clone, Debug)]
+pub struct Bucket {
+    // longest common prefix of the bucket
+    pub prefix: Box<[u8]>,
+    // prefix[0..min(len,8)] packed little-endian + a per-byte mask, so the hot-path reject is a
```
McPatate: from ne is used below, maybe update the comment?

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+            inner: VocabStore::new(),
+        }
+    }
+    /// Build the matcher from sorted `tokens`, the first-byte->bucket table, and the buckets.
+    pub fn build(
+        tokens: Vec<(Vec<u8>, u32)>,
+        first_byte_to_bucket_id: [u8; 256],
+        buckets: Box<[Bucket]>,
+    ) -> Self {
+        // tokens should be sorted
```
McPatate: why don't we sort inline here instead of assuming it's pre-sorted?

tokenizers/tk-encode/src/added_vocabulary/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+}
+
+impl Default for Buckets {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[cfg(test)]
+mod bench {
```
McPatate, line 456: move to `bench/`?

tokenizers/tk-encode/src/added_vocabulary/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+    }
+}
+
+impl Default for Buckets {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[cfg(test)]
```
McPatate, line 455: shouldn't be a test?

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+        //   BOTTOM collision: t and d share low nibble 4 (lo16[4] ends up with two bits)
+        first_byte_to_bucket[b't' as usize] = 0;
+        first_byte_to_bucket[b'o' as usize] = 1;
+        first_byte_to_bucket[b'a' as usize] = 2;
+        first_byte_to_bucket[b'd' as usize] = 3;
+
+        let mut expected_hi16 = [0u8; 16];
+        let mut expected_lo16 = [0u8; 16];
+
+        // filling hi goes from 0..16. First match gets the first id.
```
McPatate:
```suggestion
        // filling hi goes from 0..=0xF. First match gets the first id.
```

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+    pub special: bool,
+    pub normalized: bool,
+    pub single_word: bool,
+    pub lstrip: bool,
+    pub rstrip: bool,
+}
+
+/// We implement a 2-level check:
+/// 1. on the next byte after the prefix.
+/// 2. on the lenght of the added tokens.
```
McPatate:
```suggestion
/// 2. on the length of the added tokens.
```

I'm doing a pass on typos to avoid all the ai generated prs 😬

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+#[derive(Clone, PartialEq, Debug)]
+pub struct AddedTokenFlags {
+    pub special: bool,
+    pub normalized: bool,
+    pub single_word: bool,
+    pub lstrip: bool,
+    pub rstrip: bool,
+}
+
+/// We implement a 2-level check:
```
McPatate: May be worth explaining a bit more in detail what we do here; also, "2-level check" and you list 3 steps

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+/// 3. on the hashed bytes.
+#[derive(Clone, Debug)]
+pub struct Bucket {
+    // longest common prefix of the bucket
+    pub prefix: Box<[u8]>,
+    // prefix[0..min(len,8)] packed little-endian + a per-byte mask, so the hot-path reject is a
+    // single `(chunk ^ prefix_word) & prefix_mask` instead of walking the boxed prefix slice.
+    pub prefix_word: u64,
+    pub prefix_mask: u64,
+    pub next_byte_to_length_id: [u16; 256], // post-prefix byte -> sub-list index; 0xFFFF = none
```
McPatate: why u16 btw?

McPatate: `len(token) > 65k` unlikely / rejected?

tokenizers/tk-encode/src/added_vocabulary/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+#[derive(Clone, Debug)]
+pub struct Bucket {
+    // longest common prefix of the bucket
+    pub prefix: Box<[u8]>,
+    // prefix[0..min(len,8)] packed little-endian + a per-byte mask, so the hot-path reject is a
+    // single `(chunk ^ prefix_word) & prefix_mask` instead of walking the boxed prefix slice.
+    pub prefix_word: u64,
+    pub prefix_mask: u64,
+    pub next_byte_to_length_id: [u16; 256], // post-prefix byte -> sub-list index; 0xFFFF = none
+    pub length_list: Box<[Box<[u16]>]>,
```
McPatate, line 34: why is that a `Box<[Box<[u16]>]>`?

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+    pub length_list: Box<[Box<[u16]>]>,
+}
+
+impl Bucket {
+    /// Pack the prefix into `prefix_word`/`prefix_mask` for the one-instruction reject.
```
McPatate: is that really a 1-instruction reject? 🤓

tokenizers/tk-encode/src/added_vocabulary/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+            length_list,
+        }
+    }
+}
+
+/// Buckets are responsible of finding matches.
+/// lo / hi nibble
+#[derive(Clone)]
+pub struct Buckets {
+    first_byte_to_bucket_id: [u8; 256],
```
McPatate, line 81: what is that exactly?

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+        }
+    }
+}
+
+/// Buckets are responsible of finding matches.
+/// lo / hi nibble
+#[derive(Clone)]
+pub struct Buckets {
+    first_byte_to_bucket_id: [u8; 256],
+    // we use optimized SIMD to scan potential special tokens when there are more than 1 bucket.
```
McPatate:
```suggestion
    // we use optimized SIMD instructions (main one being `pshufb`) to scan potential special tokens when there are multiple unique starting bytes, noting that we have one bucket per starting byte.
```

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+    // this is taken from http://0x80.pl/notesen/2018-10-18-simd-byte-lookup.html
```
McPatate:
```suggestion
    // The algo is taken from http://0x80.pl/notesen/2018-10-18-simd-byte-lookup.html, specifically the "Special case 1 - small sets"
```

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+    lo16: [u8; 16],
+    hi16: [u8; 16],
+    #[allow(dead_code)] // TODO: honor in the scan — fall back to scalar when false
```
McPatate:
```suggestion
    #[allow(dead_code)] // TODO: honor in the scan — fall back to  universal algo, alternative impl (with fewer instructions)
```

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+    can_use_nibbling: bool, // we can't if there are >8 high nibbles (very rare)
```
McPatate: to remove once we have the universal algo impl as a fallback. will need to dev the heuristic at build time on which to choose. Perhaps something like:
```rs
enum MatchKernel {
  Universal,
  SmallSet,
  // NOTE: open to contributions to add constant nibble and unique higher and lower nibble special case variants
```
could help

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+    buckets: Box<[Bucket]>,
+    inner: VocabStore,
```
McPatate:
```suggestion
    vocab: VocabStore,
```
or 
```suggestion
    vocab_store: VocabStore,
```

Also, why does `Buckets` have a `VocabStore` attached to it?

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+    // e │ . . . . . . . . . . . . . . . .
+    // f ╵ . . . . . . . . . . . . . . . .
+    // If you have the following first bytes
+    // "<"  = 0x3c  -> col 3, row c
+    // "["  = 0x5b  -> col 5, row b
+    // "\t" = 0x09  -> col 0, row 9
+    // "\n" = 0x0a  -> col 0, row a
+    // The reason we have lo and hi is because for SIMD
+    // we can't have the index be 16 bits, so lo has 8bits length, hi as well.
+    // The key is also that we don't store the 16x16 grid, just 2x16 u8 (since u8 is up to 8 values)
```
McPatate:
```suggestion
    // `pshufb` maps bytes to bytes, so to have a 16bit row  bitmap, we need to split it into two 16x8bit bitmaps. We need two calls to `pshufb` to pull one row of the bitmap.
```

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+                // longest prefix.
+                let t = &tokens[i as usize].0;
+                min_len = min_len.min(t.len());
+                let common = lcp.iter().zip(t).take_while(|(a, b)| a == b).count();
+                lcp = &lcp[..common];
+            }
+            // cap so prefix.len() < every token's len => every token has a byte at prefix.len()
+            let plen = lcp.len().min(min_len.saturating_sub(1));
+            let prefix: Box<[u8]> = lcp[..plen].to_vec().into_boxed_slice();
+            // disc table (byte after prefix -> sub-list) + per-disc distinct lengths, longest first
```
McPatate: what's a `disc table`?

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+    // "<"  = 0x3c  -> col 3, row c
+    // "["  = 0x5b  -> col 5, row b
+    // "\t" = 0x09  -> col 0, row 9
+    // "\n" = 0x0a  -> col 0, row a
+    // The reason we have lo and hi is because for SIMD
+    // we can't have the index be 16 bits, so lo has 8bits length, hi as well.
+    // The key is also that we don't store the 16x16 grid, just 2x16 u8 (since u8 is up to 8 values)
+    // So the lo[3] = 0b00100100 ->
+    //                  ..x..x..
+    // store the entire row in the u8 format. (0-255)
```
McPatate: in the above example, `lo[3] = 0b00000000`. perhaps we can update that with:
```suggestion
    // So the lo[0xc] = 0b00010000 ->
    //                  ...x....
    // stores half the row in the u8 format. (0-255)
```

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+        let (mut lo, mut hi) = ([0u8; 16], [0u8; 16]);
+        let (mut next, mut bit, mut has) = (0u32, [0u8; 16], [false; 16]);
+        for h in 0..16 {
+            // h<<4 | l is just computing 0x{h}{l} indexed into the candidates
+            // here for each match, we update the count and fill the table with the
+            // unique id associated with that match.
+            if (0..16).any(|l| candidates[(h << 4) | l]) {
+                if next >= 8 {
+                    return;
+                } // >8 high nibbles -> caller uses scalar TODO:
```
McPatate: as mentioned before, we should have multiple implementations and decide ahead of time which one to use

tokenizers/tk-encode/src/added_vocabulary/buckets.rs
```diff
@@ -0,0 +1,740 @@
...
+                    if candidates[(h << 4) | l] {
+                        lo[l] |= bit[h];
+                    }
+                }
+            }
+        }
+        (self.lo16, self.hi16) = (lo, hi);
+    }
+    #[cfg(target_arch = "aarch64")]
+    fn nibble_match_bytes(&self, bytes: &[u8]) -> Option<usize> {
```
McPatate, line 229: would be interesting to note somewhere that a nibble is half a byte, and that we use nibbles because pshufb does parallel lookup in a 16-byte register, using 4-bit indices.

McPatate (commented):
submitting a first pass review, will finish reviewing in detail the algorithm impl tmrw

tokenizers/tk-encode/src/models/bpe/model.rs
ArthurZucker: this is in save, I think we can potentially remove it and store the vocabStore itself instead of re-building this but it will be offuscated a bit

ArthurZucker: ah its also legacy, we rarely write the merges and vocab.json nowdays

tokenizers/tk-encode/src/added_vocabulary/buckets.rs
```diff
@@ -0,0 +1,741 @@
...
+                Box::from(*b"<"),
+                next_byte_to_length_id,
+                Box::new([Box::new([7])]),
+            )]),
+        );
+        assert_eq!(
+            fake_vocab.match_bytes(b"This should be kwown<s><|eos|>"),
+            Some((0, 23, 7))
+        );
+        assert_eq!(fake_vocab.match_bytes(b"><|eos|>"), Some((0, 1, 7)));
```
McPatate, line 688: Perhaps we can return a struct instead of a tuple?

tokenizers/tk-encode/src/tokenizer/buckets.rs
```diff
@@ -0,0 +1,741 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[cfg(target_arch = "aarch64")] // calls the NEON-only nibble_match_bytes
+    #[test]
+    fn test_build_nibble_table() {
+        let mut first_byte_to_bucket = [0xFFu8; 256];
+        // spell "toad": t=0x74, o=0x6F, a=0x61, d=0x64
+        //   TOP    collision: o, a, d all have high nibble 6 (same row -> same hi bit)
```
McPatate:
```suggestion
        //   TOP    collision: o, a, d all have high nibble 6 (same row -> same hi nibble)
```

tokenizers/tk-encode/src/added_vocabulary/buckets.rs
```diff
@@ -0,0 +1,741 @@
...
+                    return Some((id, len as u32));
+                }
+            }
+        }
+        None
+    }
+
+    /// Find the leftmost added-token match in `bytes`. Returns (token_id, match_position, match_len).
+    /// Dispatch: 1 bucket -> memchr on the shared first byte; >=2 -> NEON nibble, mask-iterated.
+    pub fn match_bytes(&self, bytes: &[u8]) -> Option<(u32, u32, u32)> {
```
McPatate, line 329: could be a struct `Match { token_id: u32, start: u32, end: u32 }`

tokenizers/tk-encode/src/tokenizer/buckets.rs
ArthurZucker: we no longer need this

tokenizers/tk-encode/src/added_vocabulary/vocab_store.rs
ArthurZucker: no no because we support people passing the same keys twice (bad processing potentially)

tokenizers/tk-encode/src/tokenizer/whitespace.rs
ArthurZucker: done

tokenizers/tk-encode/src/added_vocabulary/vocab_store.rs
ArthurZucker: this allows reproducivility

PR #2134, opened 2026-07-01
refactor: extract pretok algo

McPatate:
When going through the punctuation pretokenizer implementation, I noticed that I would have to reimplement the same kind of algo for each variant of its inner `behavior` field, which is also the same as the Whitespace and Bert algo.

Extracting this into an easy to adapt API for each pretokenizer, while maintaining the perf gains shown by the other two variants.

### DIFF
--- a/tokenizers/tk-encode/src/pre_tokenizers/whitespace.rs
+++ b/tokenizers/tk-encode/src/pre_tokenizers/whitespace.rs
@@ -2,7 +2,7 @@ use std::sync::LazyLock;
 
 use regex::Regex;
 
-use crate::pipeline;
+use crate::pipeline::{self, SplitPolicy};
 use crate::tokenizer::{
     pattern::Invert, PreTokenizedString, PreTokenizer, Result, SplitDelimiterBehavior,
 };
@@ -48,6 +48,18 @@ enum CharType {
     Symbol,
 }
 
+impl CharType {
+    #[inline]
+    fn policy(self) -> SplitPolicy {
+        match self {
+            // whitespace is dropped; word and symbol groups are each emitted as one
+            // split, with a boundary between the two classes.
+            CharType::Whitespace => SplitPolicy::Remove,
+            CharType::Word | CharType::Symbol => SplitPolicy::Keep,
+        }
+    }
+}
+
 fn classify(ch: char) -> CharType {
     if ch.is_whitespace() {
         CharType::Whitespace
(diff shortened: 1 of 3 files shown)


ArthurZucker (approved):
feels like policy is not needed and classify alone can do this. 
Otherwise makes a lot of sense.its super impressive that this just gives 8x. 
Pushing to out is fine as long as out is pre allocated for worst case IMO!

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -20,3 +20,57 @@ pub trait PreTokenizer {
...
+/// `classify` maps each char to a small `Copy + Eq` class, the current
+/// split ends whenever the class changes (or on every char of an `Isolate`
+/// class), and `policy` decides what becomes of it. Ranges are byte offsets
+/// into `text`.
+#[inline(always)]
+pub fn split<C: Copy + PartialEq>(
+    text: &str,
+    out: &mut Vec<Split>,
+    classify: impl Fn(char) -> C,
+    policy: impl Fn(C) -> SplitPolicy,
```
SBrandeis, line 359: Do we want generics instead here?

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -20,3 +20,57 @@ pub trait PreTokenizer {
...
+) {
```
SBrandeis, line 360: Could be a blanket implementation in the pipeline::PreTokenizer trait, wdyt?

tokenizers/tk-encode/src/pre_tokenizers/whitespace.rs
```diff
@@ -72,48 +84,25 @@ fn is_word_char(ch: char) -> bool {
...
-/// `CharType` of each ASCII codepoint (index == its byte value). Lets the
-/// ASCII fast path skip the Unicode predicates in `classify`; the (cheap) UTF-8
-/// decode still happens via `char_indices`, and non-ASCII falls back to
-/// `classify`.
 static ASCII_CLASS: LazyLock<[CharType; 128]> =
     LazyLock::new(|| std::array::from_fn(|b| classify(b as u8 as char)));
 
 impl pipeline::PreTokenizer for Whitespace {
+    // XXX: surprisingly, inlining here yields 10-15% slower performance
+    #[inline(never)]
```
SBrandeis, line 92: Why not letting the compiler handle this?

McPatate: We're not sure it'll pick the right choice. Here I've tested and observed that if this code is inlined, perf degrades, so we tell the compiler to make sure it always makes the right decision.

tokenizers/tk-encode/src/tokenizer/pipeline.rs
McPatate: Meh, the fn knows nothing about `PreTokenizer`, doesn't need `self` so I'd rather leave it as its own separate thing.

tokenizers/tk-encode/src/tokenizer/pipeline.rs
McPatate: What do you have in mind?

PR #2138, opened 2026-07-01
feat: `impl pipeline::PreTokenizer for UnicodeScripts`

McPatate:
Not relying on `pipeline::split` because of the `Script::Any` variant, which requires specific logic for defining the end of a given split.

```
unicode_scripts (ascii):  (10000 splits)
  old (PreTokenizedString)     3463876 ns/iter
  new (Vec<Split>)              225123 ns/iter   15.39x

multilingual corpus: 162000 bytes,  100 iters/measurement

unicode_scripts (mixed):  (11000 splits)
  old (PreTokenizedString)     3805905 ns/iter
  new (Vec<Split>)             2015314 ns/iter    1.89x
```

### DIFF
--- a/tokenizers/tk-encode/src/pre_tokenizers/unicode_scripts/pre_tokenizer.rs
+++ b/tokenizers/tk-encode/src/pre_tokenizers/unicode_scripts/pre_tokenizer.rs
@@ -79,8 +79,9 @@ impl PreTokenizer for UnicodeScripts {
     }
 }
 
-static ASCII_CLASS: LazyLock<[Script; 128]> =
-    LazyLock::new(|| std::array::from_fn(|b| fixed_script(b as u8 as char)));
+static BMP_SCRIPT: LazyLock<[Script; 0x10000]> = LazyLock::new(|| {
+    std::array::from_fn(|i| char::from_u32(i as u32).map_or(Script::Common, fixed_script))
+});
 
 impl pipeline::PreTokenizer for UnicodeScripts {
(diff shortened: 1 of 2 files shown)


McPatate:
Extending the lookup table to the basic multilingual plane characters yields much better perf:
```
unicode_scripts (ascii):  (10000 splits)
  old (PreTokenizedString)     3434903 ns/iter
  new (Vec<Split>)              109613 ns/iter   31.34x

multilingual corpus: 162000 bytes,  100 iters/measurement

unicode_scripts (mixed):  (11000 splits)
  old (PreTokenizedString)     3773279 ns/iter
  new (Vec<Split>)              109924 ns/iter   34.33x
```
note that each corpus is created like so:
```rs
fn corpus() -> String {
    let s = "The quick brown fox jumps over the lazy dog. How are you doing today? \
             Numbers 123 456 and symbols @#$%, plus café résumé naïve. ";
    s.repeat(1000) // ~130 KB, ASCII-heavy with a little Unicode
}

fn multilingual_corpus() -> String {
    let s = "The quick brown fox 123. こんにちは世界、お元気ですか。 \
             你好世界 привет мир как дела مرحبا بالعالم café résumé. ";
    s.repeat(1000)
}
```

there could be alternate corpuses to test on for completeness purposes, but imo this is good signal already. This pretokenizer is a bit special in that the old `PreTokenizedString` allocations weren't really the big perf bottleneck, the huge `get_script` fn is what stalls the process.

Also, the table is 64kb, so we have to make sure we're ok with the memory overhead @ArthurZucker

PR #2143, opened 2026-07-02
feat: add `MultiRegex` for common regexes (gpt2,etc) fast path

McPatate:
add the DFA fast path for specific common regexes

used it to update the `Split` pretokenizer, perf is looking good there, ~8x

but we'll use @SBrandeis 's new benchmarks to validate

will also rebase when #2134 is merged


tokenizers/tk-encode/src/pre_tokenizers/split.rs
```diff
@@ -103,6 +113,32 @@ impl PreTokenizer for Split {
     }
 }
 
+impl pipeline::PreTokenizer for Split {
+    fn pre_tokenize(&self, text: &str, out: &mut Vec<pipeline::Split>) -> Result<()> {
+        let matches: Vec<((usize, usize), bool)> = match &self.multi {
+            Some(multi) => {
+                let mut segments = Vec::new();
```
ArthurZucker: please alocate with capacity text.len()

tokenizers/tk-encode/src/pre_tokenizers/split.rs
```diff
@@ -192,6 +228,134 @@ mod tests {
...
     }
 
+    fn pipeline_split(
+        pattern: SplitPattern,
+        behavior: SplitDelimiterBehavior,
+        invert: bool,
+        text: &str,
+    ) -> Vec<(&str, (u32, u32))> {
+        let pretok = Split::new(pattern, behavior, invert).unwrap();
+        let mut splits = Vec::new();
```
ArthurZucker: same

Issue #1775, opened 2025-05-06
DebertaV2TokenizerFast and XLMRobertaTokenizerFast has overlapping offsets/charspans which would mess up char_to_token() and token_to_char() conversions


ligz08:
Noticed this when working with `transformers.DebertaV2TokenizerFast` and `XLMRobertaTokenizerFast`.

## My `transformers` and `tokenizers` versions
```python
import transformers, tokenizers
print(f'{transformers.__version__=}')
print(f'{tokenizers.__version__=}')
```
```
transformers.__version__='4.51.3'
tokenizers.__version__='0.21.1'
```

## Repro in Python
```python
samples = ['English language', '中文 中文 中文', '日本語 日本語 日本語', 'русский язык', '한국어 한국어 한국어', 'français français français', '中 文 中 文']
tokenizer = AutoTokenizer.from_pretrained('microsoft/mdeberta-v3-base')
print(f'{type(tokenizer)=}')
encodings = tokenizer(samples)
for i in range(len(samples)):
    print(f'{i=}: {samples[i]}')
    tokens = encodings.tokens(i)
    print(f'{tokens=}')
    print(f'{encodings._encodings[i].offsets=}')
    chars = []
    for j in range(len(tokens)):
        charspan = encodings.token_to_chars(i, j)
        if charspan is not None:
            chars.append((j, encodings['input_ids'][i][j], charspan, samples[i][charspan.start:charspan.end]))
        else:
            chars.append((j, encodings['input_ids'][i][j], charspan, None))
    print(f'{chars=}')
    print(f'{encodings.char_to_token(i, 0)=}')
    print(f'{encodings.token_to_chars(i, encodings.char_to_token(i, 0))=}')
    print(f'{encodings["input_ids"][i][encodings.char_to_token(i, 0)]=}')
    print()
```
outputs:
```
type(tokenizer)=<class 'transformers.models.deberta_v2.tokenization_deberta_v2_fast.DebertaV2TokenizerFast'>
i=0: English language
tokens=['[CLS]', '▁English', '▁language', '[SEP]']
encodings._encodings[i].offsets=[(0, 0), (0, 7), (7, 16), (0, 0)]
chars=[(0, 1, None, None), (1, 5414, CharSpan(start=0, end=7), 'English'), (2, 17897, CharSpan(start=7, end=16), ' language'), (3, 2, None, None)]
encodings.char_to_token(i, 0)=1
[...]

PR #2149, opened 2026-07-05
Map prepended text to a zero-width span to avoid overlapping offsets (#1775)

v-code01:
## Problem

`char_to_token()` can point to the wrong token for tokenizers that use a Metaspace-style prefix (e.g. `DebertaV2TokenizerFast`, `XLMRobertaTokenizerFast`), because token **offsets overlap** on multi-byte scripts. Reported in #1775.

Root cause is in `NormalizedString::prepend`. It mapped the prepended text — which has no counterpart in the original input — onto the **first original character's byte range**. That is invisible while the prefix stays glued to the following text, but once the model splits the prepended marker into its own token it surfaces. For `"中文"` under Metaspace (xlm-roberta), the tokens are `["▁", "中文"]` and the offsets were:

```
▁    (0, 3)   ← borrows 中's bytes
中文  (0, 6)   ← overlaps
```

so `char_to_token(0)` resolves to `▁` instead of `中文`. Reproduces on Chinese and Japanese samples; Latin/Cyrillic/Korean don't split the `▁word` unit, so they never exposed it.

## Fix

Prepended text has no source in the original string, so it should map to a **zero-width span at the start**. `prepend` now inserts the text as new characters at an empty range, leaving the existing characters' offsets untouched:

```
▁    (0, 0)   ← zero-width, synthetic
中文  (0, 6)   ← no overlap
```

`ByteLevel`'s byte-level remap is updated to transform over the full **normalized** range instead of the original range, so the now-zero-width prefix byte (e.g. the prefix space from `add_prefix_space`) is still covered — otherwise it would be excluded and desync the transformation offsets.

## Tests

- New `prepend_multibyte_no_overlap` regression test (the exact CJK + `▁` case).
- Updated the `prepend` and `Prepend`-normalizer alignment expectations to the corrected zero-width behaviour.
- Full suite green (`cargo test`, with `make test` data), including all `byte_level`, `added_tokens`, and `normalizer` tests. `cargo fmt` / `cargo clippy` clean.

Every prepend caller in the tree (Metaspace, ByteLevel's leading space, the `Prepend` normalizer) prepends synthetic text, so zero-width alignment is correct for all of them.

### DIFF
--- a/tokenizers/src/normalizers/prepend.rs
+++ b/tokenizers/src/normalizers/prepend.rs
@@ -41,10 +41,12 @@ mod tests {
             NormalizedString::new(
                 original.to_string(),
                 normalized.to_string(),
+                // The prepended "▁" (3 bytes) maps to a zero-width span at the
+                // start; the original characters keep their own offsets.
                 vec![
-                    (0, 1),
-                    (0, 1),
-                    (0, 1),
+                    (0, 0),
+                    (0, 0),
+                    (0, 0),
                     (0, 1),
                     (1, 2),
                     (2, 3),
@@ -56,7 +58,7 @@ mod tests {
         );
         assert_eq!(
             n.alignments_original(),
-            vec![(0, 4), (4, 5), (5, 6), (6, 7), (7, 8)]
+            vec![(3, 4), (4, 5), (5, 6), (6, 7), (7, 8)]
         );
     }
 }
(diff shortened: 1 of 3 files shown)


v-code01:
cc @ArthurZucker @McPatate — this fixes #1775 (overlapping offsets → `char_to_token` pointing at the wrong token) at its root in `NormalizedString::prepend`: prepended text mapped onto the first original character's bytes, which surfaced as an overlap once the model split the metaspace `▁` into its own token on CJK. Prepend now maps synthetic prefixes to a zero-width span; ByteLevel's byte remap was adjusted to the full normalized range so the prefix space is still covered. Full suite + `make test` data green, fmt/clippy clean.

v-code01:
cc @SBrandeis @McPatate — small, self-contained fix with a regression test: root-cause fix for overlapping offsets (#1775): prepend maps to a zero-width span so char_to_token stays correct for CJK. Would appreciate a look when you have a moment.

PR #2151, opened 2026-07-05
Harden tokenizer deserialization against panics on crafted configs

v-code01:
## What

A `tokenizer.json` is untrusted input, but several components deserialized from one bypass the validation their builders enforce, so a crafted config **panics at load or encode/decode time** instead of returning an error. This PR fixes six such panics (each with a regression test); they share the theme "deserialization skips a check the code later relies on."

| Component | Crafted input | Was | Now |
|-----------|--------------|-----|-----|
| `Encoding::truncate` | `stride >= max_length` via deserialized truncation params | panic (`assert`) | clamped, no panic |
| `TemplateProcessing` | template references a special token absent from `special_tokens` | panic at encode (`HashMap` index) | `Err` at load (routes through `validate`) |
| BPE builder | merge pair whose concatenation exceeds the longest vocab token | OOB panic on scratch buffer | `MergeTokenOutOfVocabulary` |
| `Strip` decoder | crafted `start`/`stop` (e.g. token all `content` chars) | underflow / reversed-range slice panic | clamped, no panic |
| `FixedLength` pre-tokenizer | `length = 0` | `chunks(0)` panic | `Err` |
| `Precompiled` normalizer | invalid `precompiled_charsmap` | `.expect()` panic | serde `Err` (matches every sibling arm) |

## Why one PR

All six are the same class of bug (untrusted-config → panic) and several are one-line fixes; grouping them keeps the review to a single coherent security-hardening pass rather than six fragments.

## Tests

Each fix has a regression test; full lib suite green (208 passed). `cargo fmt` / `cargo clippy --lib -- -D warnings` clean.

cc @SBrandeis @McPatate @ArthurZucker

### DIFF
--- a/tokenizers/src/decoders/strip.rs
+++ b/tokenizers/src/decoders/strip.rs
@@ -42,7 +42,10 @@ impl Decoder for Strip {
                 }
 
                 let mut stop_cut = chars.len();
-                for i in 0..self.stop {
+                // Bound the loop by the token length: `self.stop` comes from the
+                // config unvalidated, so without this `chars.len() - i - 1` would
+                // underflow (and panic / index out of bounds) on short tokens.
+                for i in 0..self.stop.min(chars.len()) {
                     let index = chars.len() - i - 1;
                     if chars[index] == self.content {
                         stop_cut = index;
@@ -52,6 +55,10 @@ impl Decoder for Strip {
                     }
                 }
 
+                // `start_cut` and `stop_cut` can cross when a token is made
+                // entirely of `content` characters; clamp so the range is never
+                // reversed, which would panic when slicing.
+                let stop_cut = stop_cut.max(start_cut);
                 let new_token: String = chars[start_cut..stop_cut].iter().collect();
                 new_token
             })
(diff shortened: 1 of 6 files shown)


v-code01:
cc @ArthurZucker @McPatate — found by auditing untrusted-config panics: `with_truncation` validates `stride`, but a deserialized `tokenizer.json` bypasses it, so `stride >= max_length` panics on encode (the LongestFirst pair path truncates each seq to ~max_length/2, so even `stride < max_length` can trip `assert(stride < max_len)`). Clamps stride instead; primary encoding is unaffected.

v-code01:
cc @SBrandeis @McPatate — small, self-contained fix with a regression test: untrusted tokenizer.json can bypass with_truncation's stride check and panic in Encoding::truncate. Would appreciate a look when you have a moment.

professor-moody:
There is a seventh site in the same class as the six in this table: `DecoderWrapper`'s hand-written `Deserialize`.

https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/decoders/mod.rs#L90

```rust
let helper = DecoderHelper::deserialize(deserializer).expect("Helper");
```

Because this is `.expect` rather than `?`, any `decoder` value the helper cannot deserialize aborts the process instead of returning `Err`. It does not need a semantically interesting payload: the panic happens partway through streaming deserialization, so a truncated or corrupted `tokenizer.json` that merely *contains* a `decoder` key is enough. Every one of the 47 inputs I reduced this from is invalid JSON.

Minimal shape:

```
{"decoder"
	:<garbage bytes>
```

```
thread '<unnamed>' panicked at tokenizers/src/decoders/mod.rs:90:63:
Helper
    #19 deserialize<...> tokenizers/src/decoders/mod.rs:90:63
    #26 <TokenizerVisitor<...> as Visitor>::visit_map tokenizers/src/tokenizer/serialization.rs:141:56
    #29 visit_newtype_struct tokenizers/src/tokenizer/mod.rs:409:21
```

So `Tokenizer::from_bytes` panics on a malformed file rather than returning `Err`, which matters for anything that loads a `tokenizer.json` it did not produce.

The fix looks like the same one-liner as the `Precompiled` arm in this PR: propagate the error instead of `.expect`ing it. Happy to send it as a commit here if you would prefer it folded into this PR rather than a separate one.

Found with a fuzzer (`Tokenizer::from_bytes` on mutated `tokenizer.json`); this comment was drafted with AI assistance and checked by hand against the source at HEAD.

ArthurZucker:
Hey! We are moving towards v1, the deserializer is very much different see #2119 for more details

PR #2155, opened 2026-07-05
Reject a zero length in the FixedLength pre-tokenizer instead of panicking

v-code01:
## What

`FixedLength::length` is deserialized from the config with no validation (`FixedLength::new` performs none either). A value of `0` reaches:

```rust
for chunk in char_positions.chunks(self.length) { // panics: "chunk size must be non-zero"
```

so a crafted `{"type":"FixedLength","length":0}` pre-tokenizer crashes tokenization on any non-empty input.

## Fix

Reject `length == 0` with a clear error at the start of `pre_tokenize`:

```rust
if self.length == 0 {
    return Err("FixedLength pre-tokenizer length must be greater than 0".into());
}
```

## Tests

Adds `zero_length_is_rejected`; existing `FixedLength` tests are unchanged.

cc @ArthurZucker @McPatate

### DIFF
--- a/tokenizers/src/pre_tokenizers/fixed_length.rs
+++ b/tokenizers/src/pre_tokenizers/fixed_length.rs
@@ -23,6 +23,12 @@ fn default_length() -> usize {
 
 impl PreTokenizer for FixedLength {
     fn pre_tokenize(&self, pretokenized: &mut PreTokenizedString) -> Result<()> {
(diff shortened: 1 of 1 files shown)


v-code01:
cc @SBrandeis @McPatate — small, self-contained fix with a regression test: FixedLength length=0 panics in chunks() on any input. Would appreciate a look when you have a moment.

v-code01:

PR #2156, opened 2026-07-05
Do not panic when deserializing an invalid Precompiled normalizer

v-code01:
## What

Every arm of the tagged `NormalizerWrapper` deserializer routes parse failures through `.map_err(serde::de::Error::custom)?`, so an invalid `tokenizer.json` returns a clean serde error — **except** `Precompiled`:

```rust
EnumType::Precompiled => NormalizerWrapper::Precompiled(
    serde_json::from_str(
        &serde_json::to_string(&values).expect("Can reserialize precompiled"),
    )
    // .map_err(serde::de::Error::custom)
    .expect("Precompiled"),
),
```

A crafted `{"type":"Precompiled","precompiled_charsmap":"<invalid>"}` makes the inner `from_str` return `Err`, and `.expect("Precompiled")` turns that into a **panic while loading the tokenizer**, unlike every other normalizer. The commented-out `// .map_err(serde::de::Error::custom)` on the very next line shows the graceful path was intended.

## Fix

Route both fallible steps through `map_err(serde::de::Error::custom)?`, matching all the sibling arms:

```rust
EnumType::Precompiled => NormalizerWrapper::Precompiled(
    serde_json::from_str(
        &serde_json::to_string(&values).map_err(serde::de::Error::custom)?,
    )
    .map_err(serde::de::Error::custom)?,
),
```

## Tests

Adds `invalid_precompiled_charsmap_is_rejected`: a crafted `precompiled_charsmap` now returns `Err` instead of panicking. Full lib suite green (202); `cargo fmt`/`clippy` clean.

cc @ArthurZucker @McPatate

### DIFF
--- a/tokenizers/src/normalizers/mod.rs
+++ b/tokenizers/src/normalizers/mod.rs
@@ -137,10 +137,9 @@ impl<'de> Deserialize<'de> for NormalizerWrapper {
                     ),
                     EnumType::Precompiled => NormalizerWrapper::Precompiled(
                         serde_json::from_str(
-                            &serde_json::to_string(&values).expect("Can reserialize precompiled"),
+                            &serde_json::to_string(&values).map_err(serde::de::Error::custom)?,
                         )
-                        // .map_err(serde::de::Error::custom)
-                        .expect("Precompiled"),
+                        .map_err(serde::de::Error::custom)?,
                     ),
                     EnumType::Replace => NormalizerWrapper::Replace(
                         serde_json::from_value(values).map_err(serde::de::Error::custom)?,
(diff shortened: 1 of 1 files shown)


v-code01:
cc @ArthurZucker @McPatate — the `Precompiled` normalizer is the only arm of the tagged deserializer using `.expect()` instead of `.map_err(serde::de::Error::custom)?`, so a crafted `precompiled_charsmap` panics on load instead of returning a serde error. The commented-out `// .map_err(...)` on the same line shows the intended behavior.

v-code01:
cc @SBrandeis @McPatate — small, self-contained fix with a regression test: Precompiled is the only normalizer arm using .expect() instead of map_err, so a bad charsmap panics on load. Would appreciate a look when you have a moment.

v-code01:

PR #2165, opened 2026-07-07
feat: `impl pipeline::PreTokenizer for Sequence`

McPatate:
Will need to think about changing the inner `pretokenizers: Vec<PreTokenizerWrapper>` later when we cleanup the legacy code


SBrandeis (approved):
lgtm

PR #2166, opened 2026-07-08
Improve single thread performance of de-tokenization

sebpop:
Brings de-tokenization single thread performance up on arm64 and x86_64.

2 patches:
- byte-level: O(1) array for char->byte decode lookup
- decode: fused single-buffer fast path over the VocabStore byte slab

Perf (single-thread decode, one pinned core, baseline -> this series):
```
byte-level decode      llama3-en (English)   llama3-ja (CJK)
Nvidia Vera            40.8 -> 64.5 MiB/s    41.1 -> 181 MiB/s   (1.6x / 4.4x)
Intel Granite Rapids   29.7 -> 44.6 MiB/s    35.6 -> 125 MiB/s   (1.5x / 3.5x)
AMD Turin              46.1 -> 66.2 MiB/s    51.1 -> 180 MiB/s   (1.4x / 3.5x)
```

### DIFF
--- a/tokenizers/tk-encode/src/decoders/mod.rs
+++ b/tokenizers/tk-encode/src/decoders/mod.rs
@@ -164,6 +164,13 @@ impl Decoder for DecoderWrapper {
             Self::Fuse(bf) => bf.decode_chain(tokens),
         }
     }
+
+    fn decode_fused_bytes(&self, tokens: &[&[u8]]) -> Option<Result<String>> {
+        match self {
+            Self::ByteLevel(bl) => bl.decode_fused_bytes(tokens),
+            _ => None,
+        }
+    }
 }
 
 impl_enum_from!(BPEDecoder, DecoderWrapper, BPE);
--- a/tokenizers/tk-encode/src/pre_tokenizers/byte_level.rs
+++ b/tokenizers/tk-encode/src/pre_tokenizers/byte_level.rs
@@ -112,8 +112,16 @@ impl Decoder for ByteLevel {
             .flat_map(|t| {
                 t.chars()
                     .try_fold(vec![], |mut acc, c| {
-                        CHAR_BYTES_LOOKUP.get(&c).map(|b| {
-                            acc.push(*b);
+                        {
+                            let idx = c as usize;
+                            if idx < 512 {
+                                CHAR_BYTES_LOOKUP[idx]
+                            } else {
+                                None
+                            }
+                        }
+                        .map(|b| {
+                            acc.push(b);
                             acc
                         })
                     })
--- a/tokenizers/tk-encode/src/pre_tokenizers/byte_level.rs
+++ b/tokenizers/tk-encode/src/pre_tokenizers/byte_level.rs
@@ -130,6 +130,38 @@ impl Decoder for ByteLevel {
             .collect::<Vec<u8>>();
         Ok(vec![String::from_utf8_lossy(&toks).to_string()])
     }
+
+    fn decode_fused_bytes(&self, tokens: &[&[u8]]) -> Option<Result<String>> {
+        let total: usize = tokens.iter().map(|t| t.len()).sum();
+        let mut toks = Vec::with_capacity(total);
+        for t in tokens {
+            // Each `t` is a token's UTF-8 bytes. Map every char to its byte via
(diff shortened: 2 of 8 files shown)


tokenizers/tk-encode/src/pre_tokenizers/byte_level.rs
```diff
@@ -128,6 +136,43 @@ impl Decoder for ByteLevel {
...
         Ok(vec![String::from_utf8_lossy(&toks).to_string()])
     }
+
+    fn supports_fused_decode(&self) -> bool {
+        true
+    }
+
+    fn decode_fused_bytes(&self, tokens: &[&[u8]]) -> Option<Result<String>> {
+        let total: usize = tokens.iter().map(|t| t.len()).sum();
+        let mut toks = Vec::with_capacity(total);
```
ArthurZucker, line 136: could you try with tokens.len() * 4 ? worst case but could be faster

ArthurZucker (commented):
can you also add it to the benches please (see @SBrandeis 's pr to have them out as well)

tokenizers/tk-encode/src/pre_tokenizers/byte_level.rs
sebpop: On Vera, single-thread, byte-level: sum(len) gives `64.4 / 181.5 MiB/s` (en/ja) vs *4 at `63.5 / 178.2 MiB/s` — *4 is `~1.4–1.8%` slower. It under-allocates (these vocabs average >4 decoded bytes/token), so the fill loop hits a buffer realloc+copy; sum(len) is the exact upper bound and never reallocates, and that pre-pass only reads slice-length metadata (O(1) each), which is far cheaper than the realloc it avoids.

ArthurZucker: sounds good ty

tokenizers/tk-encode/src/tokenizer/mod.rs
```diff
@@ -937,6 +962,44 @@ where
...
+    fn decode_fused(&self, ids: &[u32], skip_special_tokens: bool) -> Option<Result<String>> {
+        let _ = skip_special_tokens;
+        let decoder = self.decoder.as_ref()?;
+        if !decoder.supports_fused_decode() {
+            return None;
+        }
+
+        // Skip the per-token added-vocab probe entirely when every added-vocab
+        // ID is above the max input ID (common: encoded text IDs are below the
+        // added-vocab range). One min/max scan replaces N HashMap probes.
```
ArthurZucker: added vocab is not  hashmap anymore, I'd separate this in another PR to bench it alone if possible

sebpop: Rebased onto today's feat/train_encode_split, split the added-vocab id-range short-circuit into a separate follow-up PR so it can be benched on its own (this PR now does a plain per-token check), and added a decode pass to the fixture benchmark.

sebpop: split the added-vocab id-range short-circuit into a separate follow-up PR:
https://github.com/huggingface/tokenizers/pull/2178

tokenizers/tk-encode/src/decoders/mod.rs
```diff
@@ -164,6 +164,20 @@ impl Decoder for DecoderWrapper {
             Self::Fuse(bf) => bf.decode_chain(tokens),
         }
     }
+
+    fn supports_fused_decode(&self) -> bool {
+        match self {
+            Self::ByteLevel(bl) => bl.supports_fused_decode(),
```
SBrandeis: This value is always `true`, can you explain why do we need the indirection?

Instead of:

```suggestion
            Self::ByteLevel(bl) => true,
```

SBrandeis (approved):

tokenizers/tk-encode/src/decoders/mod.rs
sebpop: Good call — that indirection was redundant, so I dropped supports_fused_decode entirely: decode_fused now just tries decode_fused_bytes (non-byte-level decoders return None) and non-byte-view models already bail on id_to_token_bytes, with decode output byte-for-byte unchanged.

ArthurZucker:
Closing in favor of #2305

PR #2190, opened 2026-07-12
[POC] fast encode: FlatCache + MPHF RankStore + incremental merge

ArthurZucker:
**Draft / experiment — kept as a trace, not for merge.**

Fast pipeline-encode POC on top of `feat/fast-normalization`. Byte-exact (BPE + sequence unit tests) on gpt2 / cl100k / o200k / llama-3.

### What's in it
- **`flat_cache.rs`** — thread-local, hash-keyed, L2-sized pretoken→ids cache (open-addressing + bump arena, clear-on-full). The pipeline had no cache at all.
- **`rank_store.rs`** — MPHF (`ptr_hash`) pair→(rank, merged) map: one 16-byte cache-line read per lookup, ~2× less memory than the `AHashMap` (llama-3 4.5 vs ~10 MB).
- **`word.rs` / `model.rs`** — incremental rank-array linear merge (O(n) MPHF lookups instead of O(n²)) + reused heap scratch; **0 hot-path allocations**.
- **`sequence.rs`** — drop `None` children so a single-`Split` `Sequence` (Llama-3) routes to the zero-alloc atomsplit FSM instead of the allocating multi-child loop; reused thread-local scratch for genuine multi-child.
- **`examples/xbench_{pipe,iter,alloc}.rs`** — throughput / per-iteration warm-up / allocation-count harnesses.

### Re-bench (single thread, mean MB/s, 10 Wikipedia languages)
| model | ours | wordchipper | IREE | bpe-openai | tiktoken | tok-main |
|---|--:|--:|--:|--:|--:|--:|
| gpt2 | **75** | 35 | 30 | — | 13 | 5 |
| cl100k | **57** | 40 | 24 | 22 | 13 | 6 |
| o200k | **70** | 44 | 25 | 26 | 11 | 6 |
| llama3 | **106** | n/a | 22 | — | 12 | 7 |

Multi-thread: ~6× near-linear scaling to 8 threads (gpt2 285, llama-3 307 MB/s), best of the field.

Single-thread numbers are warm-cache and data-dependent (hit-rate varies by language); cold-path merge work keeps low-repeat/CJK fast without the cache (CJK ~2× faster uncached after the incremental merge). All hot paths verified 0 allocs.


tokenizers/tk-encode/src/models/bpe/model.rs
```diff
@@ -752,62 +898,375 @@ impl PipelineBPE {
...
+        loop {
+            // leftmost lowest-rank live pair
+            let mut best = u32::MAX;
+            let mut bi = usize::MAX;
+            for i in 0..n {
+                if alive[i] && ranks[i].0 < best {
+                    best = ranks[i].0;
+                    bi = i;
+                }
+            }
```
SBrandeis, line 995: Use .min() on iter

SBrandeis: One smart trick is to pack rank in high bits with position in low bits (assuming pretoken and the merge cardinality / rank are small enough): `rank << 20 | pos`

Getting the min also gives you the position in the array for free

Probably negligible vs having another `usize` anyways

SBrandeis:
I made similar changes independently in https://github.com/huggingface/tokenizers/pull/2194, if you want to have a look :)

(your PR seems to have more scope)

Would love to see some bench numbers to see if it speeds up anything

PR #2196, opened 2026-07-14
Upgrade hf-hub to 1.0

assafvayner:
Upgrades the `hf-hub` dependency from 0.4 to 1.0 and migrates `from_pretrained` to the new API: the `ureq` feature is replaced by `blocking`, and the old `ApiBuilder`/`Repo` flow is replaced with `HFClientSync` + the `download_file` builder. Behavior is unchanged (env token/endpoint fallback, bare identifiers, and revisions all work as before); verified with `cargo test --features http --test from_pretrained` against the live Hub. cc @mcpotato

### DIFF
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -87,8 +87,8 @@ itertools = "0.14"
 log = "0.4"
 derive_builder = "0.20"
 spm_precompiled = "0.1.3"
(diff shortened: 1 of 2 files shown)


ArthurZucker (commented):
Done: https://github.com/huggingface/tokenizers/blob/main/tokenizers/tk-encode/Cargo.toml#L52

PR #2205, opened 2026-07-15
perf(wordpiece): zero-copy/zero-alloc PipelineWordPiece::tokenize_pipeline (~2× model stage)

ArthurZucker:
## What

Makes `PipelineWordPiece::tokenize_pipeline` **zero-copy / zero-alloc** on the hot path. ~**2× the WordPiece model stage**, byte-exact.

## Why

Profiling the pipeline showed the WordPiece "model" stage at ~15 ns/byte for bert — but a double-array `common_prefix_search` is just a few array loads per byte and shouldn't cost that. A microbench pinned the cause: it was **per-word allocation + string churn, not the automaton**. The old loop, per token, rebuilt a `candidate` `String` by copying the *entire remaining word tail* (to prepend the `##` continuing-subword prefix), and accumulated into a temp `Vec` — **two heap allocs + a full-tail copy per word**. That per-word allocation alone accounted for +4.7 / +1.6 / +6.1 ns/B (eng / rus / cmn).

## Change (`models/wordpiece/mod.rs` only)

- `PipelineWordPiece` gains `max_key_len` (longest vocab key, computed in `TryFrom`) — no match can extend past it.
- `start == 0`: search `&sequence.as_bytes()[start..]` **directly** — no `String`, no copy; the trie walk self-terminates.
- `start > 0`: build only `prefix + tail[..min(rest, max_key_len)]` into a per-call **`[u8; 256]` stack buffer** (heap `Vec` fallback only if `prefix + max_key_len > 256`). Byte-slicing avoids the `&str` char-boundary constraint.
- Tokens pushed **straight to `output`**; a mid-word miss does `output.truncate(rollback)` + one UNK — same all-or-nothing semantics, no temp `Vec`.
- Max-chars cap: `bytes > cap && chars().count() > cap` — the full UTF-8 decode only runs when byte length already exceeds the cap.

## Results (model-stage ns/byte, single-thread)

| fixture | before | after | speedup |
|---|--:|--:|--:|
| eng_Latn | 14.99 | 7.36 | 2.04× |
| rus_Cyrl | 14.18 | 8.77 | 1.62× |
| cmn_Hani | 19.03 | 7.15 | 2.66× |

## Correctness

- `cargo test -p tk-encode --lib wordpiece`: 6 passed (incl. 2 new regression tests — continuation `##` path + mid-word UNK rollback; this function had no direct coverage before).
- Pipeline bench, bert, all 22 fixtures: `ids_match` and `ids_match_baseline` (vs released `tokenizers`) both true, 0 failures.

## Note

What remains after this (~6 ns/B for eng) **is** the automaton, and it's L1-overflow / L2-latency-bound: bert's double-array is ~374 KB (> 128 KB L1), so each of ~2 random `get_unit` loads/byte is an ~L2-hit (~12 cycles). Squeezing that further needs a more compact/cache-local lookup (MPHF/`VocabStore`), which is out of scope here.

### DIFF
--- a/tokenizers/tk-encode/src/models/wordpiece/mod.rs
+++ b/tokenizers/tk-encode/src/models/wordpiece/mod.rs
@@ -319,6 +319,10 @@ pub struct PipelineWordPiece {
     unk_token: Option<u32>,
     continuing_subword_prefix: String,
     max_input_chars_per_word: usize,
+    /// Longest vocab key in bytes. A `common_prefix_search` can never match past
+    /// this depth, so continuing-subword candidates only need the first
+    /// `max_key_len` bytes of the remaining tail — never the whole tail.
+    max_key_len: usize,
 }
 
 impl TryFrom<WordPiece> for PipelineWordPiece {
@@ -336,6 +340,7 @@ impl TryFrom<WordPiece> for PipelineWordPiece {
         // yada requires the keyset sorted by key bytes.
         let mut keyset: Vec<_> = vocab.into_iter().collect();
         keyset.sort_unstable_by(|(a, _), (b, _)| a.cmp(b));
+        let max_key_len = keyset.iter().map(|(k, _)| k.len()).max().unwrap_or(0);
         let vocab_trie = DoubleArray::new(DoubleArrayBuilder::build(&keyset)?)?;
 
         Ok(Self {
@@ -343,6 +348,7 @@ impl TryFrom<WordPiece> for PipelineWordPiece {
             max_input_chars_per_word,
             unk_token,
             vocab_trie,
+            max_key_len,
         })
     }
 }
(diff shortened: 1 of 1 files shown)


ArthurZucker:
Closing — superseded and not worth landing:

- The alloc win this targeted is already merged in #2175 (`WordPieceScratch` buffer reuse + checkpoint/rollback). On top of that, the remaining copy-elimination is a wash on English (~1 ns/B only on long-word scripts).
- I A/B'd faster lookup structures against yada (double-array): FST is smaller (229 KB vs 374 KB) but ~3× slower; a first-byte-bucket "byte-index" was both slower and bigger (503 KB). yada wins decisively (bert ns/byte: yada 6.8–9.0 vs FST 20–39 vs byte-index 29–31, all byte-exact). The structure isn't the lever.
- yada's ~7 ns/byte is the L2-latency floor (its 374 KB double-array overflows L1). The only real lever left is skipping the lookup for repeated words — a pipeline-level word→ids cache (WordPiece has never had one; #2175's attempt was left commented-out as faulty) — which will be a separate PR.

PR #2213, opened 2026-07-17
feat: multi threaded pipeline

McPatate:
Status: ai generated iteration, will trim down and review thoroughly to cut the BS and verify for correctness in more depth

### DIFF
--- a/bindings/python/src/pipeline.rs
+++ b/bindings/python/src/pipeline.rs
@@ -28,9 +28,10 @@ struct PyStrBatch {
     views: Vec<PyStrView>,
 }
 
-// SAFETY: the raw pointers reference CPython's cached UTF-8 buffers, which are
-// immutable and outlive the owning `Py<PyString>`s in `_owners`; the batch
-// never mutates or frees them, and pool workers only read through `get`.
+// SAFETY: we only ever read from the underlying buffers,
+// and the lifetime of the buffers is guaranteed by storing
+// the Py<PyString> in _owners, which is kept alive as long as
+// a reference is held, because Python refcount.
 unsafe impl Send for PyStrBatch {}
 unsafe impl Sync for PyStrBatch {}
 
--- a/bindings/python/src/pipeline.rs
+++ b/bindings/python/src/pipeline.rs
@@ -1,8 +1,4 @@
-//! Experimental bindings for the encode-only `PipelineTokenizer` and its
-//! owned, escapable `EncodeHandle`: parallel encode on the library's persistent
-//! pool, with **zero-copy** Python-string inputs (the job keeps each
-//! `PyString` alive and reads its cached UTF-8 buffer directly — no copy into
-//! Rust `String`s).
+//! Pipeline python bindings
 
 use std::sync::Mutex;
 
@@ -10,7 +6,7 @@ use pyo3::exceptions;
 use pyo3::prelude::*;
 use pyo3::types::PyString;
 
-use tk::pipeline::{EncodeHandle, IntoInputs, PipelineToken, PipelineTokenizer, Inputs};
+use tk::pipeline::{EncodeHandle, Inputs, IntoInputs, PipelineToken, PipelineTokenizer};
 
 use crate::error::ToPyResult;
 use crate::tokenizer::PyTokenizer;
@@ -19,18 +15,20 @@ fn ids(tokens: Vec<PipelineToken>) -> Vec<u32> {
     tokens.iter().map(|t| t.id).collect()
 }
 
-/// Zero-copy batch of Python strings: keeps each `PyString` alive (a refcount
-/// held for the job's whole lifetime) and records its cached UTF-8 buffer,
-/// which CPython guarantees stable and immutable for the string's lifetime.
-struct PyStrBatch {
-    /// Keep-alives. Only pyo3's (GIL-deferred) drop ever touches them again.
-    _owners: Vec<Py<PyString>>,
-    /// (ptr, len) of each string's UTF-8 buffer, captured under the GIL by
-    /// `to_str` (which validates/caches the UTF-8 representation).
-    views: Vec<(*const u8, usize)>,
+struct PyStrView {
+    ptr: *const u8,
+    len: usize,
 }
 
-// Safety: the raw pointers reference CPython's cached UTF-8 buffers, which are
+/// Keep alive mechanism for zero-copy access to utf8 python strings
+struct PyStrBatch {
+    /// the keep alive ref
+    _owners: Vec<Py<PyString>>,
+    /// (ptr, len)
+    views: Vec<PyStrView>,
+}
+
+// SAFETY: the raw pointers reference CPython's cached UTF-8 buffers, which are
 // immutable and outlive the owning `Py<PyString>`s in `_owners`; the batch
 // never mutates or frees them, and pool workers only read through `get`.
 unsafe impl Send for PyStrBatch {}
@@ -42,7 +40,10 @@ impl PyStrBatch {
         let mut views = Vec::with_capacity(strings.len());
         for s in strings {
             let view = s.to_str()?;
-            views.push((view.as_ptr(), view.len()));
+            views.push(PyStrView {
+                ptr: view.as_ptr(),
+                len: view.len(),
+            });
             owners.push(s.unbind());
         }
         Ok(Self {
@@ -56,9 +57,10 @@ impl Inputs for PyStrBatch {
     fn len(&self) -> usize {
         self.views.len()
     }
+
     fn get(&self, i: usize) -> &str {
-        let (ptr, len) = self.views[i];
-        // Safety: see the struct invariant; `to_str` validated UTF-8 at capture.
+        let (ptr, len) = (self.views[i].ptr, self.views[i].len);
+        // SAFTEY: ptr is kept alive by PyStrBatch::_owners, and the underlying buffer is immutable UTF-8
         unsafe { std::str::from_utf8_unchecked(std::slice::from_raw_parts(ptr, len)) }
     }
 }
@@ -70,9 +72,6 @@ impl IntoInputs for PyStrBatch {
     }
 }
 
-/// Experimental encode-only pipeline (ids only — offsets and post-processing
-/// are not applied yet). Encoding runs multi-threaded on a persistent,
-/// fork-safe worker pool, with the GIL released.
 #[pyclass(module = "tokenizers", name = "PipelineTokenizer")]
 pub struct PyPipelineTokenizer {
     pipeline: PipelineTokenizer,
@@ -80,9 +79,6 @@ pub struct PyPipelineTokenizer {
 
 #[pymethods]
 impl PyPipelineTokenizer {
-    /// Build from an existing `Tokenizer`. The tokenizer is snapshotted through
-    /// its JSON form, so custom Python components are not supported (the
-    /// pipeline never runs Python callbacks on its hot path).
     #[staticmethod]
     fn from_tokenizer(tokenizer: &PyTokenizer) -> PyResult<Self> {
         let json = {
@@ -96,7 +92,6 @@ impl PyPipelineTokenizer {
         Ok(Self { pipeline })
     }
 
-    /// Build from a `tokenizer.json` file.
     #[staticmethod]
     fn from_file(path: &str) -> PyResult<Self> {
         let tok = PyResult::from(ToPyResult(tk::Tokenizer::from_file(path)))?;
@@ -119,13 +114,6 @@ impl PyPipelineTokenizer {
         Ok(PyEncodeHandle {
             job: Mutex::new(Some(job)),
         })
-    }
-
-    /// Blocking convenience: encode one `str` to its token ids.
-    fn encode(&self, py: Python<'_>, input: Bound<'_, PyString>) -> PyResult<Vec<u32>> {
-        let batch = PyStrBatch::new(vec![input])?;
-        let res = py.detach(|| self.pipeline.encode(batch).into_single());
-        Ok(ids(PyResult::from(ToPyResult(res))?))
     }
 }
 
@@ -150,8 +138,7 @@ impl PyEncodeHandle {
 
 #[pymethods]
 impl PyEncodeHandle {
-    /// Block until every input is encoded; returns one id-list per input, in
-    /// input order. Consumes the job: calling it a second time raises.
+    /// Block until every input is encoded, returning ids lists in input order
     fn wait(&self, py: Python<'_>) -> PyResult<Vec<Vec<u32>>> {
         let job = self.take()?;
         let res = py.detach(|| job.wait_for_completion());
@@ -163,8 +150,6 @@ impl PyEncodeHandle {
         slf
     }
 
-    /// The next input's ids in input order; blocks (GIL released, assisting
-    /// the pool) until that input is complete.
     fn __next__(&self, py: Python<'_>) -> PyResult<Option<Vec<u32>>> {
         let next = py.detach(|| {
             self.job
--- a/bindings/python/src/pipeline.rs
+++ b/bindings/python/src/pipeline.rs
@@ -108,31 +108,23 @@ impl PyPipelineTokenizer {
     /// pool workers encode in the background while you hold the job, `wait()`
     /// for everything, or iterate results as they complete (input order).
     /// Zero-copy: the job reads the Python strings' UTF-8 buffers in place.
-    #[pyo3(signature = (input, add_special_tokens = true))]
     fn encode_batch(
         &self,
         py: Python<'_>,
         input: Vec<Bound<'_, PyString>>,
-        add_special_tokens: bool,
     ) -> PyResult<PyEncodeHandle> {
         let batch = PyStrBatch::new(input)?;
         // Below the cost gate the job is computed inline — release the GIL.
-        let job = py.detach(|| self.pipeline.encode(batch, add_special_tokens));
+        let job = py.detach(|| self.pipeline.encode(batch));
         Ok(PyEncodeHandle {
             job: Mutex::new(Some(job)),
         })
     }
 
     /// Blocking convenience: encode one `str` to its token ids.
-    #[pyo3(signature = (input, add_special_tokens = true))]
-    fn encode(
-        &self,
-        py: Python<'_>,
-        input: Bound<'_, PyString>,
-        add_special_tokens: bool,
-    ) -> PyResult<Vec<u32>> {
+    fn encode(&self, py: Python<'_>, input: Bound<'_, PyString>) -> PyResult<Vec<u32>> {
         let batch = PyStrBatch::new(vec![input])?;
-        let res = py.detach(|| self.pipeline.encode(batch, add_special_tokens).into_single());
+        let res = py.detach(|| self.pipeline.encode(batch).into_single());
         Ok(ids(PyResult::from(ToPyResult(res))?))
     }
 }
--- a/bindings/python/src/pipeline.rs
+++ b/bindings/python/src/pipeline.rs
@@ -60,7 +60,7 @@ impl Inputs for PyStrBatch {
 
     fn get(&self, i: usize) -> &str {
         let (ptr, len) = (self.views[i].ptr, self.views[i].len);
-        // SAFTEY: ptr is kept alive by PyStrBatch::_owners, and the underlying buffer is immutable UTF-8
+        // SAFETY: ptr is kept alive by PyStrBatch::_owners, and the underlying buffer is immutable UTF-8
         unsafe { std::str::from_utf8_unchecked(std::slice::from_raw_parts(ptr, len)) }
     }
 }
--- a/bindings/python/src/pipeline.rs
+++ b/bindings/python/src/pipeline.rs
@@ -99,10 +99,11 @@ impl PyPipelineTokenizer {
         Ok(Self { pipeline })
     }
 
-    /// Encode a batch of `str`, returning an :class:`EncodeHandle` **immediately**:
+    /// Encode a batch of `str`, returning an [`PyEncodeHandle`] as soon as possible:
     /// pool workers encode in the background while you hold the job, `wait()`
-    /// for everything, or iterate results as they complete (input order).
-    /// Zero-copy: the job reads the Python strings' UTF-8 buffers in place.
+    /// for everything (returned in input order), or iterate `(index, ids)` as
+    /// each input completes. Zero-copy: the job reads the Python strings' UTF-8
+    /// buffers in place.
     fn encode_batch(
         &self,
         py: Python<'_>,
@@ -117,9 +118,10 @@ impl PyPipelineTokenizer {
     }
 }
 
-/// An in-flight (or completed) batch encode. Results surface in input order:
-/// `wait()` blocks for all of them; iterating yields each input's ids as it
-/// becomes ready. The consuming thread *assists* the pool while it waits (all
+/// An in-flight (or completed) batch encode. `wait()` blocks for all inputs and
+/// returns their ids lists in input order; iterating yields `(index, ids)` for
+/// each input as it completes (completion order, not input order — use `index`
+/// to place it). The consuming thread *assists* the pool while it waits (all
 /// with the GIL released). Dropping the job cancels unclaimed work.
 #[pyclass(module = "tokenizers", name = "EncodeHandle")]
 pub struct PyEncodeHandle {
@@ -150,7 +152,10 @@ impl PyEncodeHandle {
         slf
     }
 
-    fn __next__(&self, py: Python<'_>) -> PyResult<Option<Vec<u32>>> {
+    /// Yield `(input index, ids)` for each input as it finishes — completion
+    /// order, not input order. `index` is the position in the batch passed to
+    /// `encode_batch` (always `0` for a single input).
+    fn __next__(&self, py: Python<'_>) -> PyResult<Option<(usize, Vec<u32>)>> {
         let next = py.detach(|| {
             self.job
                 .lock()
@@ -160,7 +165,7 @@ impl PyEncodeHandle {
         });
         match next {
             None => Ok(None),
-            Some(res) => Ok(Some(ids(PyResult::from(ToPyResult(res))?))),
+            Some((seq, res)) => Ok(Some((seq, ids(PyResult::from(ToPyResult(res))?)))),
         }
     }
 }
--- a/bindings/python/src/pipeline.rs
+++ b/bindings/python/src/pipeline.rs
@@ -1,5 +1,5 @@
 //! Experimental bindings for the encode-only `PipelineTokenizer` and its
-//! owned, escapable `EncodeJob`: parallel encode on the library's persistent
+//! owned, escapable `EncodeHandle`: parallel encode on the library's persistent
 //! pool, with **zero-copy** Python-string inputs (the job keeps each
 //! `PyString` alive and reads its cached UTF-8 buffer directly — no copy into
 //! Rust `String`s).
@@ -10,7 +10,7 @@ use pyo3::exceptions;
 use pyo3::prelude::*;
 use pyo3::types::PyString;
 
-use tk::pipeline::{EncodeJob, IntoStableInputs, PipelineToken, PipelineTokenizer, StableInputs};
+use tk::pipeline::{EncodeHandle, IntoInputs, PipelineToken, PipelineTokenizer, Inputs};
 
 use crate::error::ToPyResult;
 use crate::tokenizer::PyTokenizer;
@@ -52,7 +52,7 @@ impl PyStrBatch {
     }
 }
 
-impl StableInputs for PyStrBatch {
+impl Inputs for PyStrBatch {
     fn len(&self) -> usize {
         self.views.len()
     }
@@ -63,9 +63,9 @@ impl StableInputs for PyStrBatch {
     }
 }
 
-impl IntoStableInputs for PyStrBatch {
-    type Storage = PyStrBatch;
-    fn into_stable(self) -> PyStrBatch {
+impl IntoInputs for PyStrBatch {
+    type Inputs = PyStrBatch;
+    fn into_inputs(self) -> PyStrBatch {
         self
     }
 }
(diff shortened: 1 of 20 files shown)


bindings/python/src/pipeline.rs
```diff
@@ -0,0 +1,171 @@
...
+
+/// Keep alive mechanism for zero-copy access to utf8 python strings
+struct PyStrBatch {
+    /// the keep alive ref
+    _owners: Vec<Py<PyString>>,
+    /// (ptr, len)
+    views: Vec<PyStrView>,
+}
+
+// SAFETY: the raw pointers reference CPython's cached UTF-8 buffers, which are
```
ArthurZucker:
```suggestion
// SAFETY: the raw pointers reference CPython's cached UTF-8 encoded buffers, which are
```

bindings/python/src/pipeline.rs
```diff
@@ -0,0 +1,171 @@
...
+
+impl Inputs for PyStrBatch {
+    fn len(&self) -> usize {
+        self.views.len()
+    }
+
+    fn get(&self, i: usize) -> &str {
+        let (ptr, len) = (self.views[i].ptr, self.views[i].len);
+        // SAFETY: ptr is kept alive by PyStrBatch::_owners, and the underlying buffer is immutable UTF-8
+        unsafe { std::str::from_utf8_unchecked(std::slice::from_raw_parts(ptr, len)) }
```
ArthurZucker, line 65:
```suggestion
        // SAFETY: the pointer given by view.as_ptr() is non null even for empty str. The memory range is not freed as PyStrBatch::_owners protects it.
        let slice =unsafe { std::slice::from_raw_parts(ptr, len)}
        // SAFETY: the pointed string IS a utf8 string it does not require another validation (O(len)).
        unsafe { std::str::from_utf8_unchecked(slice) };
```
I think its important to explain why we would go through unsafe when safe can work just as well + not hide to unsafe calls but split them?

bindings/python/src/pipeline.rs
```diff
@@ -0,0 +1,171 @@
...
+use tk::pipeline::{EncodeHandle, Inputs, IntoInputs, PipelineToken, PipelineTokenizer};
+
+use crate::error::ToPyResult;
+use crate::tokenizer::PyTokenizer;
+
+fn ids(tokens: Vec<PipelineToken>) -> Vec<u32> {
+    tokens.iter().map(|t| t.id).collect()
+}
+
+struct PyStrView {
```
ArthurZucker, line 18:
```suggestion
// A pointer to a utf8-encoded PyString that comes/lives in python world.
struct PyStrView {
```

bindings/python/src/pipeline.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+impl IntoInputs for PyStrBatch {
+    type Inputs = PyStrBatch;
+    fn into_inputs(self) -> PyStrBatch {
+        self
+    }
+}
+
```
ArthurZucker, line 75: as this is gonna change with the updates to support pairs of inputs, makes sense to have just a file for input?

bindings/python/src/pipeline.rs
```diff
@@ -0,0 +1,171 @@
...
+            job: Mutex::new(Some(job)),
+        })
+    }
+}
+
+/// An in-flight (or completed) batch encode. `wait()` blocks for all inputs and
+/// returns their ids lists in input order; iterating yields `(index, ids)` for
+/// each input as it completes (completion order, not input order — use `index`
+/// to place it). The consuming thread *assists* the pool while it waits (all
+/// with the GIL released). Dropping the job cancels unclaimed work.
```
ArthurZucker, line 126: how does it assist? why does it matter?

ArthurZucker: canceling return claimed and finished work in a wait?

tokenizers/tk-encode/examples/fixture_bench.rs
```diff
@@ -181,44 +189,106 @@ fn bench_threads(
...
 
-fn time_pass(encode: &dyn Fn(&str) -> usize, chunks: &[String]) -> f64 {
+fn time_pass(encode: &dyn Fn(&str), chunks: &[String]) -> f64 {
     let start = Instant::now();
-    let mut n = 0usize;
     for chunk in chunks {
-        n += encode(chunk);
+        encode(chunk);
     }
-    black_box(n);
```
ArthurZucker: mmm why?

tokenizers/tk-encode/examples/fixture_bench.rs
```diff
@@ -181,44 +189,106 @@ fn bench_threads(
...
+/// Cumulative pipeline stage levels for the ablation ladder, in execution
+/// order: each level runs every stage up to and including itself; `Model` is a
+/// full encode, `Frame` is the special-token scan + iteration only.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+enum Stage {
+    Frame,
+    Normalize,
+    Split,
+    Model,
+}
```
ArthurZucker, line 209: i think this is also defined in another file 
https://github.com/huggingface/tokenizers/blob/perf/flat-cache/tokenizers/tk-encode/src/tokenizer/pipeline.rs#L380-L389

let's standardize and re-use? unless we think its going away but it is to me a close to zero cost abstraction no?

tokenizers/tk-encode/examples/fixture_bench.rs
```diff
@@ -181,44 +189,106 @@ fn bench_threads(
...
+
+/// Median wall-time (seconds) of one pass over `chunks`, running the pipeline's
+/// stages up to and including `stage`. The partial pipeline is **recomposed
+/// here in the bench** from the pipeline's public components
+/// (`get_added_vocabulary` / `get_normalizer` / `get_pre_tokenizer` /
+/// `get_model`), mirroring the walker in the library's encode kernel — the
+/// library itself carries no staging or timing scaffolding. Timing successive
+/// levels and subtracting gives each stage's marginal cost (the ablation
+/// ladder). A runtime branch per segment is fine in a bench; the subtraction
+/// cancels its cost.
```
ArthurZucker, line 219: this kinda looks like shitty AI intuition

tokenizers/tk-encode/src/normalizers/mod.rs
```diff
@@ -238,6 +238,31 @@ impl pipeline::Normalizer for NormalizerWrapper {
...
+            | NormalizerWrapper::NFKD(_)
+            | NormalizerWrapper::Lowercase(_)
+            | NormalizerWrapper::BertNormalizer(_)
+            | NormalizerWrapper::StripAccents(_) => true,
+            NormalizerWrapper::Sequence(seq) => {
+                seq.as_ref().iter().all(Self::preserves_stride_boundaries)
+            }
+            // Strip, Prepend, Replace, Precompiled, Nmt, ByteLevel: may cross
+            // chunk boundaries.
+            _ => false,
```
ArthurZucker, line 261: yep

ArthurZucker: we could add that this can allow us to run pretokenization first or not

ArthurZucker: and why

ArthurZucker: as this only lives for the purpose of the parallelisation

tokenizers/tk-encode/src/pre_tokenizers/split.rs
```diff
@@ -121,6 +127,16 @@ impl Split {
     /// Isolated)`, the form the native FSM fast path requires (the inverted match
     /// set is the gaps, and these patterns leave no gaps). Rewrite to it so
     /// cl100k/o200k route to `fsm_cl100k`/`fsm_o200k` instead of the SysRegex fallback.
+    /// Whether the pipeline drives this `Split` with a native `atomsplit` FSM
+    /// (a recognized GPT regex in its canonical `Isolated`, non-inverted
+    /// usage). Every recognized family (gpt2 / cl100k / o200k) tokenizes such
+    /// that no token contains a non-whitespace→space transition — the invariant
+    /// the parallel runtime relies on to cut raw text at those boundaries.
```
ArthurZucker, line 134: this is a bit of a weird comment. I am also unsure it really holds as the way we do canonicalization means we are sometimes changing the usage of Isolated / removed.

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -108,6 +141,35 @@ impl PreTokenizer for PipelinePreTokenizer {
     }
 }
 
+impl PipelinePreTokenizer {
+    /// The [`StrideBoundary`] this pre-tokenizer guarantees, if any: a cut
+    /// finder its splitting rules provably cannot tokenize across. This is the
+    /// extension point for raw-text parallel chunking — a pre-tokenizer that
+    /// can prove a boundary for its grammar adds an arm here and large inputs
+    /// split under it automatically; anything without one falls back to the
+    /// model-stage split (still correct, just a serial pre-tokenize).
```
ArthurZucker: I really don't understand the comment

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
 /// Experimental encode-only pipeline built from a [`Tokenizer`]. Runs the same
 /// stages (special-token split → normalize → pre-tokenize → model) over borrowed
 /// ranges to avoid the reference path's allocations.
+///
+/// This is a cheap-clone **handle** over an `Arc`'d frozen core (the v2
+/// `Arc<Compiled>` shape): cloning shares the vocabulary/model, and an owned
+/// [`EncodeHandle`] keeps the tokenizer alive past the `encode` call by holding a
```
ArthurZucker: this is very hard to read as well. Arc is ref count, so not a real clone. And we don''t see how this is related to encode handle

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+/// The frozen pipeline components, shared by every clone of the handle.
+struct PipelineInner {
     added_vocabulary: BucketAddedVocabulary,
     normalizer: Option<NormalizerWrapper>,
     pre_tokenizer: PipelinePreTokenizer,
     model: PipelineModel,
     _post_processor: Option<PostProcessorWrapper>,
+    /// Whether some `normalized`-form added token can't survive a stride cut —
+    /// either its content carries an internal [`StrideBoundary`] cut, or it is an
+    /// affix (`lstrip`/`rstrip`) token whose absorbed adjacent whitespace a cut
```
ArthurZucker, line 338: lstrip and rstrip should not really be a problem no? you split anyways on added tokens first? is this saying "you can't use the cut as added token would eat it away"? But then you can still "use" it no?

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+    fn assert_send_sync<T: Send + Sync>() {}
+    assert_send_sync::<PipelineTokenizer>();
+};
+
+/// Reusable per-encode scratch, kept alive across sequences processed by the
+/// same worker thread (`reset`, never realloc'd).
+#[derive(Default)]
+pub(crate) struct EncodeState {
+    pre_tokens: Vec<Span>,
+    /// Classify tag scratch for the stride boundary scans (grow-and-reuse).
```
ArthurZucker: its not just for the stride boundary scan tho no?

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+    /// persistence across sequences and encode calls is the point.
+    model_scratch: Option<PipelineModelScratch>,
+}
+
+impl EncodeState {
+    pub(crate) fn new() -> Self {
+        Self::default()
+    }
+
+    /// Clear the scratch for reuse on the next sequence, keeping capacity.
```
ArthurZucker, line 367: we don't clear the tags?

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+    pub(crate) fn reset(&mut self) {
+        self.pre_tokens.clear();
+    }
+}
+
+/// Total input bytes below which `encode` runs inline on the calling thread (no pool):
+/// under this, the pool's dispatch/wakeup cost outweighs the parallelism gain. The same
```
ArthurZucker, line 374: IMO it makes sense to have this benchmarked somewhere no?

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+/// threshold sets the stride size for raw-cut chunking and the span-group target for
+/// `Pretokenized`, so it is the single "unit of parallel work" knob. Bench-tunable —
+/// splitting the three roles into separate knobs is a tracked tuning item.
```
ArthurZucker, line 377: comment is using obscure english words 😐

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+/// This is the **fused executor**'s scheduling primitive: there is no central
+/// cut list and no serial pre-scan — each worker claims a stride index and
+/// derives its chunk locally. Strides are fixed
+/// `PARALLEL_MIN_TOTAL_BYTES`-sized windows; a chunk starts at the first safe
+/// boundary inside its own window (stride 0 starts at 0) and ends at the first
+/// safe boundary found in the following windows — the exact scan those
+/// windows' owners also perform, so adjacent workers agree on every cut with
+/// no communication. Total scanning is O(input): each window is scanned at
+/// most twice (once by its owner, once by an extending predecessor). Text with
+/// no boundaries at all degrades to stride 0 owning the whole input.
```
ArthurZucker, line 394: ascii drawing of boundary and etc can help. comment to update as its also hard to read imo! 
The key is the boundary agreement without any communication, this is great

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+const NON_WS_PREV: u16 = Atom::Letter.bit()
+    | Atom::NumWord.bit()
+    | Atom::NumOther.bit()
+    | Atom::Mark.bit()
+    | Atom::Connector.bit()
+    | Atom::Punct.bit()
+    | Atom::Apostrophe.bit()
+    | Atom::SymOther.bit()
+    | Atom::NumericOther.bit();
+
```
ArthurZucker, line 446: I would split this in a new file

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+/// [`StrideBoundary`] for whitespace-delimiting pre-tokenizers (`Whitespace`,
+/// `WhitespaceSplit`, `Bert`): any whitespace char is a safe cut — these
+/// splitters drop whitespace and never emit a token spanning it, so each side
+/// drops its own run. The cut joins the right chunk and stays on a char boundary
+/// (a ws lead is `WS`, its continuation `Cont`).
+fn boundary_at_whitespace(window_tags: &[u8]) -> Option<usize> {
+    (1..window_tags.len()).find(|&i| in_mask(window_tags[i], mask::WS))
+}
+
```
ArthurZucker, line 500: pretty some boundary finding can be SIMD but that's for another PR

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+fn boundary_fsm(window_tags: &[u8]) -> Option<usize> {
+    (1..window_tags.len()).find(|&i| safe_fsm_cut(prev_char_tag(window_tags, i), window_tags[i]))
+}
+
+/// Whether `token`'s own bytes contain a cut of `boundary` strictly inside it —
+/// i.e. a stride could split this token and each half would mis-frame it. Used
+/// by `normalized_added_token_blocks_stride` to disqualify striding for a `normalized`
+/// added token whose content carries an internal cut (e.g. a `_→0` number
+/// transition). The special split peels raw-matched tokens before striding, so
+/// only normalized-form tokens reach this check.
```
ArthurZucker, line 550: i don't get it

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+
+enum HandleInner {
+    /// Fully computed; yields `(input index, result)` in input order.
+    Ready(std::iter::Enumerate<std::vec::IntoIter<Result<Vec<PipelineToken>>>>),
+    /// Filled by pool workers writing shared slots; the calling thread assists
+    /// and surfaces each input in completion order the moment its chunks are in.
+    Streaming { core: Arc<JobCore>, state: StreamState },
+}
+
+/// Completion-order streaming cursor over a job's shared slots. No channel and no
```
ArthurZucker, line 707: this is also probably a new file no?

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+        self.clone()
+    }
+}
+impl IntoInputs for &[&str] {
+    type Inputs = Vec<String>;
+    fn into_inputs(self) -> Vec<String> {
+        self.iter().map(|s| (*s).to_owned()).collect()
+    }
+}
+/// One owned encode call's worth of work, shared with the pool workers via
```
ArthurZucker, line 833: this is also a new file!

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -368,128 +1267,518 @@ impl TryFrom<&Tokenizer> for PipelineTokenizer {
...
+    /// holds it.
     ///
-    /// todo: wire the post-processing
-    pub fn encode(&self, input: &str, _add_special_tokens: bool) -> Result<Vec<PipelineToken>> {
-        let mut output = Vec::new();
-        let mut pre_tokens = Vec::new();
-        let mut scratch = self.model.init_scratch();
+    /// The job is `'static`: it holds the input storage and a cheap clone of this
+    /// tokenizer handle. Dropping the job cancels unclaimed
+    /// work, and the last worker releases the storage.
```
ArthurZucker, line 1320: this is key / core. Needs a proper doc explaining what is happening or this should be in a readme.

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -779,6 +2097,653 @@ mod tests {
...
+            "Prepend (unsafe normalizer) + WhitespaceSplit + WordPiece must pick `Normalized`"
+        );
+        let big = "aa bb cc\n".repeat(4000); // ~36 KB, forces striding
+        let ids = |v: Vec<PipelineToken>| v.iter().map(|t| t.id).collect::<Vec<_>>();
+        let serial = ids(encode_one(&tok, &big).unwrap());
+        let by_ref = ids(tok.encode(big.as_str()).into_single().unwrap());
+        assert_eq!(by_ref, serial, "Normalized (borrowed) != serial");
+        let owned = ids(tok.encode(big.clone()).into_single().unwrap());
+        assert_eq!(owned, serial, "Normalized (owned) != serial");
+    }
```
ArthurZucker, line 2814: we need to put the test somewhere else, this file is super hard to read

tokenizers/tk-encode/src/tokenizer/pool.rs
```diff
@@ -0,0 +1,152 @@
+//! Process-global worker pool for the parallel encode path: a lazily-built,
+//! library-private `rayon::ThreadPool`.
+//!
```
ArthurZucker, line 3: why? why do we need this etc!

tokenizers/tk-encode/src/utils/parallelism.rs
```diff
@@ -48,19 +74,43 @@ fn get_env_parallelism() -> bool {
...
+    crate::tokenizer::pool::invalidate();
+}
+
+/// The programmatic worker-count override, if one was set.
+pub(crate) fn num_threads_override() -> Option<usize> {
+    match NUM_THREADS.load(Ordering::SeqCst) {
+        0 => None,
+        n => Some(n),
+    }
+}
```
ArthurZucker, line 112: mega long due in this lib!

ArthurZucker (commented):
As a reviewer I am missing a lot of help on what is happening at a global scope. So I reviewed smaller scopes and some stuff here and there but we really need to do file split

bindings/python/src/pipeline.rs
```diff
@@ -0,0 +1,171 @@
...
+use crate::error::ToPyResult;
+use crate::tokenizer::PyTokenizer;
+
+fn ids(tokens: Vec<PipelineToken>) -> Vec<u32> {
+    tokens.iter().map(|t| t.id).collect()
+}
+
+struct PyStrView {
+    ptr: *const u8,
+    len: usize,
```
SBrandeis, line 20: Can't we use `PyBackedString` from pyo3 instead?

It seems to be doing exactly what you've re-implemented here, and provide safe APIs to get  `&str` from the Python-owned object

SBrandeis: https://pyo3.rs/main/doc/pyo3/pybacked/struct.pybackedstr

bindings/python/src/pipeline.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+#[pymethods]
+impl PyEncodeHandle {
+    /// Block until every input is encoded, returning ids lists in input order
+    fn wait(&self, py: Python<'_>) -> PyResult<Vec<Vec<u32>>> {
+        let job = self.take()?;
+        let res = py.detach(|| job.wait_for_completion());
+        let out = PyResult::from(ToPyResult(res))?;
+        Ok(out.into_iter().map(ids).collect())
```
SBrandeis, line 149: Ideally we should be able to send data to python without allocating (maybe through numpy array?)

Probably as a follow-up

tokenizers/tk-encode/examples/fixture_bench.rs
SBrandeis, line 209: Missing the post-process stage (reusing the other const would catch this)

SBrandeis: The issue with enum is that they cannot be used in `const` generics

tokenizers/tk-encode/examples/fixture_bench.rs
SBrandeis: Why scrapping encode_generic?

tokenizers/tk-encode/examples/fixture_bench.rs
```diff
@@ -181,44 +189,106 @@ fn bench_threads(
...
+                                                    .unwrap();
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
```
SBrandeis, line 282: Very brittle, needs an update every time we change the encode logic to be faithful

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -269,21 +321,849 @@ impl<'a, 'b, PatternMatcher: PipelinePatternMatcher> Iterator
...
+    off
+}
+
+thread_local! {
+    /// Per-thread reusable encode scratch: pool drainers, inline callers and the
+    /// caller-assist path all keep their buffers warm across encode calls
+    /// (reset, never realloc'd) — pool threads persist for the process, so this
+    /// is the cache-warmth contract of [`EncodeState`].
+    static SCRATCH: RefCell<EncodeState> = RefCell::new(EncodeState::new());
+}
```
SBrandeis, line 672: Is there a way to get rid of `thread_local!`?
For example by having each worker own its copy of EncodeState

See #2223 (scratch pool pattern)

bindings/python/src/pipeline.rs
McPatate: not encoded, this is only the input type!

bindings/python/src/pipeline.rs
McPatate: Safe would require a clone here, when it's not necessary because python guarantees the buffer won't be gc-ed as long as we have a reference (this is what `_owners` is for).
Python also guarantees immutability of the buffer, so we can safely create a `&str`  from it. Also the `.to_str` call we do in `PyStrBatch::new` guarantees it'll be utf8. So unsafe is needed here, because we're accessing the string's buffer via raw ptr mechanics (we have to if we want to avoid cloning the strings, which is not needed and thanks to this we're 0-copy from python input to rust, we won't be able to achieve something similar without unsafe), but we know the safety guarantees are upheld in this context.
```rs
            let view = s.to_str()?; // -> guarantees the string buffer's bytes are utf8, then cached and never mutated
            views.push(PyStrView {
                ptr: view.as_ptr(),
                len: view.len(),
            });
            owners.push(s.unbind());
```

bindings/python/src/pipeline.rs
McPatate: `PyString` is necessarily a python string from python world 😄 

Not sure a comment is needed here, imo quite explicit what this is?

bindings/python/src/pipeline.rs
McPatate: erm, not sure, I think we can extend this struct to support pairs of inputs rather than replacing it. Not sure what you mean by "file for input"

bindings/python/src/pipeline.rs
McPatate: > how does it assist? why does it matter?

```rs
impl Iterator for EncodeHandle {
    type Item = (usize, Result<Vec<PipelineToken>>);
    fn next(&mut self) -> Option<Self::Item> {
        match &mut self.inner {
            HandleInner::Ready(it) => it.next(),
            HandleInner::Streaming { core, state } => state.next_completed(core),
        }
    }
}

impl StreamState {
    fn next_completed(&mut self, core: &JobCore) -> Option<(usize, Result<Vec<PipelineToken>>)> {
        if self.next_k >= self.n {
            return None;
        }
        loop {
            let seq = core.completed_order[self.next_k].load(Ordering::Acquire);
            if seq != NOT_DONE {
                self.next_k += 1;
                return Some((seq, core.take_result(seq)));
            }
            // <------ HERE IS THE ASSIST LOGIC ------>
            if !self.assist_done {
                if SCRATCH.with(|st| core.run_one(&mut st.borrow_mut())) {
                    continue;
                }
                self.assist_done = true;
            }
            std::hint::spin_loop();
        }
    }
}
```

I've added a comment to show where the assist logic is.

Why:
- `sleep` is out of the question, too unpredictable and adds too many syscalls
- spin (`loop { if has_result() { break; } }`) burns a core for nothing, although it reacts fast
- condvar/futex (park) is the only viable alternative imo, reacts in the µs, but if we do implement it over assist, it has the following downsides:
  - latency: encode can start straight away, even before the pool has initialised
  - no context switching, the thread is always alive, no need to pay for µs wake
  - no need for the workers to signal the main thread that results are ready
  - overall more complicated, needs quite some code to implement, whereas assist is simply "call the work function we already have each worker do"

McPatate: > canceling return claimed and finished work in a wait?

not sure I understand the question, but if the `EncodeHandle` is dropped for wtv reason, the ongoing encode jobs are cancelled

tokenizers/tk-encode/examples/fixture_bench.rs
McPatate: link is broken? yes we can standardise, I just didn't want the generic to be part of the `encode` fn

tokenizers/tk-encode/examples/fixture_bench.rs
McPatate: perhaps, I haven't reviewed the examples files at all

tokenizers/tk-encode/src/normalizers/mod.rs
McPatate: > we could add that this can allow us to run pretokenization first or not
true, not sure it's useful, but yes.

not sure what you're asking me to do here tho, add a comment?

tokenizers/tk-encode/src/pre_tokenizers/split.rs
McPatate: this is the part I wanted review on the most, I'm not 100% sure the logic for splitting is the most correct

the structure to allow us to do that is there, which should only be simple updates if something is incorrect here though

tokenizers/tk-encode/src/tokenizer/pipeline.rs
McPatate: agree it's convoluted lol, let me rephrase

SBrandeis (commented):
Partial review will continue later

tokenizers/tk-encode/src/tokenizer/pipeline.rs
McPatate: we call reset in some places and update the scratch manually in other places. reworking to harmonise everything, sorry!

tokenizers/tk-encode/src/tokenizer/pipeline.rs
McPatate: `thread_local!` avoids having to use `Arc<Mutex<T>>`, it is already owned per thread

McPatate: it's just not shareable, which is not a problem in our case

bindings/python/src/pipeline.rs
McPatate: yes agree, will tackle later

PR #2217, opened 2026-07-18
Add a ParityBpeTrainer example and list it in the trainers API docs

cimeister:
ParityBpeTrainer was added without a usage example or an API docs entry. Its docstring assumed local per-language files that users are unlikely to have, and did not document the `ratio` argument. 

This PR:
* adds examples/train_parity_bpe.py, which pulls per-language training text from wikimedia/wikipedia and a parallel dev set from openlanguagedata/flores_plus, then demonstrates both balancing modes
* lists ParityBpeTrainer in docs/source-doc-builder/api/trainers.mdx
* describes the `ratio` parameter in more detail: it is a relative target, the trainer selects the language with the lowest compression_rate / ratio, and compression is counted in the units the pre-tokenizer emits, which for ByteLevel means bytes. Equal ratios therefore do not give equal tokenization across scripts, since Devanagari takes about 2.5x the bytes of Latin script for the same content. The example computes its ratios from mean FLORES+ bytes per sentence for that reason.


cimeister:
@ArthurZucker This is just docs for PA BPE. Could you take a look when you get a chance? Right now, the PA BPE code essentially can't be used because it doesn't have any instructions. Would be nice to get this in soon!

ArthurZucker (approved):
ty and sorry for the delay

PR #2237, opened 2026-07-23
feat(pipeline+bindings): Support encoding pairs of sequences

SBrandeis:


tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -563,23 +620,50 @@ impl PipelineTokenizer {
...
+                        _ => {}
+                    };
+                }
+            },
+            EncodeInput::Pair((first, second))  => {
+                for piece in self.encoding_template.iter_pair() {
+                    match piece {
+                        TemplatePart::Specials { ids, type_id } if add_special_tokens => output.extend_from_slice(ids),
+                        TemplatePart::Sequence { sequence_idx: SequenceIdx::First, type_id } => self.encode_single::<STAGE>(first, add_special_tokens, pre_tokens, scratch, output)?,
+                        TemplatePart::Sequence { sequence_idx: SequenceIdx::Second, type_id } => self.encode_single::<STAGE>(second, add_special_tokens, pre_tokens, scratch, output)?,
```
ArthurZucker, line 644: actually given the way we do close to 0 allocations now, this might be fine! 
Packing the two sequences in a single str might not do it any good

ArthurZucker (commented):
As much as I hate it, it is indeed used quite a lot

PR #2274, opened 2026-08-03
feat: update `encode` signature

McPatate:
* adds `EncodeHandle` return type to pave the way for async
* takes `impl Into<Inputs>` as an input
    * pairs are not properly supported atm, we still need to handle post-processor logic
* some of the benchmarks call encode* in a loop because it used to be single sequence only
  * I didn't dig too much to pass in the full batch and have kept it single-sequence for the moment, we could change that eventually

### DIFF
--- a/tokenizers/tk-encode/benches/pipeline_benchmark.rs
+++ b/tokenizers/tk-encode/benches/pipeline_benchmark.rs
@@ -80,7 +80,13 @@ fn bench_pipeline(c: &mut Criterion) {
                     b.iter(|| {
                         let mut n = 0usize;
                         for chunk in &chunks {
-                            n += pipeline.encode(chunk, true).unwrap().len();
+                            n += pipeline
+                                .encode(chunk, true)
+                                .wait()
+                                .unwrap()
+                                .first()
+                                .unwrap()
+                                .len();
                         }
                         black_box(n)
                     })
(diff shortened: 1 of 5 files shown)


tokenizers/tk-encode/benches/pipeline_benchmark.rs
```diff
@@ -80,7 +80,13 @@ fn bench_pipeline(c: &mut Criterion) {
                     b.iter(|| {
                         let mut n = 0usize;
                         for chunk in &chunks {
-                            n += pipeline.encode(chunk, true).unwrap().len();
+                            n += pipeline
+                                .encode(chunk, true)
+                                .wait()
+                                .unwrap()
+                                .first()
```
ArthurZucker, line 87: we could have a `_encode` vs `encode` as you said or something that does this for us when no using any async?

ArthurZucker: which is what happens for sequential input unbatch AFAIK

ArthurZucker (approved):

tokenizers/tk-encode/benches/pipeline_benchmark.rs
McPatate: imo we should keep a consistent API for both, this is sync but allows us to build future/async on top

PR #2276, opened 2026-08-04
perf(cache): take a word key in one load instead of a memcpy

ArthurZucker:
Closes the `TODO` on `pack_word`.

`pack_word` built the key with `lanes[..len].copy_from_slice(word)`. `len` is only
known at run time, so that is a call into `memcpy` behind a length branch that
mispredicts on nearly every short word. Read the 16 bytes at `word.as_ptr()` in one
unaligned load and mask the surplus off; the load reads past the word, so it is only
taken when all 16 bytes fall inside the page the word already sits in, and the old
copy stays as the fallback for the rest.

The packed `u128` is byte-identical to before — same value, same length in the top
byte — so slot placement, tags and key comparison are unchanged. A new test asserts
both reads agree, including a word deliberately placed to straddle a page boundary.

### Measured

Key construction timed on its own (spans from a pre-tokenizer pass over each corpus,
then key + hash per word), aarch64, `lto = "fat"`. ns per word, so it is independent
of how the text was split:

| corpus | B/word | `copy_from_slice` | one load + mask | |
|---|---|---|---|---|
| english | 4.5 | 7.49 | **1.63** | 4.6× |
| french | 5.0 | 7.78 | **2.01** | 3.9× |
| arabic | 9.2 | 6.38 | **2.03** | 3.1× |
| russian | 10.3 | 6.80 | **2.68** | 2.5× |
| chinese | 13.7 | 4.23 | **2.05** | 2.1× |

For scale: on english the old key path cost 1.59 ns/B against 0.48 ns/B for the whole
pre-tokenizer split, i.e. building the key cost more than 3× cutting the text into
words. It is now 0.36 ns/B.

Two variants I measured and did **not** ship:

- **`crc32cd` instead of aHash on the packed `u128`** — a further 1.7× on english
  (0.36 → 0.21 ns/B), but nothing on russian, and it needs a `cfg` fallback plus
  re-seeding to keep the table randomised. Happy to add it as a second commit if you
  want it.
- **Overlapping in-bounds loads instead of the page-guarded over-read** (two 8-byte
  reads covering `[0, len)` exactly, no reading past the word). Byte-identical and
  needs no `unsafe`, but the 3-way length branch costs most of the win back:
  0.92 ns/B vs 0.36 on english.

### vs #2267

Orthogonal, and they compose. #2267 changes *where* keys are computed (keyed spans
batched ahead of the model); this changes *what one key costs*, and does not touch the
[...]

PR #2262, opened 2026-07-29
perf(bpe): WordCache

SBrandeis:
# TL;DR

Implement a cache to memoize pretoken -> encoded ids

The cache is a lookup table with linear probing, capped at 16 steps

When no empty slot is found during the linear probing, evict naively the original index

## PipelineTokenizer benchmark

**10 / 10 models supported** — PipelineTokenizer vs `tokenizers` v0.23.1 (latest release) · ~10 kB inputs · add_special_tokens on · single thread + 1/2/4/8/max-thread sweep

`c503509cc · 2026-08-05 11:31 UTC` · Intel(R) Xeon(R) Platinum 8375C CPU @ 2.90GHz · 16 cores

Per-model encode throughput vs latest release

**vs base branch** (`077130b74`) — per-model geomean ×speedup of this PR's PipelineTokenizer against the base branch's; **regressions in red**.

Per-model encode throughput vs base branch

Per-model memory footprint

Minimal encode binary size

### Decode

Round-trip: v0.23.1 `encode_fast` produces the id streams (same fixtures, `add_special_tokens=true`); both implementations decode those SAME ids with `skip_special_tokens=false`. MB/s counts decoded text bytes.

Per-model decode throughput vs latest release

Per-model decode memory footprint

<details><summary><b>bert-base-uncased</b> — normalizer-heavy WordPiece · ×5.14 vs v0.23.1 · ×1.05 vs base · decode pending</summary>

bert-base-uncased speedup

bert-base-uncased stage decomposition

bert-base-uncased thread scaling

bert-base-uncased decode speedup

bert-base-uncased decode thread scaling

**Memory** (RSS MB, load+encode): v0.23.1 11+0 (peak 12) · Pipeline 8+2 (peak 17)

| Fixture | Group | v0.23.1 MB/s | Pipeline MB/s | Speedup | Δ base | added-token | normalize | pre-tokenize | model | post | Ids |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|:--|
| amh_Ethi | lang | 7.8 | 26.8 | ×3.43 | ×0.98 | 3% (1.2) | 77% (27.4) | 13% (4.7) | 7% (2.4) | 0% (0.0) | match |
| arb_Arab | lang | 4.2 | 26.6 | ×6.27 | ×1.07 | 3% (1.2) | 76% (27.4) | 9% (3.4) | 12% (4.2) | 0% (0.0) | match |
[...]

PR #2277, opened 2026-08-04
perf(cache): answer a home-slot hit without the walk, and fold the key with CRC

ArthurZucker:
Three things the warm hit path pays for and does not need. Independent of #2276 (which fixes
`pack_word`'s copy) — both apply.

1. **The walk starts at step 0 of a 16-slot window**, even though nearly every hit is in the home
   slot. A slot whose key equals ours holds our word — for a short word the key *is* the word, so
   there is nothing left for the tag to rule out. That case becomes one masked index, one load and
   one 16-byte compare.
2. **The hit path copied the whole 32-byte entry** and then indexed the same slot again to borrow
   its ids.
3. **The key is a fixed 16 bytes**, which hardware CRC folds in two instructions where a general
   hasher runs its whole mixing schedule. CRC yields 32 bits; one multiply spreads them over 64, so
   the low bits still choose the slot and the high bits still make the tag. The fold stays seeded
   from `hasher`, so the table is as randomised as before. Non-aarch64 keeps aHash.

### Measured

Warm lookup over every pretoken of a 4 MB corpus, aarch64, 100% hit rate, ns per word:

| | key | probe | total |
|---|---|---|---|
| gpt2 · english, before | 1.75 | 3.09 | 4.80 |
| gpt2 · english, after | **1.30** | **2.47** | **3.73** |
| llama-3 · english, after | 1.39 | 2.11 | **3.50** (from 4.84) |

**End-to-end this is currently noise** (493 vs 506 MB/s), and I would not claim otherwise: at 4 MB
inputs the lookup already overlaps with the split and the emit, so an isolated lookup benchmark
overstates its own contribution. It is strictly less work per lookup either way, and it becomes
visible once the per-pretoken call boundary is gone — which is the actual bottleneck:

With the model stubbed to answer every pretoken with one token, the model stage still costs
**0.64 ns/B on english — exactly what gigatoken's entire encode costs**. The per-pretoken virtual
call, slice, `Result` and push are the 5×, not the cache. #2267's batched `tokenize_keyed_spans` is
the fix for that; this PR is the hit path underneath it.

Things I measured and did **not** propose, for the record: prefetching the next word's slot
(−19%, it recomputes the key and the table is L2-resident), batching keys + prefetching with the
hash already computed (1.0×), branchless flat emit (noise), and raising capacity 10k → 1M (flat to
slightly worse).


SBrandeis:
Stale, see #2262

PR #2282, opened 2026-08-04
fix: iterate then wait potential out of bounds panic

McPatate:
Implementing `IntoIterator` on `EncodeHandle` gives us the compile time guarantee that we consume the handle on iteration and cannot call `.wait` on it later on, which won't mess up the indexing of the result vec.

This means the API is either iteration or waiting for all results.


McPatate:
@SBrandeis reviewed IRL :)

SBrandeis (approved):

PR #2275, opened 2026-08-03
perf(merge bpe): improvement suggestions

SBrandeis:

PR #2283, opened 2026-08-04
perf(pipeline): don't stride an input that has no cut points

ArthurZucker:
`plan()` chose `Raw` whenever the config exposed a stride boundary, without asking whether the *input* contains one. Punctuation-terminated CJK contains none: Chinese prose has no space anywhere, and a newline after `。` is not a legal cut because the punct rule ` ?[…]+[\r\n]*` absorbs it into the punct token — `NEWLINE_PREV` excludes punctuation for exactly that reason. `data/corpora/chinese.txt` is 100% such newlines: 722 of them, every one preceded by `。`, and zero spaces.

Striding then did what `stride_range`'s own doc comment warns about — "text with no boundaries at all degrades to stride 0 owning the whole input". On a 4 MB doc: 513 strides tiled, **1 resolved, covering all 4199874 bytes**, after scanning the document twice for cuts that don't exist. 0.28× the plain serial encode, and flat at every thread count because there is only ever one work unit.

`plan()` now probes the longest input for a real cut (bounded 32 KB, once per encode) and falls through to `Pretokenized` when there is none. That prefix runs the normal pre-tokenizer, so it's `bitsplit` where bitsplit is wired (gpt2, cl100k) — which is what makes it cheap enough to prefer over a degenerate stride.

14 × 4 MB docs, MB/s ours/gigatoken, interleaved best-of-3, quiet box:

| gpt2    | ours | giga | ratio |
|---------|------|------|-------|
| english | 6447 | 9185 | 0.70× |
| code    | 4290 | 3158 | 1.36× |
| korean  | 3774 | 1075 | 3.51× |
| russian | 4422 | 1213 | 3.65× |
| chinese | 4928 | 1289 | 3.82× |
| geomean |      |      | 2.16× |

| llama-3 | ours | giga | ratio |
|---------|------|------|-------|
| english | 5529 | 8430 | 0.66× |
| code    | 3327 | 4209 | 0.79× |
| korean  | 4776 | 1529 | 3.12× |
| russian | 5093 | 1617 | 3.15× |
| chinese | 2844 | 1875 | 1.52× |
| geomean |      |      | 1.51× |

gpt2 chinese went 0.94× → 3.82×. Single doc, one thread: 260 → 732 MB/s. Token counts unchanged (23531424) — every plan was already byte-exact, so plan choice only trades throughput. Spaced text still probes as cuttable and takes `Raw` on the identical path; english measured 0.68× → 0.70×.

`cut_exists` passes `lo = 1`, not 0: `boundary_in_window` reads one byte of left context via `block_lo - 1`, so 0 underflows to `usize::MAX` and spins forever in the char-boundary walk-back — it hangs rather than panicking in release. Byte 0 is never a cut anyway.

---

Stacked on `feat/multi_threaded_pipeline` because the plan ladder it fixes lands with #2213. Retarget to `feat/train_encode_split` once #2213 merges.

Two notes on the port from `poc/target-encode`, where this was developed and measured:

- This branch is edition 2018, so `plan()` is written with nested `if let` instead of the let-chains the original used. Same ladder, same order.
- The measurements above were taken on the merged stack (#2241 + #2275 + #2213). The fix is to plan selection only and is independent of the merge engine, but the absolute MB/s on this branch alone will be lower — the *ratios* between plans are what this PR changes. `examples/ab_giga_mt.rs` is included so the numbers are reproducible: it mirrors gigatoken's own `hf_mt` bench (one 4 MB doc per thread, `encode_docs_ragged` vs our batch `encode`, same MiB/s convention).

292 lib tests + 35 integration tests pass, including a new `uncuttable_input_skips_striding` that pins all three cases: `。`+newline is not a cut, uncuttable input is not strided, and Han-letter+newline and ordinary spaced text still take `Raw`.


McPatate (approved):

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -521,6 +521,27 @@ fn boundary_fsm(window_tags: &[u8]) -> Option<usize> {
...
+fn cut_exists(text: &str, boundary: StrideBoundary) -> bool {
+    const PROBE: usize = 4 * PARALLEL_MIN_TOTAL_BYTES;
+    // `boundary_in_window` reads one byte of left context (`block_lo - 1`), so `lo` must be
+    // at least 1 -- passing 0 underflows and hangs the char-boundary walk-back. Byte 0 is
+    // never a cut anyway: stride 0 starts at 0 unconditionally.
+    if text.len() < 2 {
+        return false;
+    }
+    let hi = text.len().min(PROBE);
+    let hi = (1..=hi).rev().find(|&i| text.is_char_boundary(i)).unwrap_or(1);
```
ArthurZucker, line 541: this can probably be improved as well btw, find is not really great

PR #2293, opened 2026-08-05
TARGET MINIMAL RUST CRATE + .tok FORMAT

ArthurZucker:
Based on `perf/bpe-merge-review` — review that first, this stacks on it.

## Target: on-device

A tokenizer that ships inside a phone app, a browser bundle, or an edge runtime is judged on what it adds to the binary. Ours added 2 MB gzipped, and almost none of it was the encoder — it was the JSON parser, the wrapper enums that exist only to be serde's deserialization target, and ~200 KB of Unicode tables kept alive by enum arms nothing ever constructs.

`.tok` is the fix: a container with no parser. 16-byte header, 16-byte section descriptors, sections at 64-byte alignment, read with a bounds check and a pointer cast. No pretty-printing, no keys, no escaping. v0 (`tokenizer.json`) stays supported — it moves to `tk-convert`, which runs offline.

## Before / after

The shipped artifact, gzipped:

| | raw | **gzipped** |
|---|---|---|
| `tokenizers.darwin-arm64.node` today | 5,145,664 | **2,010,745** |
| `.tok`-only binding, v1, all four models | 588,600 | **300,542** |

**6.69x.**

Same crate, same flags (`opt-level=z`, `panic=abort`, `cgu=1`, `lto=fat`, stripped):

| probe | stripped | gzipped |
|---|---|---|
| `binsize_pipeline` — v0 JSON load path reachable | 1,399,744 | **662,086** |
| `binsize_tok` — v1, `.tok` only | 569,608 | **286,130** |

Where it went, step by step, each measured:

| step | gzipped |
|---|---|
| v0 baseline | 662,086 |
| stop naming the wrapper enums in `from_tok` | 542,128 |
| gate the pipeline enums (`config`) — 202 KB of `__const` | 451,377 |
| serde + serde_json optional | 317,443 |
| rayon optional | 303,177 |
| `.tok` v1 gains Unigram / WordPiece / WordLevel | 334,651 |
| Unicode tables optional (`normalizers`) + ptr_hash without rayon/serde | **286,130** |

`__const` is down from 221,096 to 62,896 — what remains is `atomsplit`'s classification tables, which *are* the pre-tokenizer.

`cargo tree -p tk-encode -e normal -i {serde,serde_json,rayon,unicode-normalization,spm_precompiled}` is empty. The last two of those were never ours: both came from `ptr_hash`, where they were mandatory. Upstream PR makes each optional, patched in until it lands — https://github.com/RagnarGrootKoerkamp/PtrHash/pull/32.

## Crates

| crate | role | deps |
|---|---|---|
| `tk-serialization` | container and schema | **none**; write behind `write` |
| `tk-encode` | v1: reads `.tok`, encodes | no serde, no JSON, no rayon |
| `tk-convert` | v0 -> v1, offline | the whole JSON stack |
| `bindings/node-tok` | a Node binding whose whole surface is the read path | — |

Features, all off by default: `config` (the v0 layer — `Tokenizer`, the wrapper enums, serde), `normalizers` (the table-backed normalizers).

## What `.tok` v1 carries

All four models — the format describes a tokenizer, not one family of them.

| model | sections |
|---|---|
| BPE | `MERGE_PAIRS` in rank order, so a merge's rank is its index |
| Unigram | `VOCAB_SCORES`, `model_param` = unk id |
| WordPiece | `model_param` = `max_input_chars_per_word` |
| WordLevel | vocab + unk |

Pre-tokenizers are named by FSM family, never by regex source, so loading a `.tok` needs no regex engine: GPT-2/cl100k/o200k/tekken/DeepSeek drive `atomsplit` natively and a literal pattern is searched for directly.

## Verification

```
cargo run --release -p tk-convert --example tok_check
```

Converts, reloads through the read-only path, compares every id against the v0 path. **50/50 byte-exact** across gpt2, roberta, llama-3, deepseek-v4 and gemma-3 x 10 scripts.

```
cargo run --release -p tk-encode --example tok_ids -- data/*.tok
```

Digests every id and runs in every build config: **40/40 pairs identical** between the full build and the v1-only one. 332 existing tests pass with and without `config`.

File sizes: gemma-3 33.4 -> 9.8 MB, llama-3 17.2 -> 4.9 MB, deepseek-v4 6.4 -> 3.9 MB, gpt2 1.36 -> 1.36 MB (its JSON is mostly merges, already near-minimal as id pairs).

## Not in here

- A `.tok` cannot yet *name* a table-backed normalizer, so albert / xlm-roberta / bert convert only once the `NORMALIZER` section grows a kind list. The models work; their normalizer chains are the blocker.
- `gemma-2b` fails on a pre-existing pipeline limit: `<0x09>` is absent from its `model.vocab`, so byte-fallback cannot resolve. Unrelated to the format.

## Bugs found along the way

Both would bite any build without a regex backend, `.tok` or not:

- `Split::new` compiled every regex through the system backend. DeepSeek's three are not individually FSM-recognised (the pipeline runs them as one native pass), so constructing them demanded an engine. Added `Split::native`.
- `is_deepseek` compares against patterns with **literal** CR/LF, the way the shipped config spells them; `atomsplit::regexes` escapes them. Not interchangeable — rebuilding from the wrong one silently fell off the native FSM onto the regex fallback. `DEEPSEEK_PATTERNS` is now the single source.
- `Display for SplitDelimiterBehavior` and `for PrependScheme` got their names by delegating to the serializer, so both broke without serde. Spelled out now, matching what serde emitted.
- An empty normalizer `Sequence` (deepseek ships one) was carried as a no-op call per segment.


ArthurZucker:
## vs ExecuTorch's C++ tokenizers

[`pytorch-labs/tokenizers`](https://github.com/pytorch-labs/tokenizers) @ `9b96c39` (v1.4.0) is the closest comparison: a C++ reimplementation aimed at exactly the same on-device target. Both sides built here, same machine (M4), same corpora, single thread.

Their stack for the HF-json path: nlohmann/json, RE2 (+ 97 abseil archives), and PCRE2 as a lookahead fallback. cl100k's `\s+(?!\S)` is not RE2-compilable, so **llama-3 and every cl100k/o200k tokenizer requires the PCRE2 fallback** — the RE2-only build fails to load them.

### Size

Smallest program that loads a tokenizer and encodes. Stripped, LTO, `-Os` / `opt-level=z`.

| | can load llama-3 | stripped | **gzipped** |
|---|---|---|---|
| ExecuTorch, RE2 only | no | 903,624 | 401,076 |
| **ExecuTorch, + PCRE2 lookahead** | yes | 1,219,496 | **520,881** |
| tokenizers v0 (`tokenizer.json`) | yes | 1,399,744 | 662,086 |
| **tokenizers v1 (`.tok`)** | yes | 569,608 | **286,130** |

Against the ExecuTorch build that can actually serve llama-3: **1.82x smaller**. Against their RE2-only build: 1.40x.

### Throughput

Best of 5, ids only, one thread, `-O3` / `release`. **Every id matches ours exactly on every cell** — a good cross-check that both are correct.

**gpt2** — their RE2 path:

| corpus | ExecuTorch | tokenizers v1 | |
|---|---|---|---|
| english | 5.5 MB/s | 92.1 MB/s | **17x** |
| chinese | 8.2 MB/s | 166.0 MB/s | **20x** |
| code | 3.0 MB/s | 123.7 MB/s | **41x** |

**llama-3** — their PCRE2 lookahead path:

| corpus | ExecuTorch | tokenizers v1 | |
|---|---|---|---|
| english | 0.2 MB/s | 246.4 MB/s | ~1200x |
| chinese | 0.7 MB/s | 129.1 MB/s | ~180x |
| code | 0.1 MB/s | 188.5 MB/s | ~1900x |

The llama-3 row wants a caveat: that is their lookahead fallback being pathological, not a fair reading of their BPE engine. The honest summary is **17-41x on the path they are actually optimised for**, and that the most popular tokenizer family on the Hub falls off that path entirely because RE2 has no lookahead.

This is where `atomsplit` earns the 62 KB of tables it does keep: cl100k is a hand-written FSM, so there is no regex engine to be unable to compile it, no fallback, and no cliff.

### Load

| | ExecuTorch (JSON) | tokenizers v1 (`.tok`) |
|---|---|---|
| gpt2 | 38.0 ms | 44.0 ms |
| llama-3 | 265.8 ms | 142.0 ms |

Not a headline: we rebuild the derived tables at load on purpose, rather than freeze `tk-encode`'s internal layout into a file format. gpt2 is a wash; llama-3 is 1.9x.

### Reproducing

```sh
# ExecuTorch
git clone --recurse-submodules https://github.com/pytorch-labs/tokenizers
cmake -S tokenizers -B build -DCMAKE_BUILD_TYPE=Release -DSUPPORT_REGEX_LOOKAHEAD=ON
cmake --build build -j
c++ -std=c++20 -O3 -flto -DNDEBUG -I... et_bench.cpp -Wl,-force_load,build/libregex_lookahead.a $(find build -name '*.a')

# ours
cargo run --release -p tk-convert -- data/gpt2.json
cargo run --release -p tk-encode --example tok_bench -- data/gpt2.tok data/corpora/*.txt
```

`-Wl,-force_load` on `libregex_lookahead.a` matters: without it the linker drops the fallback's static registration and llama-3 fails to load with a message pointing at the missing link.

PR #2297, opened 2026-08-05
Perf/bpe merge review

SBrandeis:
# refactor(bpe): restructure the module

Pure reorganization of tk-encode's BPE module, split into one commit per concern so each is reviewable on its own. No behavior change: the final tree is byte-identical across the split, external paths stay stable through the mod.rs re-exports, and every commit compiles, passes the full tk-encode suite, and is cargo fmt clean.

Commit by commit:

1. chore(node): sync Cargo.lock with tk-encode's libc dependency (leftover from the fork-safe pool work).
2. Flatten the filename prefixes, group the released-crate BPE under legacy/. bpe_model.rs becomes model.rs, bpe_pretoken_to_rank.rs becomes convert.rs, bpe_build_tables.rs becomes tables.rs; the released-crate model, serialization and word move under legacy/. Almost pure git mv (96-99% similarity), plus import paths.
3. Split the tables file, document the packed merge value once. The perfect-hash pair lookup goes to pair_map.rs; SparseFold and the fold build go to fold/, next to the byte-level fold proof (fold/byte_level.rs, was bytelevel_folding.rs). The packed merge value layout is now documented once, in the tables module doc; merge_multipass and the pair-map slots refer to it instead of repeating the diagram.
4. Move code next to its users. The engine gate and BpeScratch into model.rs, Affixes and UTF8_LEN into convert.rs, the Vocab/Merges/MergeMap aliases and the At trait into mod.rs, WithFirstLastIterator into legacy/word.rs.
5. One name per concept. two_tier_queue_merge becomes merge_hot_cold_queue (with a module doc explaining the two tiers), MergeScratch becomes QueueScratch, byte_to_mode becomes byte_to_gate, and the convert docs say symbols instead of ranks throughout. Also fixes the scratch-reuse test in pipeline.rs, which still referenced the old merge field and did not compile.
6. Collapse the three copies of the char-as-bytes push into SymbolSink::push_char_bytes.

---


ArthurZucker (approved):
tty!

PR #2307, opened 2026-08-06
perf(bpe): wire the word cache into the pipeline model

ArthurZucker:
Four files, +83/−2.

## The gap

`word_cache.rs` has been on this branch since #2262 — which wired it into the pipeline BPE too. #2241 merged a day later as a squash of a branch that predated it, and its whole-file rewrite of `models/bpe/model.rs` (`@@ -1,694 +1,67 @@`) took the wiring and its tests with it. Nothing was deleted line by line; an older copy of the file simply landed on top. Unigram and WordPiece kept their cache, which is what makes it a clobber rather than a decision.

So today **only the legacy BPE uses it**. The pipeline model's scratch is `{ symbols, queue }` — no cache — so every pre-token is merged from scratch even when the same word appeared moments earlier. On text with any word repetition, that is most of the work.

`BpeScratch` gains `word_cache: Option<WordCache>`, sized from the model's own `cache` descriptor so a config setting capacity `0` still opts out. It lives in the scratch, not the model, because it has to outlive the encode call that fills it — otherwise it never sees a word twice.

One visibility change: `BPE::cache` from `pub(super)` to `pub(in crate::models::bpe)`, so the pipeline model reads the configured capacity rather than assuming the default.

## Numbers

Single thread, 10 kB documents, median of 3 interleaved rounds, MB/s:

| cell | before | after | |
|---|---:|---:|---:|
| gpt2 / hindi | 101 | 137 | **1.35×** |
| llama-3 / russian | 38 | 50 | **1.31×** |
| gpt2 / agentic-swe | 121 | 153 | **1.27×** |
| gpt2 / russian | 85 | 107 | 1.26× |
| llama-3 / xnli | 54 | 68 | 1.26× |
| gpt2 / xnli | 74 | 86 | 1.17× |
| llama-3 / agentic-swe | 110 | 128 | 1.17× |
| llama-3 / hindi | 150 | 174 | 1.16× |
| gpt2 / code | 121 | 139 | 1.15× |
| llama-3 / english | 117 | 122 | 1.04× |
| gpt2 / chinese | 84 | 84 | 1.01× |
| gpt2 / english | 143 | 141 | 0.98× |

**Median 1.16×.** Flat where words rarely repeat (chinese) and largest where they do. gpt2/english is within run-to-run noise on this box — sequential runs here are worth ±10% on some cells, so every figure above is interleaved.

## Why this matters beyond the numbers

[...]

PR #2308, opened 2026-08-06
perf(bpe): wire the word cache into the batched model path, fused

ArthurZucker:
Stacked on #2119 (`feat/train_encode_split`), and meant to go on top of #2306 + #2307.

## The bug this exists to prevent

#2306 gives the model a whole chunk of pre-tokens at a time by overriding `Model::tokenize_spans`. #2307 wires the word cache into `tokenize_pipeline`. **Those are different functions**, and the pipeline calls the batched one — so landing both as they stand leaves the cache unreachable on the path that actually runs. This wires it into both.

That is the same failure mode the `eabddcda` merge on #2279 already produced once: work that lands in a function nothing calls.

## Fused probe

On the batched path the probe is fused into the emit. `WordCache::probe_emit` checks the home slot and, on an inline hit, stores all `MAX_INLINE_IDS` lanes straight at the caller's cursor and returns only the count. `lookup` hands back a `&[u32]`, which makes the caller re-read the slot to build a fat pointer and then copy a run whose length it only learns at run time — three trips over one 32-byte line a single load already brought in. Anything that is not an inline home-slot hit falls back to the full window walk, reusing the placement so the word is never hashed twice. `lookup` and the walk are now `#[inline]`.

The output buffer is written through a cursor and closed with one `set_len`, so a hit costs no capacity check and no length store. It also reserves **2 ids per pre-token instead of 1**: `spans.len()` is a *lower* bound (92% of English pre-tokens are one id, 98% at most two), so reserving it made the buffer grow — and memcpy what it already held — partway through most chunks.

## Fold ordering

The fold stays in front of the cache. It answers a whole-vocabulary-entry word in one probe, cheaper than a cache probe, so folded words never enter the cache — they are already as cheap as a hit. This is also why the cache's marginal value looks small next to #2302: they are substitutes on the same population, not complements.

## `fold_by_flag` removed

Whether the config declared `ignore_merges` is a load-time question and does not belong in a per-pre-token branch. A declaring config asks every hit to fold, so every entry gets the bit; otherwise only entries that prove they reduce to themselves earn it. `fold_id` becomes one probe and one bit test with no policy in it, and `ignore_merges` stops being a field.

## Sparse-id fix

`prove_fold` bounded its walk by `vocab.len()` — the entry count — so any entry whose id exceeded it was left unproven. Ids may be sparse, so the bound is the id space, added as `BucketVocabStore::id_space`. No effect on gpt2 (dense ids), but the unified fold path above is wrong without it whenever a config leaves gaps; the `byte_level_ignore_merges_whole_word` test (vocab id 300, 257 entries) catches exactly that.

`PipelineToken` is asserted layout-identical to `u32`, since the fused probe writes ids through a pointer into the token buffer.

## Status

`cargo test -p tk-encode`: **367 passed, 0 failed.** Benchmarks posted below as a comment.

### DIFF
--- a/tokenizers/Cargo.lock
+++ b/tokenizers/Cargo.lock
@@ -166,6 +166,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "bitsplit"
+version = "0.1.0"
+dependencies = [
+ "ahash",
+ "atomsplit",
+]
+
+[[package]]
 name = "bitvec"
 version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2287,6 +2295,7 @@ dependencies = [
  "ahash",
  "assert_approx_eq",
  "atomsplit",
+ "bitsplit",
  "compact_str",
  "criterion 0.6.0",
  "daachorse 3.0.2",
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -1,6 +1,6 @@
 [workspace]
 resolver = "3"
-members = ["bitmap_gen", "atomsplit", "tk-encode", "tk-train"]
+members = ["bitmap_gen", "atomsplit", "bitsplit", "tk-encode", "tk-train"]
 
 [package]
 authors = [
--- a/tokenizers/bitsplit/Cargo.toml
+++ b/tokenizers/bitsplit/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "bitsplit"
+version = "0.1.0"
+edition = "2024"
+
+
+[dependencies]
+ahash = "0.8"
+atomsplit = { path = "../atomsplit" }
+
+
+
+
--- a/tokenizers/bitsplit/mask_splitter_spec.md
+++ b/tokenizers/bitsplit/mask_splitter_spec.md
@@ -1,0 +1,336 @@
+# bitsplit — mask splitter spec
+
+A Rust library for **pre-tokenizer splitting as a bitstream program**. Not a general regex engine:
+it covers exactly what `tokenizers` needs — the GPT-family regexes, the class-run family, and
+literal delimiters — and it does so with one shared operator set so a new grammar is ~150 lines of
+boolean algebra rather than a new hand-rolled FSM.
+
+Lineage: Parabix (Cameron et al., *Boosting the efficiency of text processing on commodity
+processors*) for the operator set and carry discipline; *Interleaved Bitstream Execution for
+Multi-Pattern Regex Matching on GPUs* (MICRO'25, doi 10.1145/3725843.3756052) for the execution
+model — fuse every instruction into ONE block-wise loop instead of one pass per instruction.
+
+Status: proven in `scratch/bitsplit` for three grammars, all byte-exact against their
+`atomsplit::fsm` counterparts.
+
+---
+
+## 1. Why it wins
+
+An FSM decides one token per unpredictable branch, so its cost tracks bytes-per-token. A bitstream
+program decides 64 bytes per register op, branchlessly, so its cost is flat.
+
+```
+                 ns/B (english, same run, aarch64)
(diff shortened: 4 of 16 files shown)


ArthurZucker:
## Benchmarks

tokbench, `--engine pipeline`, 10 kB chunks, 5 reps per cell, **5 interleaved A/B/C rounds** (2306 → 2279 → this PR, repeated) so within-session drift hits all three arms equally. Median MB/s, spread in parentheses.

| cell | 2306 | 2279 (closed poc) | **this PR** | vs 2306 | vs 2279 |
|---|---|---|---|---|---|
| gpt2 / code | 163 (158..164) | 282 (256..287) | **205** (197..206) | **1.26×** | 0.73× |
| gpt2 / russian | 94 (92..95) | 122 (117..123) | **121** (111..122) | **1.29×** | 0.99× |
| llama-3 / russian | 43 (43..44) | 58 (47..60) | **59** (57..60) | **1.37×** | 1.01× |
| gpt2 / english | 190 (183..193) | 172 (168..174) | 204 (179..204) | 1.08× ‡ | 1.19× |
| llama-3 / code | 173 (161..175) | 281 (239..288) | 187 (160..201) | 1.08× ‡ | 0.66× |
| llama-3 / english | 151 (148..153) | 168 (162..184) | 156 (149..159) | 1.03× ‡ | 0.93× |
| llama-3 / chinese | 50 (49..53) | 50 (47..50) | 49 (47..50) | 0.99× ‡ | 0.99× |
| gpt2 / chinese | 90 (89..93) | 86 (81..88) | 85 (81..87) | 0.94× | 0.99× |

‡ ranges overlap 2306 — not resolvable at this noise floor (±5–20% per arm). Only the top three rows and gpt2/chinese have non-overlapping ranges.

**Geometric mean over the eight cells: 105.0 → 117.6 MB/s, 1.12×.** Against 2279: 0.92×.

### What this recovers, and what it doesn't

**Russian is fully closed** — 1.29× / 1.37× over 2306, landing on 2279 (0.99× / 1.01×). **Code moves 1.26× on gpt2** but stops well short of 2279 (0.73× / 0.66×).

The residual on code is the fold and the cache both probing. On code a pretoken is usually *not* a whole vocabulary entry, so the fold probe misses and the cache probe follows it — two hashes of the same word where 2279, which has no fold at all for gpt2, pays one. This is the same shape #2302 already noted for a different gate: *"the miss path then computes the gate twice."* The fix is to share one hash between the two probes, which is a separate change and not attempted here.

**chinese is 0.94× on gpt2** and flat on llama-3: long CJK pretokens make the key expensive to build and repeat rarely, so the cache does not pay for itself there. A length gate on the cache probe is the obvious follow-up — 2279 caps at `MAX_WORD_BYTES = 1024`, and something far tighter is likely right.

### Exactness

`cargo test -p tk-encode`: **367 passed, 0 failed.**

Every cell above `verified=true` against the tokenizers 0.23.1 oracle, and the id hash is **identical across all three arms**:

| cell | ids_hash | tokens |
|---|---|---|
| gpt2/english | `d37f43f0013b8c3b` | 229925 |
| gpt2/code | `25bc83272a674597` | 347251 |
| gpt2/chinese | `4f4eab89e79b8f07` | 737407 |
| gpt2/russian | `c43780fa4421a02a` | 601558 |
| llama-3/english | `e213ebacf20e250b` | 224950 |
| llama-3/code | `150094c0efb654ba` | 243054 |
| llama-3/chinese | `38bb555aea70ddb8` | 314182 |
| llama-3/russian | `f0df162dd53c58a9` | 177217 |

### Caveat on absolute numbers

These are tokbench's protocol: 10 kB chunks, ~1 MB of distinct text per cell, and the ids copied out into the harness buffer. They are **not** comparable to the ~1000 MB/s figures in #2279's description, which come from `ab_giga` — one 4 MB buffer built by *repeating* a 200 kB corpus, which drives the cache to a hit rate distinct text never reaches, with no id copy. Both are valid; only same-protocol comparisons mean anything, which is why all three arms here were built and run identically.

PR #2313, opened 2026-08-07
perf(bpe): Last missing nits to get up to speed

ArthurZucker:
## What

On the BPE pipeline path, every pretoken that misses the fold was hashed **twice**:

1. `fold_id` → `BucketVocabStore::get_bytes_foldable` hashes the word for the MPHF probe;
2. `WordCache::lookup` → `make_lookup_key` hashes the same bytes again for the home slot and tag.

`BucketVocabStore::SEEDS` and `word_cache::PLACEMENT_HASHER` are seeded with the *same four
constants*, so the second pass was recomputing a `u64` the first had already produced.

This shares that one value: `hash_word` exposes it, `get_bytes_foldable_hashed` and `lookup_hashed`
accept it, and `tokenize_pipeline` computes it once per word.

## What this does not change

**Verification is untouched, so ids cannot change.** The vocabulary still compares the entry's bytes
to the query in full (perfect hashing only guarantees a valid slot for in-vocab keys), and the cache
still compares its 128-bit key. Only the duplicated hashing goes away.

A word longer than fifteen bytes still pays the cache's second, independently seeded
`DISCRIMINANT_HASHER` pass — that is what makes its key 127 bits rather than 64, and it is
deliberately left alone.

The public `lookup` / `get_bytes_foldable` entry points keep their old signatures and hash the word
themselves, so Unigram and WordPiece are unaffected.

## The invariant, and why there is a test for it

Sharing the hash is only sound while both sides seed `ahash` identically. If someone re-seeds either,
the failure is quiet and nasty: the cache would place a word under one hash and look it up under
another, producing **no wrong ids** but a permanent 0% hit rate — a large silent slowdown that no
correctness test would catch.

`vocab_and_word_cache_hash_a_word_identically` pins the two together across empty, short, boundary
(15/16 byte) and long words, so that change fails loudly here instead.

## Verification

- `cargo test -p tk-encode --all-features`: **357 passed**. In particular
  `the_proven_fold_never_changes_the_ids` passes, which compares pipeline ids against the legacy
  reference over eight strings covering folded words, merged words, `<|endoftext|>` (whose entry
  deliberately does not fold), CJK, code and a long repeated run.
- Two tests fail **both before and after this change**, on a clean checkout of the base, purely
  because their fixtures are absent from a fresh worktree (`Os { code: 2, NotFound }`):
  `models::bpe::model::fold_tests::the_proven_fold_never_changes_the_ids` (wants `../data/gpt2.json`
[...]

PR #2315, opened 2026-08-07
perf(bpe): pack a short word into its key instead of hashing it

ArthurZucker:
Stacked on #2313 — review that first; this PR's diff is only the second commit.

## What

A pretoken is short: english averages **4.83 bytes**, code 4.08, and the `<|...|>` shapes in
`added-special-dense` **2.29**. Running aHash over that is most of what the fold probe costs, and it
buys nothing — the vocabulary compares the entry's bytes regardless, so the hash only has to spread
well enough for the MPHF to separate keys.

For a word of **≤7 bytes**, pack the bytes and the length into a `u64` and mix them with one
multiply. Seven, so the length still fits in the top byte — that is what stops `"ab"` colliding with
`"ab\0"`. Longer words keep aHash, which mixes the length in itself.

## One definition, not two

`word_hash` is now the single place a word becomes a `u64`. `BucketVocabStore::build`, every probe,
and the word cache's placement all route through it, so a pretoken probed in both tables is hashed
once for the pair *and* the two cannot drift apart. The per-struct `RandomState` field goes away with
it: consistency used to come from carrying the hasher on the struct, and now it comes from there
being one function.

The `debug_assert` #2313 added earned its keep immediately — it caught `WordCache::lookup` still
calling a hasher of its own while `lookup_hashed` expected `word_hash`, which would have left the
cache placing short words under one value and looking them up under another (no wrong ids, but a
permanent 0% hit rate).

## Exactness is unchanged

Nothing verifies with `mix`. The vocabulary still compares the entry's bytes to the query in full;
the cache still compares its 128-bit key. So this cannot change any id, and it does **not** carry the
2^-32 digest weakening discussed on #2313.

Dropping the mixing entirely does not work, incidentally: packed short keys share their high bytes
and MPHF construction fails outright with `indistinguishable hashes in bucket`.

## Why this path was chosen

Measured on tokbench (gpt2, 29 cells, `--reps 1`, all byte-exact), splitting the suite by ASCII byte
fraction separates it cleanly:

| group | n | mean ratio vs gigatoken | mean excess per pretoken |
|---|---|---|---|
| >97% ASCII | 14 | 0.766x — **all 14 are losses** | **+3.01 ns** (median +2.93) |
| non-ASCII | 15 | 1.203x | −13.46 ns |

A near-constant **~3 ns ≈ 10 cycles per ASCII pretoken**, and the sign flips on non-ASCII where words
exceed the inline key and both sides hash for real. The one ASCII cell at parity is `dense` (+0.07),
which has 24.3 bytes per pretoken so a per-pretoken constant amortises away. aHash over a short word
is ~8–12 cycles against a masked load plus a multiply at ~2–3, which lands on that measurement.

Ruled out first, all measured: merges (`added-special-dense` does the *least* merge work of any cell —
1.17 `get_value`/pretoken vs english's 1.94 — and is the *worst* cell at 0.485x), the `GATE_ASCII`
crossover (sweep is monotonically worse below 20), the added-token matcher (0.05–0.11 ns/B at real
candidate densities, <10% of the excess anywhere), and per-call overhead (`ScratchPool`'s mutex plus
the output allocation is ~0.12 ns/pretoken, ~4% of the gap).

## Verification

`cargo test -p tk-encode --all-features`: **357 passed**. `the_proven_fold_never_changes_the_ids`
passes, which is the id-for-id gate against the legacy reference over folded words, merged words,
`<|endoftext|>`, CJK and code.

One unrelated failure, `normalizers::precompiled::tests::pipeline_precompiled_matches_legacy`, is
present on a clean checkout of the base too and is a missing fixture. `fold_tests` need
`tokenizers/data/gpt2.json`; `make data` fetches it.

No throughput number is claimed here — the per-cell noise floor on this machine is 8–13% on exactly
the ASCII cells this targets (identical code, three runs: english 8.5% spread, math-latex 12.8%),
so a ~3 ns/pretoken effect is below what it can resolve. The argument is the instruction count.

## Follow-up this sets up

The fold probe is 3 dependent loads (MPHF pilot → entry → byte-slab `memcmp`) where the cache's is 2,
purely because the cache keeps its key *in* the slot. Now that a short token's key **is** its bytes
plus length, storing that key in `Entry` makes verification an exact `u64` compare and drops the fold
probe to 2 loads with no loss of exactness — only tokens over 7 bytes would still need the slab.
Separate change, separate review, since it alters `Entry`'s layout.

## Note on the base

#2313 is based on the tip before #2304 merged, and `feat/train_encode_split` does not currently build
(see #2314). This stack wants a rebase once that lands; the `#[allow(dead_code)]` on `fold_id` is an
artefact of that stale base — the batched `tokenize_spans` is its caller.

### DIFF
--- a/tokenizers/tk-encode/src/models/bpe/model.rs
+++ b/tokenizers/tk-encode/src/models/bpe/model.rs
@@ -242,6 +242,9 @@ impl PipelineBPE {
 
     /// The id to emit for `sequence` without merging, when the whole pretoken is a vocabulary
     /// entry that may be folded. `None` sends the word to the merge engines.
+    // Dead on this branch only: the batched `tokenize_spans` is its caller and this stack is based
+    // on a tip that predates it. Rebasing onto a base that has the batched path uses it again.
+    #[allow(dead_code)]
     #[inline(always)]
     fn fold_id(&self, sequence: &str) -> Option<u32> {
         let bytes = sequence.as_bytes();
--- a/tokenizers/tk-encode/src/utils/word_cache.rs
+++ b/tokenizers/tk-encode/src/utils/word_cache.rs
@@ -56,17 +56,8 @@ use ahash::RandomState;
 use std::iter::Iterator;
 use wide::i8x16;
 
-/// Hashes a word to the 64 bits its home slot and tag are taken from, and to the
-/// bottom half of a long word's key ([`LookupKey::new_hash`]).
-static PLACEMENT_HASHER: RandomState = RandomState::with_seeds(
-    0x243f_6a88_85a3_08d3,
-    0x1319_8a2e_0370_7344,
-    0xa409_3822_299f_31d0,
-    0x082e_fa98_ec4e_6c89,
-);
-
 /// Hashes a long word a second time, to fill the half of its key that
-/// [`PLACEMENT_HASHER`] does not reach. The two hashes must be independent, or the
+/// [`placement_hash_of`] does not reach. The two hashes must be independent, or the
 /// key would carry 64 bits of information instead of 127.
 static DISCRIMINANT_HASHER: RandomState = RandomState::with_seeds(
     0x4528_21e6_38d0_1377,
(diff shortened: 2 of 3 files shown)


ArthurZucker:
Rebased onto `feat/train_encode_split` now that #2314 has landed (rebase is local until the branch is force-pushed).

Benched the whole stack: **geomean 1.1493x over 29 gpt2 cells, 29/29 faster, all ids identical**, interleaved A/B/A/B/A/B. Full numbers and method on #2316.

Note the stack has to land together to show that. `pipeline.rs` routes encode through `tokenize_spans`, and these PRs only wired `tokenize_pipeline` — so on its own this PR touches a path the encode loop does not use. The final commit of the stack wires the batched path.

PR #2316, opened 2026-08-07
perf(bpe): put the key in the entry, so the fold probe is two loads

ArthurZucker:
Third in the stack: #2313 → #2315 → this. Review the parents first; this PR's diff is the third commit.

## What

The fold probe was **three dependent loads**: the MPHF pilot, the entry, then the byte slab to
compare the stored token against the query. The word cache does the same job in **two**, and the
reason is layout rather than luck — its key lives in the slot it verifies, so nothing else has to be
read.

Give the vocabulary the same shape:

- `Entry` becomes `{ key: u64, id: u32 }` — everything a probe reads and nothing it does not.
- `(start, len)` moves to a parallel `spans` array, touched only by the reverse lookup and by
  enumeration. Keeping them in `Entry` made every probe drag six dead bytes through cache.

A probe is now **pilot + entry**.

## Exactness is unchanged — deliberately

This is the part worth reviewing closely, because the obvious version of this change is not exact.

- A word of **≤ `INLINE_KEY_BYTES` (7)** bytes has a key that *is* its bytes plus its length (#2315).
  So comparing keys **is** proof of identity, and the slab is never read.
- A **longer** word keys by aHash, which is *not* proof, so it still confirms against the slab — the
  load it was paying anyway.

So the saving lands exactly on the short pretokens that are the measured gap — english averages 4.83
bytes per pretoken, code 4.08, `added-special-dense` 2.29 — and **nothing gives up the never-wrong
guarantee**. This is not the `~2^-64` key-only compare, and not the `~2^-32` digest.

`confirm()` is where that split lives, and it is `#[inline(always)]` with the length test first, so
a short word compiles to no slab access at all.

## The zero-key sentinel

`LEN_TAG` now biases the length by one. A non-minimal MPHF returns padding slots whose
`Entry::default()` key is `0`, and the probe rejects those with the *same single compare* it uses for
everything else — which only works while no real word can key to `0`. The empty word keyed to
exactly that before the bias. `no_real_key_is_zero` pins it, including seven zero bytes, where the
length tag is the only thing keeping the key off `0`.

## Tests

- `out_of_vocabulary_words_are_rejected` — OOV words at both key regimes, including a prefix of a
  present token, a same-length miss against a short token, and a same-length-class miss against a
  long one. This is the test that would catch a key-only compare being wrong.
- `no_real_key_is_zero` — the sentinel invariant above.
- `the_proven_fold_never_changes_the_ids` still passes: ids compared against the legacy reference
  over folded words, merged words, `<|endoftext|>`, CJK and code.

`cargo test -p tk-encode --all-features`: **359 passed**. The one failure,
`normalizers::precompiled::tests::pipeline_precompiled_matches_legacy`, is present on a clean
checkout of the base and is a missing fixture, not a regression.

## Why this rather than an open-addressed tag row

The literal way to copy the cache's 2 loads is to drop the MPHF for open addressing with a tag row.
Measurement here already went the other way: on gpt2's pair table, MPHF (CHD) at **292 KiB / 2.30 ns**
beat open-addressed 8x at **4096 KiB / 2.79 ns**, because at 4–8 MiB the open table overflows L2
while the MPHF fits with its pilots in L1. The pilot load is small and hot; the third load was the
expensive one. Removing that keeps the MPHF's size advantage *and* reaches two loads.

## No throughput claim

Per-cell noise on this machine is 8–13% on exactly the ASCII cells this targets (identical code,
three runs: english 8.5% spread, math-latex 12.8%, code 9.2%), so the effect is below what it can
resolve. The argument is one fewer dependent load on the ~79%-of-pretokens path in
`added-special-dense` and the ~22% path in english.

## Base

The stack sits on the tip before #2304 merged, and `feat/train_encode_split` does not currently
build — see #2314. All three want a rebase once that lands.

### DIFF
--- a/tokenizers/tk-encode/src/models/bpe/model.rs
+++ b/tokenizers/tk-encode/src/models/bpe/model.rs
@@ -12,7 +12,7 @@ use crate::pipeline::{self, PipelineToken};
 use crate::tokenizer::Result;
 use crate::utils::byte_level::{self};
 use crate::utils::word_cache::{Lookup, WordCache};
-use crate::vocab::bucket_vocab_store::BucketVocabStore;
+use crate::vocab::bucket_vocab_store::{BucketVocabStore, key_and_hash};
 
 const GATE_MULTI: u16 = 8;
 const GATE_ASCII: u16 = 24;
@@ -248,16 +248,16 @@ impl PipelineBPE {
     #[inline(always)]
     fn fold_id(&self, sequence: &str) -> Option<u32> {
         let bytes = sequence.as_bytes();
-        self.fold_id_hashed(bytes, self.vocab.hash_word(bytes))
+        let (key, hash) = key_and_hash(bytes);
+        self.fold_id_keyed(bytes, key, hash)
     }
 
-    /// [`Self::fold_id`] for a caller that already hashed the word with
-    /// [`BucketVocabStore::hash_word`].
+    /// [`Self::fold_id`] for a caller that already ran [`key_and_hash`] on the word.
     #[inline(always)]
-    fn fold_id_hashed(&self, bytes: &[u8], hash: u64) -> Option<u32> {
+    fn fold_id_keyed(&self, bytes: &[u8], key: u64, hash: u64) -> Option<u32> {
         // One probe; the foldable bit is part of the id that probe already returned. Which entries
         // carry it was settled at load -- see `from_bpe`.
-        let (id, foldable) = self.vocab.get_bytes_foldable_hashed(bytes, hash)?;
+        let (id, foldable) = self.vocab.get_bytes_foldable_keyed(bytes, key, hash)?;
         foldable.then_some(id)
     }
 
@@ -326,8 +326,8 @@ impl pipeline::Model for PipelineBPE {
         // way -- the vocabulary compares the entry's bytes, the cache compares its key -- so this
         // shares the hash and nothing else.
         let bytes = sequence.as_bytes();
-        let hash = self.vocab.hash_word(bytes);
-        if let Some(id) = self.fold_id_hashed(bytes, hash) {
+        let (key, hash) = key_and_hash(bytes);
+        if let Some(id) = self.fold_id_keyed(bytes, key, hash) {
             output.push(PipelineToken { id });
             return Ok(());
         }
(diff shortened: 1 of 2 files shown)


ArthurZucker:
## Benched: geomean **1.1493x**, 29/29 cells faster, ids identical

Rebased the stack onto `feat/train_encode_split` now that #2314 has landed, and measured it.

**One thing had to be fixed first, and it invalidates any earlier reading of this stack:**
`pipeline.rs` routes encode through `tokenize_spans`, not `tokenize_pipeline`. All three PRs had only
wired `tokenize_pipeline`, so none of them were on the path that actually runs. A fourth commit wires
the batched path — `key_and_hash` once per word, then `fold_id_keyed` and `lookup_hashed`. `fold_id`
had exactly one caller and folded into `fold_id_keyed` with it.

### Method

tokbench, gpt2, 29 verified cells, `--reps 1`, **interleaved A/B/A/B/A/B** on one machine. Interleaved
because run-to-run drift here is at run granularity, not rep granularity: three runs of *identical*
code gave english an 8.5% spread and math-latex 12.8%, and more reps did not help. The build now
aborts on failure — the first attempt at this silently re-ran a stale binary and produced six
meaningless numbers.

### english (ns/byte, lower is better)

| round | base | stack | ratio |
|---|---|---|---|
| 1 | 5.936 | 5.203 | 1.141x |
| 2 | 6.216 | 5.026 | 1.237x |
| 3 | 6.010 | 5.108 | 1.177x |

The distributions are **disjoint** — base min 5.936 > stack max 5.203 — so this is not the noise floor.

### All 29 cells

Per-round geomean **1.1381x / 1.1404x / 1.1693x**, mean **1.1493x**. Every cell faster; every cell's
id hash identical to the base in every round.

Median per-cell ratio, extremes:

| biggest | | smallest | |
|---|---|---|---|
| tamil | 1.326x | japanese | 1.021x |
| agentic-swe | 1.322x | xnli | 1.024x |
| added-special-dense | 1.262x | korean | 1.032x |
| chat-llama3 | 1.248x | georgian | 1.048x |
| english | 1.177x | greek | 1.069x |

**The shape is the confirmation, not just the size.** The gain scales with pretokens-per-byte and
shrinks on long-word CJK — exactly what a per-*pretoken* saving predicts (one hash and one dependent
load per word). `added-special-dense`, where ~79% of pretokens pay a failed fold probe followed by a
cache hit over only 10 distinct words, gains 1.262x.

### Caveat on the absolute numbers

These were taken with tokbench's pipeline adapter locally patched to call `encode_generic` instead of
`encode_generic_into`, because the latter does not exist upstream — it is an API only on the
`perf/2306-cache-fused` WIP branch. The extra allocation is ~one 10 kB `malloc` per 10 kB chunk
(~0.01 ns/B), so it does not explain the absolute level; both sides pay it identically and the ratio
is unaffected. The gap between these ~6 ns/B and the ~3 ns/B the WIP reaches is the rest of that
branch — the multipass ranks/prods cache, the `live` bitmap, the fused `probe_emit` — none of which is
upstream or in this stack.

PR #2331, opened 2026-08-10
feat: wire in the thread pool

McPatate:
Wiring in the thread pool.

For now we only parallelise on special tokens, the intra-seq splitting will come in a subsequent PR.

`Pair`s and post processing have been handled in this PR too

Left to do in this PR:
- [x] add some testing

### DIFF
--- a/tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
+++ b/tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
@@ -512,34 +512,12 @@ impl TryFrom<&PostProcessorWrapper> for PipelinePostProcessor {
     }
 }
 
-fn is_sequence(s: &Slice) -> bool {
-    matches!(s, Slice::Sequence { .. })
-}
-
-/// Split into three groups, depending on the template's content:
-/// - [<before>], [SeqA<middle>SeqB], [<after>]
-/// - [<before>], [SeqA], [<after>]
-fn into_parts(template: &Template) -> Result<(Vec<Slice>, Vec<Slice>, Vec<Slice>)> {
-    let s = &template.slices;
-    let (Some(first), Some(last)) = (
-        s.iter().position(is_sequence),
-        s.iter().rposition(is_sequence),
-    ) else {
-        return Err("not supported: could not find any sequence in post processor template".into());
-    };
-    Ok((
-        s[..first].to_vec(),
-        s[first..=last].to_vec(),
-        s[last + 1..].to_vec(),
-    ))
-}
-
-/// An identity core arranges the sequences exactly as the defaults: `$A` alone, or `$A $B` with
-/// the default type ids (0 then 1). Only such a core is transparent and safe to drop when composing
-/// a Sequence. Any other all-sequence core reorders or retags, so it is a real arrangement.
-fn is_identity_core(core: &[Slice]) -> bool {
+/// A pass-through template does nothing: no special tokens, and sequences in the default
+/// arrangement (`$A`, or `$A $B` with the default type ids 0 then 1). Such a member is a no-op in a
+/// Sequence and is dropped when composing. Anything else adds tokens or reorders/retags.
+fn is_pass_through(slices: &[Slice]) -> bool {
     matches!(
-        core,
+        slices,
         [Slice::Sequence {
             seq: Seq::A,
             type_id: 0
@@ -556,31 +534,29 @@ fn is_identity_core(core: &[Slice]) -> bool {
     )
 }
 
+/// Compose the members of a Sequence post processor.
+///
+/// The reference applies each member in turn, and every member retags the whole previous output to
+/// its own sequence type id. A static template cannot represent that chaining, so we only support
+/// the representable case: at most one member that adds tokens or reorders/retags, wrapped by any
+/// number of pass-through members (which are no-ops and dropped). More than one is rejected.
 fn compose<'a>(templates: impl Iterator<Item = &'a Template>) -> Result<Template> {
-    let parts = templates.map(into_parts).collect::<Result<Vec<_>>>()?;
-
-    let mut core = None;
-    for (_, c, _) in &parts {
-        if !is_identity_core(c) && core.replace(c.clone()).is_some() {
+    let templates = templates.collect::<Vec<_>>();
+    let mut chosen: Option<&Template> = None;
+    for template in &templates {
+        if is_pass_through(&template.slices) {
+            continue;
+        }
+        if chosen.replace(template).is_some() {
             return Err(
                 "post processor Sequence with multiple sequence referencing members is not supported".into(),
             );
         }
     }
-
-    let core = core
-        .or_else(|| parts.first().map(|(_, c, _)| c.clone()))
+    let chosen = chosen
+        .or_else(|| templates.first().copied())
         .ok_or("empty Sequence post processor is not supported")?;
-
-    let mut slices = Vec::new();
-    for (prefix, _, _) in parts.iter().rev() {
-        slices.extend(prefix.iter().cloned());
-    }
-    slices.extend(core);
-    for (_, _, suffix) in parts {
-        slices.extend(suffix.iter().cloned());
-    }
-    Ok(Template::new(slices))
+    Ok(Template::new(chosen.slices.to_vec()))
 }
 
 impl Default for PipelinePostProcessor {
@@ -675,7 +651,7 @@ pub enum Segment<'a> {
 /// for segment in SpecialSegmentIterator::new(input, pattern_matcher, false) {
 ///     match segment {
 ///         Segment::SpecialToken(id) => { /* emit the special token */ }
-///         Segment::Text(chunk) => { /* tokenize this chunk */ }
+///         Segment::Text { text, input_offset } => { /* tokenize this chunk */ }
 ///     }
 /// }
 /// ```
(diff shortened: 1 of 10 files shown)


tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -1,17 +1,21 @@
...
 use std::mem;
-use std::sync::{Mutex, PoisonError};
+use std::ops::Range;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex, PoisonError};
 use std::vec::IntoIter;
 use std::{borrow::Cow, convert::TryFrom};
 
 use atomsplit::classify::classify;
+use rayon::iter::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIterator};
```
SBrandeis: Unused imports I think, don't forget to run clippy and cargo fmt!

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
```diff
@@ -648,103 +672,267 @@ impl TryFrom<&Tokenizer> for PipelineTokenizer {
...
     }
 }
 
-pub enum Inputs {
+pub enum Input {
     Single(String),
     Pair(String, String),
-    Batch(Vec<String>),
-    PairBatch(Vec<(String, String)>),
+}
```
SBrandeis, line 899: Pre-existing to this PR, but does this require to copy the string from Python to Rust?

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
```diff
@@ -648,103 +672,267 @@ impl TryFrom<&Tokenizer> for PipelineTokenizer {
...
+
+    fn len(&self) -> usize {
+        self.outputs.len()
+    }
+
+    fn encode_unit(&self) -> bool {
+        if self.cancelled.load(Ordering::Relaxed) {
+            return false;
+        }
+        let i = self.next_unit.0.fetch_add(1, Ordering::Relaxed);
```
SBrandeis: What happens when the fetch_add overflows?
ie when the counter is usize::MAX

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -648,103 +672,267 @@ impl TryFrom<&Tokenizer> for PipelineTokenizer {
...
+            return false;
+        }
+
+        let unit = &self.units[i];
+        let res = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+            let input = self
+                .inputs
+                .get(unit.seq)
+                .ok_or_else(|| format!("invalid unit index: {}", unit.seq))?;
+            let Input::Single(seq) = input else {
```
SBrandeis: to differentiate from `unit.seq` which has a different meaning
```suggestion
            let Input::Single(string) = input else {
```

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -648,103 +672,267 @@ impl TryFrom<&Tokenizer> for PipelineTokenizer {
...
+        self.completion_queue[next].store(unit.seq, Ordering::Release);
+
+        true
+    }
+
+    fn take_result(&self, seq: usize) -> Result<Vec<PipelineToken>> {
+        let unit_results = &self.outputs[seq];
+        if unit_results.len() == 1 {
+            return unit_results[0].take().unwrap_or_else(|| {
+                unreachable!("failed to take the unit's result when we expect it to be present")
```
SBrandeis: That can happen if you try to take the result before the work producing it is completed, right?

Should we return an Error here instead of panicking?

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -648,103 +672,267 @@ impl TryFrom<&Tokenizer> for PipelineTokenizer {
...
+            });
+        }
+        let mut chunks = Vec::with_capacity(unit_results.len());
+        for res in unit_results {
+            match res.take() {
+                Some(Ok(tokens)) => chunks.push(tokens),
+                Some(Err(e)) => return Err(e),
+                None => {
+                    unreachable!("failed to take the unit's result when we expect it to be present")
```
SBrandeis: What guarantees this branch is unreachable?

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -648,103 +672,267 @@ impl TryFrom<&Tokenizer> for PipelineTokenizer {
...
+                }
+            }
+        }
+        let total_tokens = chunks.iter().map(Vec::len).sum();
+        let mut out = Vec::with_capacity(total_tokens);
+        for c in &chunks {
+            out.extend_from_slice(c);
+        }
+
```
SBrandeis:
```suggestion
        let out = chunks.concat();
```

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
```diff
@@ -791,6 +989,29 @@ impl Iterator for HandleIter {
...
+                    return None;
+                }
+                loop {
+                    let seq = job.completion_queue[self.next_unit].load(Ordering::Acquire);
+                    if seq == Job::NOT_DONE {
+                        // XXX: caller has to wait for results regardless: instead of implementing some parking or
+                        // just empty spinning, we actually do useful work
+                        // it's also very simple to implement
+                        if !job.encode_unit() {
+                            std::hint::spin_loop();
```
SBrandeis: What does this do?

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
```diff
@@ -811,6 +1045,24 @@ thread_local! {
     static PRE_TOKENS_SCRATCH: RefCell<Vec<Span>> = const { RefCell::new(Vec::new()) };
 }
 
+const PARALLEL_MIN_BYTES: usize = 8 * 1024;
+
+struct Unit {
+    /// Sequence index in the [`Inputs`] batch
+    seq: usize,
```
SBrandeis: :see_no_evil:
```suggestion
    sequence_idx: usize,
```

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
```diff
@@ -811,6 +1045,24 @@ thread_local! {
     static PRE_TOKENS_SCRATCH: RefCell<Vec<Span>> = const { RefCell::new(Vec::new()) };
 }
 
+const PARALLEL_MIN_BYTES: usize = 8 * 1024;
+
+struct Unit {
```
SBrandeis: naming suggestion, feel free to disregard 
```suggestion
struct WorkUnit {
```

tokenizers/tk-encode/src/tokenizer/pipeline.rs
```diff
@@ -837,44 +1130,73 @@ impl PipelineTokenizer {
     pub fn encode(&self, inputs: impl Into<Inputs>, add_special_tokens: bool) -> EncodeHandle {
         let inputs = inputs.into();
 
+        if inputs.size_bytes() < PARALLEL_MIN_BYTES {
+            EncodeHandle::blocking(self.encode_serial(inputs, add_special_tokens))
+        } else {
+            let Some(pool) = pool() else {
+                // XXX: unable to get a pool, handle, reverting to single threaded
```
SBrandeis: Could add a `warn!` here

SBrandeis (commented):
first pass of review 🥵

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
McPatate: You can make it 0-copy from a python string all the way to the argument in rust, although somewhere in the process you need to materialise the string to utf8 from the python side which will necessarily copy the string. We simply make access from it's utf8 storage 0-copy via some unsafe, but we can hold a reference to the object making it safe.

tokenizers/tk-encode/src/tokenizer/pipeline.rs
McPatate: It should not be possible for this value to be `None`. If we call `take_result` before the work has been completed, this is a bug. Imo it shouldn't be an error for the user to handle because it is unrecoverable anyways and hard panic is valid behaviour.

tokenizers/tk-encode/src/tokenizer/pipeline.rs
McPatate: for special tokens, they'll always be set, for normal sequences we coordinate across threads via atomics to make sure that only once the last unit of a given sequence is completed we call this function. Each unit completion sets the inner `UnitResult` to `Some`. We will not call `take_result` twice.

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
McPatate: from the fn's doc:
> core::hint
>
> pub fn spin_loop()
> ------------------------------------------------------------
> Emits a machine instruction to signal the processor that it is running in
a busy-wait spin-loop ("spin lock").
>
> Upon receiving the spin-loop signal the processor can optimize its behavior by,
for example, saving power or switching hyper-threads.
>
> This function is different from [`thread::yield_now`](https://doc.rust-lang.org/stable/std/thread/fn.yield_now.html) which directly
yields to the system's scheduler, whereas `spin_loop` does not interact
with the operating system.
>
> A common use case for `spin_loop` is implementing bounded optimistic
spinning in a CAS loop in synchronization primitives. To avoid problems
like priority inversion, it is strongly recommended that the spin loop is
terminated after a finite amount of iterations and an appropriate blocking
syscall is made.
>
> **Note**: On platforms that do not support receiving spin-loop hints this
function does not do anything at all.

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
McPatate: I fear it makes the `Unit::idx` less clear afterwards actually, I think I'd rather leave it as `seq`!

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
McPatate: Well, it's a unit to work on, not a unit of work technically 🤓

tokenizers/tk-encode/src/tokenizer/pipeline.rs
McPatate: do we have logging? lol

PR #2332, opened 2026-08-11
chore: 2nd round of cleanup in preparation for 1.0.0rc0

SBrandeis:
## PipelineTokenizer benchmark

**10 / 10 models supported** — PipelineTokenizer vs `tokenizers` v0.23.1 (latest release) · pipeline: whole corpus · release: ~2 MB/fixture sample · ~10 kB chunks · cold caches · add_special_tokens on · 2/4-thread sweep per group

`df184d665 · 2026-08-11 16:00 UTC` · Intel(R) Xeon(R) Platinum 8375C CPU @ 2.90GHz · 16 cores

Per-model encode throughput vs latest release

Per-fixture encode throughput vs latest release, across models

**Decode** — the release encodes each fixture's decode sample (`add_special_tokens=true`) and both implementations decode those SAME ids with `skip_special_tokens=false`, so the comparison is decode alone. MB/s counts the input bytes the ids came from, the same denominator as the encode charts.

Per-model decode throughput vs latest release

**vs base branch** (`62a897655`) — per-model geomean ×speedup of this PR's PipelineTokenizer against the base branch's; **regressions in red**.

Per-model encode throughput vs base branch

<sub>base numbers measured in this run's jobs, on the same runners as the PR's</sub>

**Work vs base** (`62a897655`, allocation lane):

- encode allocations: ✓ exactly identical on all 150 fixtures
- decode allocations: ✓ exactly identical on all 150 fixtures

Encode allocations vs latest release

<sub>allocation lane measured on: Intel(R) Xeon(R) Platinum 8375C CPU @ 2.90GHz · glibc 2.39</sub>

Per-model memory footprint

Per-model decode memory footprint

Minimal encode binary size

<details><summary><b>bert-base-uncased</b> — normalizer-heavy WordPiece · ×4.99 vs v0.23.1 · ×1.00 vs base · decode ×1.00</summary>

bert-base-uncased speedup

bert-base-uncased input-size response

bert-base-uncased thread scaling (lang)

bert-base-uncased thread scaling (modalities)

bert-base-uncased decode speedup

bert-base-uncased decode thread scaling (lang)

bert-base-uncased decode thread scaling (modalities)

**Memory** (RSS MB, load+encode+decode): v0.23.1 7+11+0 (peak 37) · Pipeline 4+4+1 (peak 43)

[...]

PR #2333, opened 2026-08-11
fix(bpe): stop build() panicking on a merge longer than any vocab token

### LINKED ISSUE #2198: BPE `build()` panics on load for a crafted `tokenizer.json` (merge concatenation exceeds the scratch buffer)
**Summary.** A crafted `tokenizer.json` with a BPE model causes a panic during `Tokenizer::from_file` /
`from_str`, before any tokenization runs. The BPE builder writes a merge concatenation into a scratch buffer sized
for the longest *single* vocab token; a merge whose two tokens concatenate to something longer than any single vocab
entry overflows the slice and panics.

**Details.** In `tokenizers/src/models/bpe/model.rs`, `BpeBuilder::build`:

```rust
let mut buffer: Vec<u8> = vec![0; max_len];                       // max_len = longest single vocab key (bytes)
...
buffer[0..a.len()].copy_from_slice(a.as_bytes());
let b_len = b.len() - prefix_len;                                 // usize subtraction
buffer[a.len()..merge_len].copy_from_slice(&b.as_bytes()[prefix_len..]);   // panics: merge_len can exceed max_len
```

Both `a` and `b` are individually in the vocab, so each is `<= max_len`, but the write spans `[a.len() ..
a.len()+b_len]`, up to ~`2*max_len`, which exceeds the buffer whenever the merged token is longer than the longest
single vocab entry. The write happens *before* the "is the concatenation in the vocab" check. Two ways to trip it:
a slice-index-out-of-bounds panic at the write (fires in debug and release, since Rust always bounds-checks slice
ranges), and a `usize` underflow at `b.len() - prefix_len` when `continuing_subword_prefix` is longer than `b`.

**Reproduction** (`tokenizers` Python binding):
```python
import tokenizers, json
crafted = {"version":"1.0","model":{"type":"BPE","vocab":{"aa":0,"bb":1},"merges":[["aa","bb"]]}}
tokenizers.Tokenizer.from_str(json.dumps(crafted))
# -> PanicException: range end index 4 out of range for slice of length 2  (at models/bpe/model.rs:267)
```
(A well-formed tokenizer never trips this because the merged token `"aabb"` is itself a vocab entry, so `max_len`
covers it; the crash requires a merge whose result is absent from / larger than the vocab.)

**Impact.** Loading an untrusted `tokenizer.json` (e.g. from a model repository) crashes the tokenizer load before
inference — a denial of service. Native (Rust) consumers abort the process; the Python binding raises a catchable
`PanicException`.

**Suggested fix.** Validate the pair length before the write (size the buffer to `2*max_len`, or check `merge_len <=
buffer.len()` and return a build error), and guard `b.len() - prefix_len` against `prefix_len > b.len()`; ideally
move the write after the vocab-membership check so a malformed merge returns a `TokenizerBuilderError` instead of
panicking.

---
Reported by Halo Forge Labs.

axiom-of-choice:
### Summary

`BpeBuilder::build()` sizes its merge-map scratch buffer to the longest single vocab entry, but each merge writes the concatenation of two entries into it. When that concatenation is not itself in the vocab, the copy runs past the end of the buffer and panics instead of returning `MergeTokenOutOfVocabulary`. A `tokenizer.json` with `vocab: {"aa":0,"bb":1}` and `merges: [["aa","bb"]]` is enough:

```
range end index 4 out of range for slice of length 2
```

The same loop computed `b.len() - prefix_len` unchecked. A right-hand merge token shorter than `continuing_subword_prefix` underflows it: `attempt to subtract with overflow` in debug, a wrap then a slice panic in release.

Both are reachable from `Tokenizer::from_file` / `from_str`, before any tokenization runs: the deserializer funnels through `builder.build()` in `serialization.rs`.

Fixes #2198.

### Changes

- Size the buffer to twice the longest vocab entry. Both merge halves are looked up in the vocab and error out before anything is written, so that is the bound, and it is exact: `{"a":0}` with merge `("a","a")` needs 2.
- Take the right-hand suffix with `str::get(prefix_len..)` instead of an unchecked subtraction and byte slice. Out of range, or off a char boundary, now returns an error. This also makes the `SAFETY` claim on the `from_utf8_unchecked` below it true rather than assumed, since `str::get` will not hand back a non-boundary split.
- Three regression tests: two on the builder, one deserializing a crafted model JSON.

### Behavior

No change on input that worked before. `str::get` returns `Some` for every `prefix_len <= b.len()` on a char boundary, which is every case the old code completed. The existing blind-strip behavior when `b` does not actually start with the prefix is preserved on purpose; `strip_prefix` would have changed it.

One thing worth a maintainer opinion: the prefix-underflow case reports `MergeTokenOutOfVocabulary(b)`, and `b` is in the vocab, so the variant name is a slight stretch. I reused it rather than adding a variant to the public `Error` enum right before 1.0.0rc0. Happy to add a dedicated variant if you prefer.

### Evidence

The three new tests fail before the change and pass after. Verified by reverting only the two source hunks, keeping the tests, and re-running: 3 failed, 20 existing bpe tests passed, with the panic above. Restored: 23 passed. They also pass under `--release`, where the underflow manifests differently.

- `cargo test --lib`: 204 passed, 0 failed.
- `cargo clippy --all-targets --all-features -- -D warnings`: clean.
- `cargo fmt -- --check`: clean.

Not run: the Python and Node binding suites (this is a core-crate change with no API surface change), and `tokenizers/tests/`, which needs the `make test` data download.

### Risk

Low. Two hunks in one function, both on paths that panic today. The buffer grows by at most the length of the longest vocab token, allocated once per `build()`.

### Note on the in-flight 1.0 work

This loop survives unchanged on the `feat/train_encode_split` stack at `tokenizers/tk-encode/src/models/bpe/legacy/model.rs:186-220`, so #2293 and #2332 do not obsolete this fix, but it will need carrying across the directory move when that stack lands. Otherwise the bug ships in 1.0.

### DIFF
--- a/tokenizers/src/models/bpe/model.rs
+++ b/tokenizers/src/models/bpe/model.rs
@@ -248,7 +248,10 @@ impl BpeBuilder {
         } else {
             0
         };
-        let mut buffer: Vec<u8> = vec![0; max_len];
+        // A merge concatenates two vocabulary entries, and both are looked up in
+        // the vocab before anything is written here, so the result never exceeds
+        // twice the longest single entry.
+        let mut buffer: Vec<u8> = vec![0; max_len * 2];
         let merge_map: MergeMap = self
             .config
             .merges
@@ -261,11 +264,13 @@ impl BpeBuilder {
                 let b_id = vocab
                     .get(&b)
                     .ok_or_else(|| Error::MergeTokenOutOfVocabulary(b.to_owned()))?;
+                let b_suffix = b
+                    .get(prefix_len..)
+                    .ok_or_else(|| Error::MergeTokenOutOfVocabulary(b.to_owned()))?;
                 buffer[0..a.len()].copy_from_slice(a.as_bytes());
-                let b_len = b.len() - prefix_len;
-                let merge_len = a.len() + b_len;
-                buffer[a.len()..merge_len].copy_from_slice(&b.as_bytes()[prefix_len..]);
-                // SAFETY: buffer contains a concatenation of two valid UTF-8 strings, so it is itself valid UTF-8, even considering prefix_len
+                let merge_len = a.len() + b_suffix.len();
+                buffer[a.len()..merge_len].copy_from_slice(b_suffix.as_bytes());
+                // SAFETY: buffer contains a concatenation of two valid UTF-8 strings, so it is itself valid UTF-8: `str::get` rejects a prefix_len that is not a char boundary of b
                 let new_token = unsafe { from_utf8_unchecked(&buffer[..merge_len]) };
                 let new_id = vocab
                     .get(new_token)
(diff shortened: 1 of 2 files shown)


aether-png:
ran an expanded edge-case suite against this (empty-string merges, exact buffer boundary, identical max-length halves, UTF-8 boundary both sides, empty continuing-prefix) : 6/6 clean, no panics, no wrong results.

nice catch on the release-mode underflow too, easy to miss since it behaves differently under --release vs debug.

PR #2337, opened 2026-08-17
perf(pipeline): make short-document batches scale [prototype]

ArthurZucker:
**Prototype**, stacked on #2336 (base is that same commit, on a clean branch — see the note at the bottom). Prototypes the three things I proposed for the `batch` shape. One of them is the whole fix; the other two are worth about what the numbers say and no more.

## What was wrong

Batches of short documents did not benefit from the pool at all — 26k chat lines peaked at 229 MiB/s on 2 threads and *fell* to 146 on 8.

It was **not** the allocator, which is where I (and the original report) started. Allocations per document are flat at 3.28 from 2 to 14 threads while throughput halves. I also tried claiming units in runs to cut cursor traffic: worth nothing on its own. Both hypotheses dead before the profile.

**The profile settled it.** At 8 threads the workers sit in `rayon_core::WorkerThread::wait`, and the live samples are in `pthread_mutex_firstfit_lock_slow` / `__psynch_mutexwait`, with `scratch_pool` right there. `encode_sequence` borrows a scratch buffer from a **process-wide `Mutex<Vec<_>>`** and returns it — two locks per input. Free next to 40 us of work on an 8 KiB document; the entire cost next to 1 us on a 200-byte one.

## The three changes

1. **One scratch per claim, not per document.** This is the fix. Workers take the guard once and hold it across the whole claimed run.
2. **Reconstruction on the worker.** `drain` + `post_process` ran on the consumer, once per document — a serial stage every document passed through, and the reason ids were allocated on a worker and freed on the consumer. The worker whose `remaining` decrement hits zero now owns the sequence; the completion queue carries finished sequences instead of units, and the consumer stops counting units entirely.
3. **`encode_batch_flat` → `BatchEncoding`.** One contiguous id buffer plus row offsets, filled through per-chunk arenas, so allocations are per *chunk* not per document. `encode_sequence_into` / `encode_sequence_scratched` expose the buffer-writing path it needs. Falls back to `encode` for pair templates and templates with type ids.

Claiming units in runs is in here because #2 needs it to own whole documents — not for the atomics.

## Numbers

gpt2, 26k documents one per line, 5 MiB of english, M3 Max:

| threads | 1 | 2 | 4 | 8 | 14 |
|---|---|---|---|---|---|
| before | 152 | 229 | 216 | 146 | 156 |
| after | 163 | 324 | 423 | 587 | **686** |
| after, flat | 190 | 325 | 598 | 999 | **1348** |

Allocations per document on the flat path: **2.28 → 0.05** (two allocations total for the whole batch, single-threaded).

`mdoc` (619 x 8 KiB) is unaffected to slightly better: 1570 → 1656 MiB/s at 14 threads. Short-document batches are no longer a special case — 1348 against mdoc's 1656.

## Correctness

- `pipeline_oracle`: 9/9 models, 15 corpora, two window sizes, `add_special_tokens` both ways.
- New `tests/flat_batch.rs`: `encode_batch_flat` compared against `encode` document by document, 10 models, both `add_special_tokens` values, batches large enough to cross into the parallel path.
- Full `tk-encode` suite: 370 passed, 1 failed — `normalizers::precompiled::tests::pipeline_precompiled_matches_legacy`, which fails identically on the base commit (`NotFound` on a file `make fixtures models` does not fetch). clippy and fmt clean.

## Open questions before this is more than a prototype

- **Re-sweep on the real boxes.** Everything here is 14 cores and macOS libmalloc. The reported glibc arena cliff (44% of cycles in `queued_spin_lock_slowpath`, `MALLOC_ARENA_MAX` moving 0.5 → 55 MiB/s) cannot reproduce here, and changes 2 and 3 are exactly the ones that should bite it. The 88/96/128-core numbers are the ones that decide this.
- **`BatchEncoding` is a new public type** overlapping `Encoding`/`EncodingParts`. Worth deciding whether the flat form should be *the* batch API rather than a second one.
- The per-sequence `Vec<UnitResult>` the plan allocates is still one allocation per document on the non-flat path.
- `doc` (whole corpus as one sequence) is still flat at ~200 MiB/s; nothing here addresses it.

---

*Branch note:* I pushed this commit onto #2336's branch by accident before cutting a branch for it, so `perf/post-process-reuse-sequence-buffer` currently carries both commits and #2336's diff is wrong. `perf/post-process-reuse` (this PR's base) is that branch at its intended single commit. Restoring #2336 needs a force push, which I can't do — either run `git push --force-with-lease origin db9d5cca:perf/post-process-reuse-sequence-buffer`, or close #2336 and I'll reopen it from the clean branch.


ArthurZucker:
Dropped the ad-hoc `examples/batch_alloc.rs` and moved the measurement into `examples/fixture_bench.rs` as `bench_batch`, behind a new `--parallel-only` flag. It reuses the harness's median-of-`REPS`, its `CountingAlloc`, and its regime stamp, and sweeps powers of two up to the machine's parallelism instead of the fixed `[2, 4]` — which is why nothing here caught this: phase 3 sweeps threads at 10 KiB chunks and the size sweep reaches 256 B but runs single-threaded, so short-documents × many-threads fell between the two axes.

`bench/batch-sweep-baseline` is the same phase on `feat/train_encode_split` (minus the flat path, which does not exist there), so the two are directly comparable on the same corpus under the same regime.

gpt2, 20k documents, 5.9 MB, M3 Max — **all three rows from the same harness**:

| threads | 1 | 2 | 4 | 8 | 14 |
|---|---|---|---|---|---|
| baseline | 98 | 155 | 202 | 199 | **208** |
| this PR, `encode` | 129 | 305 | 503 | 671 | **718** |
| this PR, `encode_batch_flat` | 136 | 252 | 514 | 932 | **1130** |

Allocations per document: **3.86** on `encode`, **0.146** on the flat path.

Baseline flatlines at ~200 MB/s past 4 threads (2.1× best case); the branch reaches 5.6× and the flat path 8.3×. Note these are decimal MB/s over the fixture corpus, so they are not comparable with the MiB/s-over-big.txt figures in the description — the same-harness rows above supersede those.

To reproduce, on both branches:

```bash
cd tokenizers && make fixtures models
cargo run --release -p tk-encode --features bench-baseline \
    --example fixture_bench -- --model gpt2 --parallel-only
```

ArthurZucker:
## Measured on a 224-thread box (supersedes every number above)

Two runs of each branch, same harness, same corpus, `--parallel-only`, medians below. Decimal MB/s over the fixture corpus.

| threads | 1 | 2 | 4 | 8 | 16 | 32 | 64 | 128 | 224 |
|---|---|---|---|---|---|---|---|---|---|
| `bench/batch-sweep-baseline` | 48.8 | 95.6 | 115 | 139 | 130 | 156 | **162** | 143 | 125 |
| this PR, `encode` | 50.4 | 107 | 150 | 212 | 259 | 269 | 297 | **402** | 331 |
| this PR, `encode_batch_flat` | 56.3 | 87.4 | 160 | 249 | 390 | 519 | 796 | **861** | 671 |

**Scaling 3.3x -> 8.0x (`encode`) -> 15.3x (flat).** Peak-to-peak 162 -> 402 (2.5x) -> 861 (5.3x).

The baseline is the finding as much as the fix is: **1 -> 64 threads buys it 3.3x, and past 64 it goes backwards.** That is the shape the original report described, reproduced in-tree.

### Single-thread: no regression

An earlier run showed 86.8 -> 50.1 at one thread, which looked like a 42% regression. It was the harness: `bench_batch` had no warmup, so the first thread count in the sweep absorbed first-touch page faults on the 20k-document vector. With a discarded pass per count, both runs agree — baseline 48.8/48.8, this PR 50.2/50.6, i.e. ~3% faster. Fixed on both branches.

### Still open

- **Both paths peak at 128 and drop at 224.** SMT tail; #2335 (default the pool to physical cores) caps it at the physical count and should recover part of this. NUMA-aware unit distribution is the remaining piece and is in none of these PRs.
- **`encode` plateaus 16-64 (259/269/297) before jumping at 128.** Not explained. The flat path climbs smoothly through the same region, so it is something in the `Vec<Encoding>` path rather than the work itself.
- **Per-cell variance is ~±20%** on this machine (flat at 32 threads: 420 vs 617 across the two runs). Read the shape, not individual cells; `REPS = 3` is thin for a box this size.
- The earlier M3 Max figures in the description are MiB/s over a different corpus and should be ignored in favour of this table.

Issue #629, opened 2021-02-13
Encoding.merge yields inconsistent sequence_ids


jvamvas:
The Encoding.merge method does not create a list of `sequence_ids` that could be expected given the documentation:

```python
from transformers import BertTokenizerFast
from tokenizers import Encoding

seq1 = "a b"
seq2 = "c d"
tokenizer = BertTokenizerFast.from_pretrained("bert-base-cased")

encoding1 = tokenizer.encode_plus(seq1, add_special_tokens=False)[0]
encoding2 = tokenizer.encode_plus(seq2, add_special_tokens=False)[0]
print(encoding1.sequence_ids)
# [0, 0]
print(encoding2.sequence_ids)
# [0, 0]
print(Encoding.merge([encoding1, encoding2]).sequence_ids)
# [None, None, 0, 0]
```

Expected could be either `[0, 0, 0, 0]` or `[0, 0, 1, 1]`. I think the latter would be preferable, but then `n_sequences` would also have to be fixed, as it is currently set to 1:

```python
print(Encoding.merge([encoding1, encoding2]).n_sequences)
# 1
```

PR #2341, opened 2026-08-18
fix: reject invalid sequence IDs in encoding mappings

tandede:
## What does this PR do?

Keep sequence-ID mapping behavior consistent after the Tokenizers v1 reorganization.

For an encoding with no explicit `sequence_ranges`, only sequence ID 0 denotes the implicit full-length sequence. When ranges are present, mapping APIs now require an exact stored ID instead of silently falling back to sequence 0. `get_sequence_ids()` iterates the IDs actually stored in the map, including sparse IDs such as 5. Assigning an ID to the whole encoding also clears previous ranges, so repeated `set_sequence_id(5); set_sequence_id(6)` calls cannot leave overlapping ranges and make public Rust mapping methods disagree.

The original PR included Python-binding regressions. Upstream's Tokenizers v1 merge removed the old Python `Encoding` sequence-mapping API and moved the Rust implementation into `tk-encode`; the obsolete Python tests were therefore not carried forward. The current diff and regression tests are confined to the Rust encoding module. This does not claim that the removed Python API is covered on v1.

## Tests

- `cargo test -p tk-encode --lib --quiet` in `tokenizers/`: 191 passed, 1 ignored.
- `cargo fmt -p tk-encode --check` in `tokenizers/`: passed.

### DIFF
--- a/tokenizers/src/tokenizer/encoding.rs
+++ b/tokenizers/src/tokenizer/encoding.rs
@@ -137,9 +137,10 @@ impl Encoding {
     pub fn get_sequence_ids(&self) -> Vec<Option<usize>> {
         let mut sequences = vec![None; self.len()];
         for seq_id in 0..self.n_sequences() {
-            let range = self.sequence_range(seq_id);
-            let seq_len = range.len();
-            sequences.splice(range, std::iter::repeat_n(Some(seq_id), seq_len));
+            if let Some(range) = self.sequence_range(seq_id) {
+                let seq_len = range.len();
+                sequences.splice(range, std::iter::repeat_n(Some(seq_id), seq_len));
+            }
         }
         sequences
     }
@@ -201,11 +202,12 @@ impl Encoding {
 
     /// Returns the range to target to retrieve something (word_id, offsets, ..) related to the
     /// given sequence id
-    fn sequence_range(&self, sequence_id: usize) -> Range<usize> {
-        self.sequence_ranges
-            .get(&sequence_id)
-            .cloned()
-            .unwrap_or(0..self.len())
+    fn sequence_range(&self, sequence_id: usize) -> Option<Range<usize>> {
+        if self.sequence_ranges.is_empty() {
+            (sequence_id == 0).then(|| 0..self.len())
+        } else {
+            self.sequence_ranges.get(&sequence_id).cloned()
+        }
     }
 
     /// Returns the index of the sequence containing the given token
@@ -229,7 +231,7 @@ impl Encoding {
     /// with the form (start_token, end_token + 1)
     pub fn word_to_tokens(&self, word: u32, sequence_id: usize) -> Option<(usize, usize)> {
         let (mut start, mut end) = (None, None);
-        let sequence_range = self.sequence_range(sequence_id);
+        let sequence_range = self.sequence_range(sequence_id)?;
 
         self.words
             .get(sequence_range.clone())?
@@ -283,7 +285,7 @@ impl Encoding {
 
     /// Get the token that contains the given char.
     pub fn char_to_token(&self, pos: usize, sequence_id: usize) -> Option<usize> {
-        let sequence_range = self.sequence_range(sequence_id);
+        let sequence_range = self.sequence_range(sequence_id)?;
 
         self.offsets
             .get(sequence_range.clone())?
@@ -603,6 +605,27 @@ mod tests {
                 ..Default::default()
             }
         );
+    }
+
+    #[test]
+    fn mapping_rejects_invalid_sequence_ids() {
+        let encoding = Encoding {
+            ids: vec![1],
+            tokens: vec![String::from("Hello")],
+            words: vec![Some(0)],
+            offsets: vec![(0, 5)],
+            ..Default::default()
+        };
+
+        assert_eq!(encoding.word_to_tokens(0, 0), Some((0, 1)));
+        assert_eq!(encoding.word_to_chars(0, 0), Some((0, 5)));
+        assert_eq!(encoding.char_to_token(0, 0), Some(0));
+        assert_eq!(encoding.char_to_word(0, 0), Some(0));
+
+        assert_eq!(encoding.word_to_tokens(0, 1), None);
+        assert_eq!(encoding.word_to_chars(0, 1), None);
+        assert_eq!(encoding.char_to_token(0, 1), None);
+        assert_eq!(encoding.char_to_word(0, 1), None);
     }
 
     #[test]
--- a/tokenizers/src/tokenizer/encoding.rs
+++ b/tokenizers/src/tokenizer/encoding.rs
@@ -135,12 +135,13 @@ impl Encoding {
     }
 
     pub fn get_sequence_ids(&self) -> Vec<Option<usize>> {
+        if self.sequence_ranges.is_empty() {
+            return vec![Some(0); self.len()];
+        }
+
         let mut sequences = vec![None; self.len()];
-        for seq_id in 0..self.n_sequences() {
-            if let Some(range) = self.sequence_range(seq_id) {
-                let seq_len = range.len();
-                sequences.splice(range, std::iter::repeat_n(Some(seq_id), seq_len));
-            }
+        for (seq_id, range) in &self.sequence_ranges {
+            sequences[range.clone()].fill(Some(*seq_id));
         }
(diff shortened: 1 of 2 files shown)


tokenizers/src/tokenizer/encoding.rs
```diff
@@ -137,9 +137,10 @@ impl Encoding {
     pub fn get_sequence_ids(&self) -> Vec<Option<usize>> {
         let mut sequences = vec![None; self.len()];
         for seq_id in 0..self.n_sequences() {
-            let range = self.sequence_range(seq_id);
-            let seq_len = range.len();
-            sequences.splice(range, std::iter::repeat_n(Some(seq_id), seq_len));
+            if let Some(range) = self.sequence_range(seq_id) {
```
Jokasa7: **[P2] Iterate the sequence IDs actually stored in the range map**

`n_sequences()` returns the number of entries in `sequence_ranges`, not the largest key plus one, while the public Rust and Python `Encoding.set_sequence_id()` APIs accept an arbitrary non-negative ID. Consequently, if an encoding is assigned sequence ID 5, this loop probes only ID 0 and the new conditional silently skips the sole stored range.

I reproduced this with the exact head Python binding: after `encoding.set_sequence_id(5)`, `token_to_sequence(0)` is `5` and `word_to_tokens(0, 5)` is `(0, 1)`, but `encoding.sequence_ids` is `[None, None]`. The exact base returned `[0, 0]`, which was also incorrect, so the mapping fix leaves these public APIs mutually inconsistent. This is related to the older stale report in #629, but the current PR has no test for an explicitly assigned nonzero ID.

Could the non-empty case iterate the actual `(seq_id, range)` entries in `self.sequence_ranges`, retaining a separate empty-map fallback for the implicit sequence 0, and add a `set_sequence_id(5)` regression?

Jokasa7 (commented):
Thanks for tightening the invalid-sequence-ID behavior. I reviewed exact head `7e45e0ace87f9a92d56f98c1c507ef7271ea16fe` and built both the exact base and head Python bindings locally.

The direct mapping APIs now correctly stop falling back to sequence 0 when an explicit range map does not contain the requested ID. I found one remaining consistency problem in `get_sequence_ids()`, though: it still assumes that the stored sequence IDs are dense from zero, while the public `Encoding.set_sequence_id()` API can create a range keyed by any non-negative ID.

On the exact head, after `encoding.set_sequence_id(5)`, `token_to_sequence(0)` returns `5` and `word_to_tokens(0, 5)` returns `(0, 1)`, but `encoding.sequence_ids` returns `[None, None]`. I left the exact reproduction and suggested regression shape inline.

tokenizers/src/tokenizer/encoding.rs
tandede: > **[P2] Iterate the sequence IDs actually stored in the range map**
> 
> `n_sequences()` returns the number of entries in `sequence_ranges`, not the largest key plus one, while the public Rust and Python `Encoding.set_sequence_id()` APIs accept an arbitrary non-negative ID. Consequently, if an encoding is assigned sequence ID 5, this loop probes only ID 0 and the new conditional silently skips the sole stored range.
> 
> I reproduced this with the exact head Python binding: after `encoding.set_sequence_id(5)`, `token_to_sequence(0)` is `5` and `word_to_tokens(0, 5)` is `(0, 1)`, but `encoding.sequence_ids` is `[None, None]`. The exact base returned `[0, 0]`, which was also incorrect, so the mapping fix leaves these public APIs mutually inconsistent. This is related to the older stale report in #629, but the current PR has no test for an explicitly assigned nonzero ID.
> 
> Could the non-empty case iterate the actual `(seq_id, range)` entries in `self.sequence_ranges`, retaining a separate empty-map fallback for the implicit sequence 0, and add a `set_sequence_id(5)` regression?

Thanks for the detailed review and exact reproduction. You were right: `n_sequences()` reports the number of stored ranges, so it cannot be used to enumerate non-dense sequence IDs.

I updated `get_sequence_ids()` to retain the implicit sequence-0 fallback when `sequence_ranges` is empty and otherwise iterate the actual `(seq_id, range)` entries. I also added Rust and Python regressions using `set_sequence_id(5)` and verified that `token_to_sequence()`, `word_to_tokens()`, and `sequence_ids` now agree.

Validation:

- Rust core `make test`: all unit, integration, and documentation tests passed
- Rust `make lint`: rustfmt and Clippy passed
- Python bindings `make test`: 204 passed, 3 skipped; 20 Rust binding tests passed
- Python `make check-style`: Ruff, formatting, and type checks passed
- `git diff --check`: passed

Jokasa7: Verified against current head `2607be5421ddfb8eaa8c2ddd07cee310a0590b42`. The non-empty path now iterates the stored `(seq_id, range)` entries, while the empty map still returns implicit sequence 0. The new Rust and Python `set_sequence_id(5)` regressions cover the exact inconsistency, including `token_to_sequence`, `word_to_tokens`, and `sequence_ids`. This resolves the issue I reported. Thanks for the focused fix and validation.

tokenizers/src/tokenizer/encoding.rs
```diff
@@ -135,11 +135,13 @@ impl Encoding {
...
+        if self.sequence_ranges.is_empty() {
+            return vec![Some(0); self.len()];
+        }
+
         let mut sequences = vec![None; self.len()];
-        for seq_id in 0..self.n_sequences() {
-            let range = self.sequence_range(seq_id);
-            let seq_len = range.len();
-            sequences.splice(range, std::iter::repeat_n(Some(seq_id), seq_len));
+        for (seq_id, range) in &self.sequence_ranges {
```
Jokasa7, line 144: **[P2] Replace prior sequence ranges when assigning the whole encoding**

`set_sequence_id()` inserts a new full-length range without removing existing entries. Reassigning an encoding with `set_sequence_id(5); set_sequence_id(6)` therefore leaves two overlapping `0..len` ranges. `token_to_sequence()` returns the first matching `AHashMap` entry, while this new loop overwrites each token with the last matching entry, so the two public APIs necessarily disagree and which ID each reports depends on hash iteration order. The added tests only call the setter once on an empty map.

Because the API says it sets the sequence ID for the whole encoding, could the setter clear the prior ranges before inserting and add a repeated-assignment regression? This overlapping state predates the PR; the issue here is preserving cross-API consistency with the new getter.

Jokasa7 (commented):
Thanks for addressing the original sparse-ID inconsistency. I re-reviewed current head `2607be5421ddfb8eaa8c2ddd07cee310a0590b42`; that finding is fixed. I found one separate consistency edge case when the public setter is called more than once, and left it inline. This comes from an existing setter state interacting with the new map iteration, not solely from this patch.

tokenizers/src/tokenizer/encoding.rs
tandede: > **[P2] Replace prior sequence ranges when assigning the whole encoding**
> 
> `set_sequence_id()` inserts a new full-length range without removing existing entries. Reassigning an encoding with `set_sequence_id(5); set_sequence_id(6)` therefore leaves two overlapping `0..len` ranges. `token_to_sequence()` returns the first matching `AHashMap` entry, while this new loop overwrites each token with the last matching entry, so the two public APIs necessarily disagree and which ID each reports depends on hash iteration order. The added tests only call the setter once on an empty map.
> 
> Because the API says it sets the sequence ID for the whole encoding, could the setter clear the prior ranges before inserting and add a repeated-assignment regression? This overlapping state predates the PR; the issue here is preserving cross-API consistency with the new getter.

Thanks for catching this. I updated `set_sequence_id()` to replace the previous ranges and added Rust and Python regressions for repeated assignment. All local tests pass.

Jokasa7: Verified against current head `cdff60f2e72889c0ca163b917a8c19eba0a9938c`. `set_sequence_id()` now clears the previous range map before inserting the new full-length range, so assigning 5 and then 6 leaves exactly one sequence range instead of two overlapping entries. The new Rust and Python regressions cover the important cross-API behavior: the old ID no longer resolves, `n_sequences` remains 1, and both `token_to_sequence()` and `sequence_ids` report 6. This resolves the repeated-assignment inconsistency I reported. Thanks for the focused follow-up.

PR #2345, opened 2026-08-19
feat(tk-encode): slim by default — read tokenizer.json with no serde (−47% gzipped)

ArthurZucker:
Stacked on #2344. Takes the size win from #2293 **without** the `.tok` container and without a forked dependency — `tk-encode` still reads a plain `tokenizer.json`, and it stays publishable to crates.io.

| build | stripped | gzipped |
|---|---:|---:|
| `config` (what the umbrella and both bindings ship) | 1,416,288 | 666,281 |
| **slim default** | **736,192** | **349,997** |
| | **−48%** | **−47%** |

For contrast, #2293 reached 286 KB gzipped, but needed a `.tok` binary container *and* `[patch.crates-io] ptr_hash = { git = "…" }`, which cannot be published.

## The shape

`PipelineTokenizer::try_from(&Tokenizer)` already *lowered* every wrapper enum into a pipeline-native type and threw the wrapper away — so the wrappers were only ever a deserialization target. And the bindings never touch `PipelineTokenizer` (60 `Tokenizer` references in `bindings/python/src/tokenizer.rs`, zero pipeline references in either binding). That makes the split fall out:

- **`config` off** (default): a hand-rolled JSON reader builds the `Pipeline*` types directly. No serde, no `serde_json`, no `monostate`.
- **`config` on**: `Tokenizer`, the five wrapper enums, serde and all the BC. Unchanged for python and node.

`cargo tree -p tk-encode -e normal -i serde` at default features:

```
serde v1.0.228
└── ptr_hash v2.0.1
    └── tk-encode v0.23.2-dev.0
```

Only `ptr_hash`, which declares serde non-optional upstream. `-i serde_json` prints nothing; `-i monostate` matches no packages.

## Verification

**`json_oracle`** — 10 models × 15 fixtures × 2 `add_special_tokens` = **300 comparisons**, each one comparing the slim reader's ids against the config path's *in the same process*:

```
300/300 ok, 0 failed, 0 model(s) skipped, 300 slim comparison(s)
0 slim mismatches · 0 digest drifts · 0 unsupported
```

Every cell compared, every id identical, and nothing that ships today changed. The gate was verified to actually fail: corrupting one digest produces `DRIFT want=… got=…` and exit 1.

**`scripts/feature_matrix.sh`** — 8/8 configurations pass `check` + `clippy -D warnings` + tests (264 / 264 / 284 / 270 / 286 / 285 / 312 / 416). Test counts tracking the gates is the useful signal: each rung compiles genuinely different code.

Both bindings build **unmodified**, including `bindings/node` standalone — it has its own workspace and lockfile, which is where #2293's version would have failed, because nothing there passed `tk-encode/config` and it only compiled via workspace feature unification.

## Floats: deliberate bug-compatibility

`serde_json` without `float_roundtrip` (off by default) accumulates digits into a `u64` and applies one multiply/divide by a power of ten. Past 2^53 the `as f64` rounds and the divide rounds again — up to 1 ULP off. On t5-base that is **8,334 of 32,100 Unigram scores**, and since scores feed a Viterbi lattice it flips a near-tie about **twice per 1.25M tokens**.

`f64::from_str` is correctly rounded, so using it made the slim reader *disagree* with the config path on 20 cells. Enabling `serde_json/float_roundtrip` was measured and does fix that (20 mismatches → 0) — but at the cost of 20 digest drifts, i.e. it changes ids that ship today.

Correct or not, **ids must not move**. So the reader reproduces serde's arithmetic instead, including its `POW10` table and its overflow-drops-digits rule. Two tests state this out loud, and one fails if `serde_json` ever becomes correctly rounded by default, so the emulation gets deleted rather than silently outliving its reason.

Side effect: it is *smaller*. Rust's `dec2flt` is a full correctly-rounded algorithm; a `u64` accumulate plus a table lookup is much less code — 357,493 → 349,997 gzipped.

## Legacy shapes

The slim reader accepts exactly two, because between them they cover most of the Hub and `gpt2.json` hits both:
- a `model` with no `"type"` — inferred from which keys are present, ~4 lines where the untagged machinery takes ~480
- `merges` spelled `"a b"` rather than `["a","b"]`

Everything else legacy is refused with a message naming what to convert. Notably, **every component object in all 22 real configs carries its `"type"`** — the untagged wrapper machinery defends a shape that does not occur in practice.

## New tooling

- `scripts/feature_matrix.sh` — the regression class it exists for has bitten this branch **seven times** (a test borrowing a type from a feature it does not name; an `_ => ()` arm unreachable with only BPE+Unigram; a `Model as _` import needed by two of three models). Every one was caught by this script, none by review. Added to `all-checks`.
- `scripts/size_ladder.sh` — per-rung stripped/gzipped with deltas, `--json` for CI. The CI job reports into the step summary and deliberately does **not** gate on growth; there is no agreed budget and a job that reddens on any increase gets ignored. It still fails if a rung does not build.

Per-rung size deltas are order-dependent (removing serde first makes some of the same code unreachable), so only the total is a clean claim.

## Not in here

- `tk-serialization` extraction — moving `Tokenizer` + the wrappers + the BC into their own crate. The `config` boundary makes it mechanical; it is in progress separately.
- `bpe` is an honest marker: BPE is still compiled unconditionally, and `default = ["bpe"]` states intent. Gating it for real means threading cfg through `PipelineBPE`, the byte-level fold and the merge queues, and nobody wants a BPE-less build — the ask was BPE-*only*. Noted in the manifest.


ArthurZucker:
Pushed two more things: the `tk-serialization` extraction, and a fixture-pinning fix that turned out to matter.

## `tk-serialization` extracted

`tk-encode` is now runtime-only. Moved out (~5,000 lines): all five wrapper enums with their hand-written `Deserialize` impls, `Tokenizer`/`TokenizerImpl`/`TokenizerBuilder`, the legacy `AddedVocabulary`, `tokenizer/serialization.rs`, `DecodeStream`, and the whole `TryFrom<&Tokenizer>` lowering.

`PipelineTokenizer::from_parts` is now the only constructor of `TokenizerInner`, so the slim reader and the config path assemble through one place and `TokenizerInner` never became public. Better than expected: **`PipelineNormalizer` lost its `Declared(NormalizerWrapper)` variant entirely** — the config path now flattens normalizer sequences exactly as the slim reader does, so that is one code path where there were two.

**Three things could not move**, all blocked by the orphan rule or the bindings rather than by effort:

- the four per-model `serialization.rs` + `OrderedVocabIter` — `impl Deserialize for BPE` cannot be written from `tk-serialization` while `BPE` lives in `tk-encode`, and `BPE` has to stay there (the pipeline consumes it, the slim reader builds it). A free function instead would remove the impl from public `tokenizers::models::bpe::BPE`, which the node binding uses via `#[serde(flatten, with = "arc_rwlock_serde")]`.
- `Model::save` — a trait method both bindings implement inside their own `impl Model`.
- `BPE::from_file` / `WordPiece::from_file` / `WordLevel::from_file` and the `files` route — inherent associated functions called as `Type::fn(..)` from 11 binding call sites; an extension trait would need a `use` at each.

That residue stays behind `tk-encode`'s `config`, which is now documented as "the leaf components' own serde derives" and is what `tk-serialization` switches on.

**No re-export lost.** `tokenizers::Tokenizer`, `tokenizers::models::bpe::BPE`, `tokenizers::normalizers::NFD`, the three distinct `Sequence` types, `step_decode_stream` — all resolve unchanged. **No binding source, manifest or stub was modified**; the only `bindings/` diff is 17 lines in each `Cargo.lock` for the new crate, which is mechanically unavoidable. Both build, including `bindings/node` standalone.

## The fixture bug, which invalidated numbers I had already posted

The `json_oracle` digests were recorded against a **stale `data/llama-2.json`**. `make models` uses existence-based targets, so it never re-downloads a fixture you already have, and `HF_REVISION` defaulted to `main`. Between them, one machine's months-old copy became the committed baseline — and **every "300/300, 0 drifts" figure in this PR's description would have shown 4 spurious drifts on a fresh clone.**

Found because a fresh worktree disagreed. Force-refreshing the ten bench fixtures reproduced the four drifts locally with `got` hashes byte-identical to the other environment's — same code, same fixtures, same ids. So it was never a regression; the baseline was recorded against the wrong file.

Fixed at the root:
- `HF_REVISION` pinned to `e20ac5914f5ada06874e4f9d45e49b31a049904e`, with a comment saying to bump it and re-record digests in the same commit.
- The two CI `hf download` lines had **no `--revision` at all**; both pinned.
- All 300 digests re-recorded.

## Verification after both changes

```
json_oracle:  300/300 ok, 0 failed, 300 slim comparison(s), 0 drifts, 0 mismatches
feature_matrix.sh: 8/8 pass (238/238/258/243/241/258/266/334 tests)
cargo tree -p tk-encode -e normal -i serde  ->  ptr_hash only
clippy --workspace --all-targets --all-features -- -D warnings   exit 0
cargo fmt --all -- --check                                       exit 0
tk-encode / tk-serialization / tokenizers / tk-train             all exit 0
bindings/python, bindings/node (standalone)                      exit 0
size ladder, default rung: 348,309 gzipped
```

## One caveat on the size ladder

`binsize_pipeline` no longer has a `config` arm — there is no `Tokenizer` in `tk-encode` to deserialize into — so the ladder's top two rungs now measure "leaf serde derives compiled but unreachable" rather than a linked config layer. The default rung (348,309) is still honest, but rungs 1–2 no longer mean what their labels say. Getting the config layer's real footprint back on the ladder needs a `binsize_config` example in `tk-serialization`; flagging rather than quietly leaving misleading labels in the CI summary.

tokenizers/tk-convert/src/convert.rs
ArthurZucker, line 1: genuinely new code

tokenizers/tk-train/src/trainers/wordlevel.rs
ArthurZucker, line 1: we removed derive builder's dependency

tokenizers/tk-convert/src/convert.rs
ArthurZucker: It's just the BC code used by `python` to read any of the old tokenizer.json files.

tokenizers/tk-encode/src/models/bpe/serialization.rs
ArthurZucker, line 1: To manually check and re-write

tokenizers/tk-serialize/src/from_json/mod.rs
ArthurZucker, line 1: this and the entire of tk-serialize are new

tokenizers/tk-encode/src/models/bpe/model.rs
```diff
@@ -50,14 +49,64 @@ fn content_start(bytes: &[u8]) -> usize {
...
 const _: () = assert!(align_of::<PipelineToken>() == align_of::<u32>());
 
+/// Everything a [`PipelineBPE`] needs besides its vocabulary and its merge list.
+///
+/// A `Default`-able struct rather than a nine-argument constructor, so a caller spells only the
+/// fields its config actually declares — which for most of the Hub is `with_byte_level` and nothing
+/// else. It carries no serde: naming these options in a config file is the readers' business, and
+/// one of those readers ([`tk-serialize`](https://docs.rs/tk-serialize)) links no serde at all.
+pub struct PipelineBpeOptions {
+    /// The token to emit for a character with no vocabulary entry. Must itself be in the vocab.
```
ArthurZucker: has nothing to do here AFAIK and can just be in the serialization?

tokenizers/tk-encode/src/pre_tokenizers/whitespace.rs
ArthurZucker, line 1: to actually check

PR #2363, opened 2026-08-25
feat: padding in PipelineTokenizer

SBrandeis:


tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
```diff
@@ -207,6 +208,8 @@ struct TokenizerInner {
     /// the special-token metadata that used to need a separate `tokenizer_config.json`. Empty
     /// when the config declares none. `BTreeMap` so the writer emits a stable key order.
     role_to_token: BTreeMap<String, String>,
+    /// Padding configuration, can be overridden at runtime with [`EncodeHandle::wait_with_padding`].
+    padding: Option<PaddingParams>,
```
ArthurZucker, line 212: actually great, this used to not be serialized and was in the `tokenizer_config.json`. Perfect

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
```diff
@@ -1116,4 +1158,143 @@ mod tests {
             ]
         );
     }
+
+    #[test]
```
ArthurZucker, line 1162: fiel is already big as is, we can probably move this around

tokenizers/tk-encode/src/tokenizer/encoding.rs
```diff
@@ -466,80 +463,6 @@ impl Encoding {
         self.attention_mask.extend(pair.attention_mask);
         self.overflowing = overflowings;
     }
-
-    pub fn pad(
```
ArthurZucker, line 470: nice to si it gone!

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
```diff
@@ -401,7 +410,22 @@ impl EncodeHandle {
...
+        let padding = self.padding.take();
+        self.wait_with_padding(padding.as_ref())
+    }
+
+    /// [`wait`](Self::wait), with `params` standing in for the padding the tokenizer was built
+    /// with. `None` pads nothing, so it is how a caller turns a configured padding off.
+    pub fn wait_with_padding(self, params: Option<&PaddingParams>) -> Result<Vec<Encoding>> {
+        let mut out = self.wait_inner()?;
+        if let Some(params) = params {
+            pad_encodings(&mut out, params)?;
```
ArthurZucker, line 423: sorry, I might not have read everything, writing down still. It's true that we have to wait to get the maximum padding length, but we can already allocate "more" (the max length of the input sequence). And then we just have to slice ? 

Of course I guess there's a tradeoff, but if our pipeline is optimized and we can granularly wait for a single output -> copy into the padding out or vice versa, the overlap means we are gonna be faster than copying the entirety of the output. WDYT

tokenizers/tk-encode/src/utils/padding.rs
```diff
@@ -58,15 +59,11 @@ pub fn pad_encodings(encodings: &mut [Encoding], params: &PaddingParams) -> Resu
         PaddingStrategy::BatchLongest => {
             #[cfg(feature = "parallelism")]
             {
-                encodings
-                    .maybe_par_iter()
-                    .map(|e| e.get_ids().len())
-                    .max()
-                    .unwrap()
+                encodings.maybe_par_iter().map(Encoding::len).max().unwrap()
```
ArthurZucker: Well, this is probably very fast even without par iter(), won't be the bottleneck, its simple so fine by me!

ArthurZucker: but we are traversing yet again (post processor traverses as well, tokenizers_spans would be a good place to compute the max

ArthurZucker: Or store the length in an array? 
AoS vs SoA here can bite us: llvm will def not compile this into a SIMD max

tokenizers/tk-encode/src/utils/padding.rs
```diff
@@ -80,65 +77,80 @@ pub fn pad_encodings(encodings: &mut [Encoding], params: &PaddingParams) -> Resu
...
-        });
+        encodings
+            .iter_mut()
+            .for_each(|encoding| pad_one(encoding, pad_length, params));
     }
 
     Ok(())
 }
 
+fn pad_one(encoding: &mut Encoding, target_length: usize, params: &PaddingParams) {
```
ArthurZucker, line 73: This is the core function! 

super fine for rc0. For v1 I am wondering if we can't:
- allocate before we encode a potentially bigger vectore
- copy on the fly, when the tokenizer_span is done, into the padded array 
- avoiding O(n) done again, we would just fill with 0 / fill with 1 and the crop the final pad

tokenizers/tk-serialize/src/from_json/tests.rs
```diff
@@ -117,6 +119,11 @@ const REFUSED: &[(&str, &str, &str)] = &[
     ("pre_tokenizer", r#"{"type": "Metaspace", "replacement": "▁", "prepend_scheme": "always"}"#,
         "`Metaspace` pre-tokenizer"),
     ("post_processor", r#"{"type": "BertProcessing", "sep": ["b"], "cls": ["a", 0]}"#, "[token, id] pair"),
+    ("padding", r#"{"direction": "Right", "pad_id": 0, "pad_type_id": 0, "pad_token": "[PAD]"}"#, "no `strategy`"),
+    ("padding", r#"{"strategy": "Invented", "direction": "Right", "pad_id": 0, "pad_type_id": 0,
+        "pad_token": "[PAD]"}"#, "unknown padding strategy"),
```
ArthurZucker, line 124: if we have the ID, the token is in the added vocab / vocab -> duplicate info. Same for the type id no?

ArthurZucker (approved):

tokenizers/tk-encode/src/utils/padding.rs
SBrandeis: Good point, I think it makes a lot of sense to make `Encoding` a struct of arrays
I'll do that in a follow-up PR

tokenizers/tk-serialize/src/from_json/tests.rs
SBrandeis: The `padding` values never reach the vocab / added vocab I think

tokenizers/tk-encode/src/utils/padding.rs
SBrandeis: > Well, this is probably very fast even without par iter(), won't be the bottleneck, its simple so fine by me!

Removed maybe_par_iter which turned out to be slower than serial (probably contention issues)

PR #2364, opened 2026-08-25
🐛 Fix inline vocab key packing on big-endian targets

gkumbhat:
### Changes

`key_and_hash_readable` takes the fast path for a word of ≤ 7 bytes by reinterpreting 8 readable bytes as a `u64`:

```
    let raw = unsafe { word.as_ptr().cast::<u64>().read_unaligned() };
```

That load is native-endian, but the key layout it feeds is defined little-endian.  `key_and_hash` builds the same key with `u32::from_le_bytes` plus shifts, with the length in the top byte. The two only agree on a little-endian host.

On big-endian platforms, they diverge two ways: `KEY_MASK[len]` keeps the low len bytes, which after a native load are `word[8-len..8]` rather than `word[0..len];` and `LEN_TAG (len << 56)` lands on top of `word[0]`. So the fast path returns a different key and hash than the stitched pack for the same word, and vocab and word-cache lookups miss or collide.

The fix reads the bytes and interprets them little-endian explicitly

### DIFF
--- a/tokenizers/tk-encode/src/vocab/bucket_vocab_store.rs
+++ b/tokenizers/tk-encode/src/vocab/bucket_vocab_store.rs
@@ -447,6 +447,57 @@ impl BucketVocabStore {
 mod tests {
     use super::*;
 
+    // The key is a little-endian pack of the word with its length on top -- not whatever a `u64`
+    // load of those bytes happens to give. Built with shifts here so the expected value is the
+    // same on a big-endian target.
+    #[test]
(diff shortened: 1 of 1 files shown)


SBrandeis:
Thanks for this fix, lgtm!
cc @ArthurZucker who owns this code as well

gkumbhat:
Hi @ArthurZucker , does this PR look good to you?

ArthurZucker (approved):
Yep sorry!

PR #1136, opened 2022-12-25
[FIX] In CharBPETokenizer, when Vocab or merges is None, unk_token cannot be used.

SeongBeomLEE:
Same Error #1120

Thanks.

## before
char_level_bep.py
```
if vocab is not None and merges is not None:
    tokenizer = Tokenizer(
        BPE(
            vocab,
            merges,
            dropout=dropout,
            unk_token=str(unk_token),
            end_of_word_suffix=suffix,
        )
    )
else:
    tokenizer = Tokenizer(BPE())
```

main.py
```
from tokenizers import CharBPETokenizer

tokenizer = CharBPETokenizer(
    unk_token = "[UNK]",
    suffix = "</w>",
)

tokenizer.train(
    files = './vocab.txt',
    vocab_size = 1000,
    min_frequency = 1,
    special_tokens = ["[PAD]", "[BOS]", "[EOS]", "[UNK]", "[SEP]", "[CLS]", "[MASK]"],
)

line = "나는 😀 😃 😄 축구를 😀 😃 😄 좋아한다."
pieces = tokenizer.encode(line)
print("CharBPETokenizer", pieces.tokens, tokenizer.decode(pieces.ids))

output:

CharBPETokenizer ['나는</w>', '축구를</w>', '좋아한다</w>', '.</w>'] 나는 축구를 좋아한다 .
```
## after

char_level_bep.py
```
if vocab is not None and merges is not None:
    tokenizer = Tokenizer(
        BPE(
            vocab,
            merges,
            dropout=dropout,
            unk_token=str(unk_token),
            end_of_word_suffix=suffix,
        )
    )
else:
    tokenizer = Tokenizer(BPE(unk_token=str(unk_token),))
```

main.py
```
from tokenizers import CharBPETokenizer

tokenizer = CharBPETokenizer(
    unk_token = "[UNK]",
    suffix = "</w>",
)

tokenizer.train(
    files = './vocab.txt',
    vocab_size = 1000,
    min_frequency = 1,
    special_tokens = ["[PAD]", "[BOS]", "[EOS]", "[UNK]", "[SEP]", "[CLS]", "[MASK]"],
)

line = "나는 😀 😃 😄 축구를 😀 😃 😄 좋아한다."
pieces = tokenizer.encode(line)
print("CharBPETokenizer", pieces.tokens, tokenizer.decode(pieces.ids))

output:

[...]

PR #2367, opened 2026-08-27
Pass dropout through in ByteLevelBPETokenizer

VenishPaneliya:
`ByteLevelBPETokenizer.__init__` forwards `dropout` to `BPE(...)` only when both `vocab` and `merges` are supplied. The other branch — the train-from-scratch path — builds a bare `BPE()`:

```python
    tokenizer = Tokenizer(
        BPE(
            vocab,
            merges,
            dropout=dropout,
            ...
        )
    )
else:
    tokenizer = Tokenizer(BPE())
```

so `dropout` is silently discarded, and the object ends up disagreeing with itself:

```python
t = ByteLevelBPETokenizer(dropout=0.1)
t._parameters["dropout"]     # 0.1
t._tokenizer.model.dropout   # None
```

The two sibling classes both pass it in the equivalent branch — `char_level_bpe.py` and `sentencepiece_bpe.py`. #1136 made exactly this change to `char_level_bpe.py` in 2022 and did not touch this file.

The value survives training, so BPE-dropout goes from inert to working. Encoding one sentence 300 times after `train_from_iterator`:

```
before: 1 distinct tokenization
after: 36 distinct tokenizations
```

Anyone not passing `dropout` is unaffected: `BPE(dropout=None)` serializes identically to `BPE()`, which I checked for the default, `add_prefix_space`, `lowercase` and `trim_offsets` configurations.

I kept this to `dropout` alone. `continuing_subword_prefix` and `end_of_word_suffix` are dropped in the same branch, but mirroring the `if`-branch's `or ""` there would change the serialized `tokenizer.json` for every default user from `null` to `""`, and passing them through as `None` raises `TypeError`. That one seems worth handling separately, if at all.

Added two tests next to the existing ones: `dropout` reaching the model, and the default still being `None`.

### DIFF
--- a/bindings/python/py_src/tokenizers/implementations/byte_level_bpe.py
+++ b/bindings/python/py_src/tokenizers/implementations/byte_level_bpe.py
@@ -36,7 +36,7 @@ class ByteLevelBPETokenizer(BaseTokenizer):
                 )
             )
         else:
-            tokenizer = Tokenizer(BPE())
+            tokenizer = Tokenizer(BPE(dropout=dropout))
 
         # Check for Unicode normalization first (before everything else)
         normalizers = []
--- a/bindings/python/tests/implementations/test_byte_level_bpe.py
+++ b/bindings/python/tests/implementations/test_byte_level_bpe.py
@@ -101,3 +101,12 @@ class TestByteLevelBPE:
 
         output = tokenizer.encode("A sentence")
         assert output.tokens == ["A", "Ġsentence"]
+
+    def test_dropout_is_used_without_vocab(self):
+        # Without vocab/merges the tokenizer builds its own BPE, and `dropout`
+        # has to reach it the same way it does for the other BPE tokenizers.
+        tokenizer = ByteLevelBPETokenizer(dropout=0.1)
+        assert tokenizer._tokenizer.model.dropout == pytest.approx(0.1)
+
+    def test_dropout_defaults_to_none(self):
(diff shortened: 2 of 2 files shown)


ArthurZucker:
hey! ty, we are in the middle of a big refactoring, let's see after that!

VenishPaneliya:
Sounds good, thanks for the heads up — happy to leave this parked until the refactor lands.

Whenever it settles, just ping me: I'll rebase this on top if the convenience wrappers survive in some form, or close it out if that layer goes away entirely. Either outcome is fine by me, and I'd rather not add noise to the refactor.

Out of curiosity, is `PipelineTokenizer` intended to replace `ByteLevelBPETokenizer` and friends outright, or will they stick around as thin wrappers over it? Asking mostly because the same `dropout` plumbing would need to land wherever those constructors end up, and I'd be glad to help with that part once the shape is clear.

VenishPaneliya:
Following up after the refactor — #2391 removed the whole `implementations/` package, so `ByteLevelBPETokenizer` (and the file this patch touches) no longer exists on `main`. Nothing left to fix here, so I'll close this rather than leave a conflicting PR around.

If the convenience wrappers come back in some form, the underlying issue was that `ByteLevelBPETokenizer.__init__` accepted `dropout` and then never passed it to the `BPE` model, so it silently did nothing. Worth keeping in mind if that constructor gets reintroduced.

PR #2368, opened 2026-08-27
Add a `rusty-expressions` regex backend (Oniguruma semantics, pure Rust, wasm-capable)

Ttimmahlax:
## What

Adds a third optional `SysRegex` backend behind a `rusty-expressions` feature,
alongside the existing `onig` and `fancy-regex` ones.

[`rusty_expressions`](https://crates.io/crates/rusty_expressions) is Oniguruma
reimplemented in pure Rust. It aims at the same semantics as `onig` — it is
developed against live `libonig` as a differential oracle — while removing the
C from the build.

**Nothing changes by default.** `onig` remains the default backend; the new
feature is additive and off unless asked for. When enabled it takes precedence
over the other two.

## Why a third backend rather than a swap

The two existing options each give up something:

| | C toolchain | wasm32 | Oniguruma semantics |
|---|---|---|---|
| `onig` | required | ✗ | ✓ |
| `fancy-regex` | none | ✓ | different engine |
| **`rusty-expressions`** | **none** | **✓** | **✓** |

`unstable_wasm` exists precisely because `onig` cannot build for wasm32. That
workaround also changes the engine, which is presumably why it is not the
default. This backend is the combination that was missing.

## Verification on this branch

- **201 lib tests pass** with `--no-default-features --features rusty-expressions`.
- **Pre-tokenizer output is byte-identical to the `onig` backend.** I built the
  same dumper under both backends and diffed: 95 dumps across `ByteLevel`,
  `Whitespace`, `Punctuation`, a `\s+(?!\S)` negative-lookahead `Split`, and a
  `(\w)\1` backreference `Split`, over a corpus covering ASCII, CJK, Cyrillic,
  Greek, emoji with skin-tone modifiers, NFD vs NFC, zero-width characters,
  URLs, timestamps and long repeated strings. No differences.
- `wasm32-unknown-unknown` compiles with `rusty-expressions` (given the same
  `getrandom` configuration `unstable_wasm` already supplies) and fails with
  `onig`, as expected.

## Notes for review

- The dependency is taken with `default-features = false` deliberately:
  `rusty_expressions` installs `rusty_alloc` as a `#[global_allocator]` under
  its default feature, and a library must never impose an allocator on its
  consumers.
- `find_iter` advances by a whole character on an empty match, so it cannot
  land inside a UTF-8 sequence, and stops on an engine limit rather than
  propagating — matching the `fancy-regex` backend's behaviour.
- The engine is `unsafe_code = "deny"`; the only `unsafe` in that crate is an
  optional C-ABI shim which is not enabled here.

Happy to adjust naming, feature precedence, or add the backend to CI if you'd
like it covered there. I can also run a wider differential over real
tokenizer.json files from the Hub if that would help review.

### DIFF
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -73,6 +73,9 @@ harness = false
 [dependencies]
 rand = "0.9"
 onig = { version = "6.5.1", default-features = false, optional = true }
+# `default-features = false`: rusty_expressions installs a #[global_allocator]
+# under its default feature, and a library must never impose one on consumers.
+rusty_expressions = { version = "0.1", default-features = false, optional = true }
 regex = "1.10"
 regex-syntax = "0.8"
 rayon = "1.10"
@@ -108,6 +111,9 @@ esaxx_fast = ["esaxx-rs/cpp"]
 progressbar = ["indicatif"]
 http = ["hf-hub"]
 unstable_wasm = ["fancy-regex", "getrandom/wasm_js"]
+# Oniguruma remade in pure Rust: same engine semantics as `onig`, no C in the
+# build, works on wasm32, and measured faster than libonig.
+rusty-expressions = ["dep:rusty_expressions"]
 rustls-tls = ["hf-hub?/rustls-tls"]
 parity-aware-bpe = []
 
--- a/tokenizers/Cargo.toml
+++ b/tokenizers/Cargo.toml
@@ -117,6 +117,13 @@ wasm = ["getrandom/wasm_js"]
 unstable_wasm = ["fancy-regex", "wasm"]
 # Oniguruma remade in pure Rust: same engine semantics as `onig`, no C in the
 # build, works on wasm32, and measured faster than libonig.
+#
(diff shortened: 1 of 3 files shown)


Ttimmahlax:
Pushed a second commit that I think makes the first more useful on its own.

`unstable_wasm` currently means two things at once — "configure getrandom for
wasm32" **and** "use fancy-regex". A wasm build cannot ask for the plumbing
without also accepting that particular engine, which is exactly what I ran into
wiring this up downstream.

So this adds a `wasm` feature carrying only the getrandom configuration, and
redefines `unstable_wasm = ["fancy-regex", "wasm"]` — unchanged meaning for
anyone using it today. A wasm build can now choose its engine:

```
--features unstable_wasm            # fancy-regex, exactly as before
--features rusty-expressions,wasm   # Oniguruma semantics, still no C
```

That half stands alone and is useful even if you'd rather not take the new
backend. Happy to split it into a separate PR if you'd prefer to review them
independently.

ArthurZucker:
Hey! ty, we are working on tokenizers v1, see #2119

PR #2372, opened 2026-08-31
feat: support `first` prepend scheme for metaspace

SBrandeis:
# TL;DR

- Support `prepend: "first"` in Metaspace
- Support `split: false` in Metaspace

Updated Oracle tests


tokenizers/tk-encode/src/normalizers/bert.rs
```diff
@@ -128,7 +128,7 @@ impl BertNormalizer {
 }
 
 impl pipeline::Normalizer for BertNormalizer {
-    fn normalize<'a>(&self, input: &'a str) -> Result<Cow<'a, str>> {
+    fn normalize<'a>(&self, input: &'a str, _is_sequence_start: bool) -> Result<Cow<'a, str>> {
```
ArthurZucker, line 131: arf, feels like a loss to have to add this to all normalizers....

tokenizers/tk-encode/src/normalizers/utils.rs
```diff
@@ -17,7 +17,7 @@ pub(crate) fn lowercases_to_self(c: char) -> bool {
 }
 
 impl pipeline::Normalizer for Lowercase {
-    fn normalize<'a>(&self, input: &'a str) -> Result<Cow<'a, str>> {
+    fn normalize<'a>(&self, input: &'a str, _is_sequence_start: bool) -> Result<Cow<'a, str>> {
```
ArthurZucker, line 20: yeah, passing the offset? or cursor might be better no?

ArthurZucker (commented):
are you sure we have no other way to accommodate? 
Passing a cursor is prehemptive as normalizer will need them anyways to compute offsets, so would be more general / make more sense WDYT?

PR #2375, opened 2026-09-01
feat: experimental C bindings

SBrandeis:


.github/workflows/c.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+
+      # Only fixture the example needs (see examples/Makefile's FIXTURE variable).
+      - name: Download the gpt2 fixture
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+        run: uvx --from huggingface_hub hf download hf-internal-testing/tokenizers-test-data gpt2.json --repo-type dataset --revision e20ac5914f5ada06874e4f9d45e49b31a049904e --local-dir ../../tokenizers/data
+
+      # examples/Makefile's `test` target builds and runs encode.c and null_safety.c against
```
ArthurZucker: I don't understand much of this 😓 but yeah, we don't want to run too many test, just test the bindings (test we can encode, we can decode, we can load, but not test results in a whole bunch of stuff)

bindings/c/src/tokenizer.rs
```diff
@@ -0,0 +1,151 @@
...
+/// Instantiates a tokenizer from its JSON config file.
+///
+/// # Safety
+/// 1. `path` must be non-NULL, point to a NUL-terminated byte string, and be valid for reads
+///    up to and including that NUL byte for the duration of this call. It must not be mutated
+///    while this function runs.
+/// 2. `out` must be valid, writable pointer to a [`TkHandle<TkTokenizer>`]. On return, it
+///    holds a live handle if this function returns NULL, or NULL otherwise.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn tk_tokenizer_from_file(
```
ArthurZucker, line 21: from pretrained would be nice is well, but fine without

bindings/c/src/tokenizer.rs
ArthurZucker, line 1: my only comment is to align as much as possible with the name of the function we have for node when relevant. Same for the structure names and etc?

ArthurZucker (approved):
Overall looks good to me, not sure how much of an in depth review is needed, but decode is kind of a must, we can have streaming for a later use case. 

Parallel would be nice, not sure how hard it is to pull it off

bindings/c/src/tokenizer.rs
SBrandeis: sounds good!

PR #2365, opened 2026-08-25
perf(pipeline): give each thread its own scratch sub-pool

sebpop:
`ScratchPool` is one `Mutex<Vec<EncodeScratch>>` and every `encode` locks it
twice: once to pop a scratch, once to push it back when `ScratchGuard` drops.
Every encoding thread therefore serialises on one lock, which is what the note
in the source predicted:
```
    TODO @McPatate : The Mutex can create contention, to be replaced by a
    better access pattern
```
Measured with N threads encoding a 12 kB prompt of 30k distinct words through
one shared `PipelineTokenizer`, throughput stops improving between 16 and 32
threads and then falls back (encodes/s):
```
    threads   Intel GNR 128c   NVIDIA Vera 88c   AMD Turin 96c
          1           52,474            67,399          56,076
          8          205,768           308,619         408,580
         16          298,086           327,975         640,830
         32          226,496           609,208         675,155
         64          191,749           475,348         668,325
       full          223,578           314,715         661,471
```
Scaling efficiency at full width is 3.3%, 5.3% and 12.3% of linear.

This gives each thread one of 64 sub-pools, picked once per thread from a
counter and kept in a thread-local. A scratch returns to the sub-pool it came
from, so a thread reuses its own buffers and the `WordCache` inside each scratch
stays warm per thread instead of being passed around. A thread whose own
sub-pool is empty takes from another before building -- `try_lock`, so it skips
a busy sub-pool rather than queueing behind it, which would reintroduce what the
sharding removes. That keeps the property the pool exists for: it hands out what
it has rather than allocating alongside it.
`keeps_at_most_one_scratch_per_concurrent_holder` is what makes this non-optional;
without the borrow it fails.

Same machines, same inputs, with the sharded pool:
```
    threads   Intel GNR 128c   NVIDIA Vera 88c   AMD Turin 96c
          1           52,557            67,927          56,054
          8          401,004           538,663         438,338
         16          798,554         1,074,139         871,056
         32        1,568,314         2,135,057       1,698,626
         64        2,858,405         4,265,264       2,832,353
       full        4,623,304         4,904,611       3,719,211
```
20.7x, 15.6x and 5.6x at full width, at 69%, 82% and 69% efficiency. Single
thread is unchanged on all three (1.00x), so the serial path pays nothing.

Measured with the `scratch_scaling` example added in this series, 5 s per point,
on otherwise idle machines. `cargo test -p tk-encode` goes from 131 to 132
tests passing, the new one being the borrow path above.

[...]

PR #2378, opened 2026-09-02
perf(pipeline): partition plan_work's chunks instead of sorting them

ArthurZucker:
`plan_work` sorted the chunk table by descending length in the serial prologue —
O(n log n) moves of a 40-byte struct on the calling thread, where n is the batch size
when each sequence is one chunk. A single partition pass is O(n) and keeps the only
property the scheduler needs. **+2.6% to +16.9%** on 80k-document batches, ids identical.

### Relationship to #2361

**#2361 already contains this**, reached independently, and its eight-line version is
better than the counting sort I first wrote — so this PR now uses that shape. Whichever
lands first, the other should drop the hunk.

The reason to keep this one separate: #2361 bundles three changes, and #2365 undercuts
two of them. `encode_batch_flat` (+222 in `mod.rs`, +167 in `parallelism.rs`) and the
concat-to-pool machinery exist so workers share nothing — but #2365 shows the shared
`ScratchPool` mutex *was* the contention, taking scaling efficiency from 3.3–12.3% to
69–82%. #2361's "3.4× over the general path" was measured against a baseline whose
contention #2365 removes, so that comparison needs redoing before the flat path earns
its keep.

This PR is the planning fix alone: one file, +88/−2.

### The change

```rust
let mut boundary = 0;
for i in 0..chunks.len() {
    if chunks[i].range.len() >= PARALLEL_MIN_BYTES {
        chunks.swap(boundary, i);
        boundary += 1;
    }
}
```

Tasks are already byte-balanced — each grown until it holds `PARALLEL_MIN_BYTES` — so how
the sub-threshold chunks sit among themselves cannot change a task's size. Only one thing
matters: a chunk big enough to be a task on its own must not be picked last and become the
straggler the batch waits on.

### Measured

Batched `encode`, 64 B documents, corpus repeated to fill 250 ms per pass (borrowed from
`measure_scaling`, for its reason — one call is a few ms and spawns up to
`current_num_threads()` rayon tasks), min over passes, 3 interleaved rounds, separately
built binaries.

| cell | n_docs | base MB/s | partition |
|---|---|---|---|
| gpt2/chinese | 80457 | 361.5 | **+16.9%** |
| llama-3/chinese | 80457 | 337.7 | **+12.9%** |
| llama-3/code | 32691 | 701.3 | +6.3% |
| llama-3/english | 81983 | 651.4 | +5.4% |
| gpt2/code | 32691 | 675.9 | +3.0% |
| gpt2/english | 81983 | 668.3 | +2.6% |

Gone by 1 kB documents, which is the shape an O(n log n) → O(n) change in a serial prefix
should have. `ids_hash` identical to base in every cell.

The middle commit is the counting-sort version and the last one removes it; the history is
kept so the comparison is on record, and it squashes to the eight-line form.

### The other commit

The single-sequence early-out has **no measurable effect** and its message says so:
`plan_sequence` short-circuits before its added-token scan, so the only waste in the
8–16 KiB window is ~6 small allocations. Measured at 4 kB / 9 kB / 20 kB single documents
(first and last as controls, where the guard cannot fire): every cell within ±1.5%. Kept
only because it removes provably-discarded work. Drop it if you would rather have the
smaller diff.

### Note on measuring this

`Engine::encode` in tokbench takes a single `&str`, so `plan_work` there only ever sees one
chunk, fails `tasks.len() < 2`, and falls back to `encode_serial` — the planner is
unreachable through that trait, and `--scaling` runs N independent single-document engines.
These numbers come from a batched harness. A batch method on `Engine` with a default impl
looping the single-document one would make this shape measurable in tokbench proper.


tokenizers/tk-encode/src/tokenizer/pipeline/parallel.rs
```diff
@@ -359,8 +359,23 @@ impl PipelineTokenizer {
...
+        // Scheduling wants the big chunks taken first, so one cannot be picked last and become the
+        // straggler the whole batch waits on. It does *not* need a total order: the tasks below are
+        // already byte-balanced, each grown until it holds `PARALLEL_MIN_BYTES`, so how the ones
+        // below that threshold are arranged among themselves cannot change a task's size.
+        //
+        // Sorting every chunk to get that was O(n log n) of 40-byte moves on the calling thread,
+        // before any worker starts -- and n is the batch size when each sequence is one chunk, so a
+        // batch of short documents sorted tens of thousands of structs to answer a question that
+        // only needs two classes. Partitioning the chunks that are a task by themselves to the
+        // front is O(n) and keeps the property that matters.
```
McPatate: this is a lot of comment we don't need 😄 
```suggestion
         // make sure larger chunks are first in the batch to avoid having a straggler at the end
```

tokenizers/tk-encode/src/tokenizer/pipeline/parallel.rs
```diff
@@ -400,6 +415,19 @@ pub(crate) fn encode(
...
         return EncodeHandle::blocking(tok.encode_serial(inputs, add_special_tokens));
     }
+    // One sequence under `2 * PARALLEL_MIN_BYTES` is emitted by `plan_sequence` as a single chunk,
+    // which becomes a single task, which fails the `tasks.len() < 2` check below -- but only after
+    // `plan_work` has allocated `chunks`, `outputs`, `side_a_len`, `chunk_count` and `tasks`, and
+    // thrown all five away. Every `encode` of a lone document between `PARALLEL_MIN_BYTES` and
+    // twice it paid for that. Decide it before paying.
+    //
+    // `Input::Single` only: a pair is two `plan_sequence` calls, so it can reach two tasks and has
+    // to go through planning to find out.
```
McPatate:
```suggestion
```

tokenizers/tk-encode/src/tokenizer/pipeline/parallel.rs
```diff
@@ -442,3 +470,61 @@ pub(crate) fn encode(
     }
     EncodeHandle::streaming(StreamingIter::new(batch))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// The partition has to leave every chunk that is a task by itself ahead of every chunk that is
+    /// not, and it has to keep all of them -- it reorders, it never drops.
```
McPatate: either reword this or delete pls, it's not very clear

tokenizers/tk-encode/src/tokenizer/pipeline/parallel.rs
```diff
@@ -442,3 +453,61 @@ pub(crate) fn encode(
...
+            "a whole-task chunk got left in the tail"
+        );
+
+        let mut got: Vec<usize> = chunks.iter().map(|c| c.range.len()).collect();
+        let mut want = lens.to_vec();
+        got.sort_unstable();
+        want.sort_unstable();
+        assert_eq!(got, want, "the partition lost or invented a chunk");
+    }
+}
```
ArthurZucker:
```suggestion
```

PR #2387, opened 2026-09-03
perf(decode): index the vocab slab by id, not through the MPHF slot

ArthurZucker:
Closes the decode gap against `tokie`, which was the fastest decoder in [tokbench](https://github.com/huggingface/tokbench) and 1.19× ahead of us on english.

## What was slow

`decode_byte_level` reached every token through `id_to_token_bytes`:

```
id_to_slot[id]  →  spans[slot]  →  bytes[start .. start+len]
```

Three dependent loads, and the middle hop is a **random** access — slot order is the MPHF's, not id order. `cargo asm` on the loop shows that chase plus three bounds checks per token, and the vocabulary base pointers spilled to the stack and reloaded every iteration. A flat-arena decoder does one load of an adjacent offset pair and a copy.

## What changed

`build` now lays the slab out in ascending id order and records `decode_off`, so `bytes[decode_off[id] .. decode_off[id + 1]]` is the token. The two offsets are adjacent, usually in the same cache line.

The reorder is free: nothing below the placement loop depends on iteration order — `entries`, `spans` and `id_to_slot` are written by slot or by id, and `keys` only feeds an order-independent `HashSet`. An id the vocabulary does not hold gets equal offsets, i.e. an empty slice, which is exactly what a decoder appends for it.

**Cost: 4 bytes per id in the id space** — ~513 kB on a 128k-id vocabulary — and no duplicated token bytes.

## Measured

`tk-serialize`'s decode bench, llama-3, 10 kB chunks, criterion, p = 0.00 on both:

| corpus | before | after | |
|---|---:|---:|---:|
| english | 367.4 | **428.0 MiB/s** | +16.5% |
| japanese | 449.2 | **490.1 MiB/s** | +9.1% |

And in tokbench, which hashes every engine's decoded text and compares it against the released crate — **all cells byte-exact**:

| corpus | before | after | tokie | was | now |
|---|---:|---:|---:|---:|---:|
| english | 380.8 | **464.4** | 466.3 | 0.841× | **0.996×** |
| chinese | 353.3 | **401.3** | 386.0 | 0.924× | **1.04×** |
| japanese | 390.0 | **442.6** | 425.2 | 0.954× | **1.04×** |

## The bench is new, and that matters

The existing decode benches call `decode` once per **line**. big.txt lines are short, so one allocation and one full `from_utf8` validation per line dominate, and nothing happening in the token loop is visible. Concretely: a change that cut the loop from 420 to 348 instructions moved those benches not at all.

The added `decode, 10 kB chunks` bench uses tokbench's chunking, so it is both loop-sensitive and directly comparable with that matrix.

## Two things that did not work

Recorded so nobody burns the time again:

1. **Raising the output `Vec` estimate from `ids.len() * 4` to `* 5`.** English output is 4.55 B/token so the 4× guess does under-allocate — but english came out flat (p = 0.12) and **japanese regressed 4.9%**. Over-reserving 26% costs more than the grow it avoids.
2. **Outlining the added-token arm as `#[inline(never)] #[cold]`.** It worked structurally — confirmed in asm, the base pointers moved out of the stack into registers and the loop went 420 → 348 instructions, 37 → 30 calls — and the clock did not move. Those spills were L1 hits.

## Tests

`decode_index_matches_the_pointer_chase` asserts the new index agrees with the chase it replaces for **every id in the id space**, on a deliberately sparse vocabulary. Sparse ids are the real risk: a hole is encoded as `decode_off[id] == decode_off[id + 1]`, and an off-by-one in that walk would hand a hole its neighbour's bytes — a silently wrong decode, not a crash.

`cargo test -p tk-encode -p tk-serialize` is green (176 + 38 + 2 + 2). clippy and `fmt --check` clean. The now-dead `PipelineBPE::id_to_token_bytes` passthrough is removed; the store's own `id_to_token_bytes` stays, still used by serialization and `id_to_token`.

### DIFF
--- a/tokenizers/tk-encode/src/models/bpe/model.rs
+++ b/tokenizers/tk-encode/src/models/bpe/model.rs
@@ -82,13 +82,17 @@ impl PipelineBPE {
     /// A token's bytes, borrowed from the vocab store's slab. For a byte-level model these are
     /// the decoded bytes (see [`Self::is_byte_level`]) and a single entry is not necessarily
     /// valid UTF-8 on its own -- only the concatenation of a whole id sequence usually is.
-    pub(crate) fn id_to_token_bytes(&self, id: u32) -> Option<&[u8]> {
-        self.vocab.id_to_token_bytes(id)
+    /// The decode-path lookup: one adjacent-pair load instead of the `id_to_slot` -> `spans` ->
+    /// `bytes` chase. Empty slice for an id the vocabulary does not hold, which is what a decoder
+    /// appends for it anyway. See `BucketVocabStore::id_to_token_bytes_for_decode`.
+    #[inline]
+    pub(crate) fn id_to_token_bytes_for_decode(&self, id: u32) -> &[u8] {
+        self.vocab.id_to_token_bytes_for_decode(id)
     }
 
     /// A token as a `String`, for the decoder-chain route. Only meaningful when the entries are
     /// the token strings as written, i.e. when [`Self::is_byte_level`] is false; a byte-level
-    /// model decodes through [`Self::id_to_token_bytes`] instead.
+    /// model decodes through [`Self::id_to_token_bytes_for_decode`] instead.
     pub(crate) fn id_to_token(&self, id: u32) -> Option<String> {
         self.vocab.id_to_token(id)
     }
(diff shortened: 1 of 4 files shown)


ArthurZucker (commented):
Will make this again, its too convoluted

PR #2391, opened 2026-09-03
feat: rework python bindings

SBrandeis:
# TL;DR

Rework of the Python bindings to make them slimmer (minimalistic API)

Some caveats:
- Unclear whether this runs properly on free-threaded python


lhoestq:
I could confirm that pickling, multiprocessing and pickle-dump-hash determinism are all good wrt `datasets`.

I can try e2e with `transformers` tokenizers once ready !

bindings/python/python/tokenizers/tokenizers.pyi
```diff
@@ -0,0 +1,211 @@
...
+            texts: List[str]
+                The batch of text to encode.
+            add_special_tokens: bool
+                Whether the post-processor adds its special tokens, such as `[CLS]` and `[SEP]`.
+
+        Returns:
+            List[Encoding]
+        """
+    @staticmethod
+    def from_file(path: str | PathLike[str], padding: Padding | None = None) -> Tokenizer:
```
ArthurZucker, line 195: `from_pretrained` is a must!

bindings/python/src/tokenizer.rs
```diff
@@ -1,2020 +1,175 @@
...
-    pub lstrip: Option<bool>,
-    pub rstrip: Option<bool>,
-    pub normalized: Option<bool>,
+/// A tokenizer. Encodes text into token ids, and decodes token ids back into text.
+#[pyclass(frozen, module = "tokenizers")]
+pub struct Tokenizer {
+    pipeline: Pipeline,
+    // Needs a mutex so a concurrent thread can access the value while
+    // encode is running.
+    padding: Mutex<Option<PaddingParams>>,
```
ArthurZucker, line 20: fine, its also fine if its just copied across threads tbf

bindings/python/tests/test_tokenizer.py
```diff
@@ -0,0 +1,457 @@
...
+
+    assert encoding.ids_array.tolist() == encoding.ids
+    assert encoding.type_ids_array.tolist() == encoding.type_ids
+    assert encoding.attention_mask_array.tolist() == encoding.attention_mask
+
+
+def test_encodings_with_the_same_fields_are_equal():
+    tokenizer = Tokenizer.from_file(BERT)
+
+    assert tokenizer.encode("Hello") == tokenizer.encode("Hello")
```
ArthurZucker: this is a very weird

bindings/python/tests/test_tokenizer.py
ArthurZucker, line 1: the test IMO need a bit more work! they are way to unitest like

bindings/python/tools/stub-gen/Cargo.toml
```diff
@@ -2,13 +2,8 @@
 name = "stub-gen"
 version = "0.1.0"
 edition = "2024"
-description = "Stub generation tool for tokenizers Python bindings"
-
-[[bin]]
-name = "stub-gen"
-path = "src/main.rs"
+publish = false
+description = "Generates the tokenizers module's .pyi stub from the built cdylib"
```
ArthurZucker, line 10: can we leave that please? should not go away

bindings/python/Makefile
```diff
@@ -1,67 +1,36 @@
...
-	ruff format --check $(check_dirs)
-	ty check py_src --exclude py_src/tokenizers/implementations --exclude py_src/tokenizers/tools/visualizer.py
-
+# Debug build of the extension, installed editable into the active Python environment.
+develop: stubs
+	maturin develop
 
-TESTS_RESOURCES = $(DATA_DIR)/small.txt $(DATA_DIR)/roberta.json
+stubs:
+	cargo run --manifest-path ./tools/stub-gen/Cargo.toml
```
ArthurZucker, line 13: ah okay fine as long as its still ran

bindings/python/README.md
```diff
@@ -1,252 +1,80 @@
...
-tokenizer.train([ "./path/to/files/1.txt", "./path/to/files/2.txt" ])
-
-# Now, let's use it:
-encoded = tokenizer.encode("I can feel the magic, can you?")
-
-# And finally save it somewhere
-tokenizer.save("./path/to/directory/my-bpe.tokenizer.json")
+from tokenizers import Padding, Tokenizer
+
+tokenizer = Tokenizer.from_file("tokenizer.json")
```
ArthurZucker, line 18: needs from pretrained + a big disclaimer that this is v1 only!

bindings/python/README.md
```diff
@@ -1,252 +1,80 @@
...
+encoding = tokenizer.encode("Hello there, how are you?")
+encoding.ids             # [101, 7592, 2045, 1010, 2129, 2024, 2017, 1029, 102]
+encoding.type_ids        # [0, 0, 0, 0, 0, 0, 0, 0, 0]
+encoding.attention_mask  # [1, 1, 1, 1, 1, 1, 1, 1, 1]
+
+# The same fields as numpy arrays: read-only views over the encoding, not copies. Reading one
+# costs nothing, and an array keeps its encoding alive for as long as the array exists.
+encoding.ids_array             # array([101, 7592, 2045, 1010, 2129, 2024, 2017, 1029, 102], dtype=uint32)
+encoding.type_ids_array        # array([0, 0, 0, 0, 0, 0, 0, 0, 0], dtype=uint32)
+encoding.attention_mask_array  # array([1, 1, 1, 1, 1, 1, 1, 1, 1], dtype=uint32)
```
ArthurZucker, line 32: nice! I like this solution

bindings/python/README.md
```diff
@@ -1,252 +1,80 @@
...
-```python
-from tokenizers import Tokenizer
+def encode(tokenizer, texts):
+    return tokenizer.encode_batch(texts)
 
-tokenizer = Tokenizer.from_file("byte-level-bpe.tokenizer.json")
 
-encoded = tokenizer.encode("I can feel the magic, can you?")
+with multiprocessing.get_context("spawn").Pool(4) as pool:
+    encoded = pool.starmap(encode, [(tokenizer, chunk) for chunk in chunks])
```
ArthurZucker, line 68: we need to check, potentially disabling the parallelism automatically when in subprocess -> user wants to handle it himself

ArthurZucker (approved):

SBrandeis:
Hey @ArthurZucker I addressed some of your concerns:
- verbose tests
- from_pretrained
- stub_gen built as a runnable binary

PTAL at those changes 🤗

bindings/python/.cargo/config.toml
ArthurZucker, line 1: Mmm interesting that this is no longer needed

bindings/python/python/tokenizers/__init__.py
ArthurZucker, line 1: Why are we not exposing https://github.com/huggingface/tokenizers/blob/feat/rework-python-bindings/tokenizers/tk-encode/src/utils/from_pretrained.rs#L26-L26 ?

bindings/python/python/tokenizers/tokenizers.pyi
```diff
@@ -0,0 +1,276 @@
+from collections.abc import Sequence
```
ArthurZucker, line 1: let's make sure to add a header that hits is auto generated no?

bindings/python/tools/stub-gen/python_members.py
ArthurZucker, line 1: I'd like to avoid that IMO and keep stuff in rust to always auto generate!

bindings/python/tools/stub-gen/src/main.rs
ArthurZucker, line 1: 😢 hahah all the efforts I did to get this right I am sadddd

bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
ArthurZucker, line 1: Let's maybe show an example of how transformers does this instead since that's where it lives!

bindings/python/tools/stub-gen/python_members.py
SBrandeis: we'll do as a follow-up

bindings/python/python/tokenizers/tokenizers.pyi
```diff
@@ -0,0 +1,278 @@
+# Generated by tools/stub-gen from the built extension module. Do not edit by hand.
+# Regenerate with `make stubs` in bindings/python.
+
+from collections.abc import Sequence
+from os import PathLike
+from typing import Any, Final, Literal, final
+
+from numpy import integer, uint32
+from numpy.typing import NDArray
```
ArthurZucker, line 9: mmm numpy should not be a hard dep!!!! not sure how easy that's to do 😐

ArthurZucker (commented):
One thing that we want to keep API wise are the getters like: 
```rust
    #[getter]
    fn get_ids(&self) -> Vec<u32> {
        self.encoding.get_ids().to_vec()
    }
``` 
and then in python `tok.encode(["hey", "heyyyyyy"]).ids`. 

Also, if we don't need anything else than ids. we could also have a "id_only" -> no allocation for attention mask, no encoding, no type ids -> could be faster!


For the __repr__ i'd say let's have `\n` ! 
Interestingly we can't do `tokenize` anymore! That's something we'll add back in v1, but it can be as easy as just taking the vocab-decoded id (no decoder). I'll add it IMO it can be nice. 

Someting missing in decode is to take list of list. Only supports numpy list, but not batch. SO either decode batch, or just support both. 

- `role_to_token` needs to get exposed! probably just overwrite the getter !

Image
this is also fine!

PR #2399, opened 2026-09-10
chore(post-process): remove polymorphism + encode_into

SBrandeis:


tokenizers/tk-serialize/benches/encode_shapes.rs
ArthurZucker, line 1: a bit weird to have that single bench in tk-serialize,  can you explain a bit the motivation?

ArthurZucker (approved):
SG

PR #2400, opened 2026-09-10
feat(python): add Arrow string batch tokenization

granthamtaylor:
Arrow callers currently need to convert arrays to Python lists before calling `encode_batch`. This adds `Tokenizer.encode_batch_arrow`, which accepts Arrow `string` and `large_string` arrays through `__arrow_c_array__` and returns the existing pipeline `Encoding` objects.

Depends on #2119 and targets `feat/train_encode_split`. The diff contains one commit confined to the Python bindings.

### Behavior

- Supports sliced and empty arrays, special tokens, and the existing padding configuration and per-call overrides.
- Nulls raise an error by default. `null_handling="empty"` encodes them as empty strings; `"skip"` omits them while preserving retained row order.
- Validates string buffers and capsule ownership, with release callbacks running after reattaching to Python.
- Avoids materializing Python strings. The existing pipeline still copies text into Rust-owned inputs.

Includes generated type hints, parity and ownership tests, and a standalone benchmark. PyArrow is a development dependency only. Chunked arrays and non-string Arrow types are outside the method's scope.

### Validation

On macOS ARM64 with Python 3.13:

- Local Python suite: **211 passed, 1 skipped**, with 1 network test deselected.
- Rust formatting and Clippy, Ruff, and `ty check python tests tools`: passed.
- Release-build benchmark smoke test: passed for both Arrow string widths.

The full example type check still reports errors in the unchanged upstream `examples/train_parity_bpe.py`, which uses the previous binding API.

Related: #1415.


ArthurZucker:
Happy to have a look once rebased on #2119

granthamtaylor:
@ArthurZucker thank you so much !

I have rebased.

ArthurZucker:
Ty! this won't make it to rc0 but we'll se after that!

granthamtaylor:
I understand!

Please feel free to tag me with any next steps.

I'd love to see this feature integrated for my current project in which I am trying to tokenize many, many small strings.

Thank you for your consideration here.

PR #2402, opened 2026-09-11
chore(cleanup): wrap encode options in a struct

SBrandeis:


tokenizers/tk-encode/src/tokenizer/pipeline/encode_options.rs
```diff
@@ -0,0 +1,37 @@
+use crate::PaddingParams;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EncodeOptions {
+    pub add_special_tokens: bool,
+    pub padding: Override<PaddingParams>,
```
ArthurZucker, line 6: encode special tokens as well!

PR #2413, opened 2026-09-16
Fix custom prefixes in BertWordPieceTokenizer

### LINKED ISSUE #2412: BertWordPieceTokenizer ignores wordpieces_prefix when loading a vocabulary
`BertWordPieceTokenizer(..., wordpieces_prefix="@@")` configures the decoder with `@@`, but the underlying WordPiece model keeps its default `##` prefix. A vocabulary containing custom-prefixed subwords therefore encodes known text as `[UNK]`.

```python
from tokenizers import BertWordPieceTokenizer

tokenizer = BertWordPieceTokenizer(
    vocab={"[UNK]": 0, "[CLS]": 1, "[SEP]": 2, "h": 3, "@@ello": 4},
    wordpieces_prefix="@@",
)
output = tokenizer.encode("hello", add_special_tokens=False)
print(output.tokens)           # Actual: ['[UNK]']
print(tokenizer.decode([3, 4])) # 'hello'
assert output.tokens == ["h", "@@ello"]
```

Expected: encoding and decoding use the same prefix, yielding `['h', '@@ello']` for `hello`.

This also breaks the training/save/reload workflow: training with `wordpieces_prefix="@@"` works, but loading the saved vocabulary with `BertWordPieceTokenizer.from_file(..., wordpieces_prefix="@@")` changes the same input to `[UNK]`.

Both `WordPiece(...)` constructor branches in `bert_wordpiece.py` omit `continuing_subword_prefix`. Forwarding `wordpieces_prefix` fixes dictionary construction and vocabulary-file reloads while preserving the default `##` behavior.

Reproduced with the released `tokenizers==0.23.2` package and with the current `main` Python wrappers loaded over that real native extension. Python 3.13.15, macOS ARM64; no model downloads or Rust rebuild. I checked existing issues and open PRs and did not find the same parameter-forwarding fix.

HaokaiDing:
Fixes #2412.

`BertWordPieceTokenizer` passes `wordpieces_prefix` to its decoder but leaves the WordPiece model using `##`. Loading a vocabulary trained with a custom prefix such as `@@` consequently turns known words into `[UNK]`.

Forward the prefix to both WordPiece constructor branches. Add tests for default/custom prefixes that check tokens, IDs, offsets and decoding, plus a training/save/reload round trip.

Validation:

- Before the fix: 2 failed, 3 passed. After: 5 passed, 2 network tests deselected in `tests/implementations/test_bert_wordpiece.py`.
- Tested the current checkout's Python package with the released `tokenizers 0.23.2` native extension; the local runner verifies the wrapper's source path and native module identity. Current Rust was not rebuilt and network-dependent tests were not run.
- Ruff lint and format checks pass on both changed files.
- Python 3.13.15 on macOS ARM64. No dependency changes.


HaokaiDing:
The conflicts come from #2391 removing `BertWordPieceTokenizer` and its implementation tests from `main`. Keeping those files during a merge would restore an API that the rewrite removed.

I prepared the same fix against `v0.23-release` and tested its Python source with the matching 0.23.1 wheel's native extension: 5 offline tests pass, including the two custom-prefix regressions that fail without the fix. The two network tests were deselected; a full Rust source build has not been run.

Since the 0.23.2 release notes call it the last v0 release, would you accept this as a backport, and which base should this PR target? I'll keep the current base unchanged until the intended target is clear.

PR #2417, opened 2026-09-17
feat: Implement tokenize() / decode_tokens()

SBrandeis:


bindings/python/src/tokenizer.rs
```diff
@@ -185,6 +185,36 @@ impl Tokenizer {
         Ok(Encoding::from(&encodings[0]))
     }
 
+    /// Encodes the given text and returns the string representation of each token.
+    ///
```
ArthurZucker, line 189: Let's make it return both the ids maybe? so they are not "thrown" away?

tokenizers/tk-encode/src/tokenizer/pipeline/mod.rs
```diff
@@ -620,6 +620,26 @@ impl PipelineTokenizer {
         }
     }
 
+    /// Decode a slice of token ids to a Vec of the tokens' string representation
```
ArthurZucker, line 623: perfect

tokenizers/tk-encode/src/models/bpe/model.rs
```diff
@@ -86,11 +87,18 @@ impl PipelineBPE {
...
+        if !self.is_byte_level() {
+            return self.vocab.id_to_token(id);
+        }
+        let bytes = self.id_to_token_bytes(id)?;
+        Some(
+            bytes
+                .iter()
+                .map(|&b| BYTES_CHAR_LOOKUP[b as usize])
+                .collect(),
+        )
```
ArthurZucker, line 101: MMM okay, you're making sure they are printable I guess. 
K, makes sense.

bindings/python/src/tokenizer.rs
```diff
@@ -238,6 +268,27 @@ impl Tokenizer {
             .map_err(err)
     }
 
+    /// Converts token ids to their string representation.
```
ArthurZucker:
```suggestion
    /// Converts token ids to their string representation. This does NOT apply the decoder.
```

ArthurZucker (commented):
A bit too many unitary tests! Python should not test the rust layer, just the python layer.
Test should be quite minal please! 

juste making sure the outputs are tokens, not string!

tokenizers/tk-encode/src/models/bpe/model.rs
SBrandeis: Matches the legacy .tokenize()

bindings/python/src/tokenizer.rs
SBrandeis: See [46e0ade](https://github.com/huggingface/tokenizers/pull/2417/commits/46e0ade7b3fe917764f701a7f302a6c48dc00da9) for an alternative

ArthurZucker (approved):

PR #2430, opened 2026-09-20
fix(tk-encode): longest-prefix matching in PrecompiledNormalizer

### LINKED ISSUE #2116: Precompiled normalizer uses grapheme-level shortest-prefix matching and diverges from SentencePiece on NFD input
The `Precompiled` normalizer's `normalize` walks the input grapheme by grapheme and (1) only does a whole-grapheme charsmap lookup for graphemes under 6 bytes, falling back to per-codepoint otherwise, and (2) calls `Precompiled::transform`, which returns the *shortest* prefix match. Native SentencePiece (`NormalizePrefix`) instead does a byte-stream *longest*-prefix match, with no grapheme split and no length cutoff. Both differences make the fast tokenizer's normalization diverge from the reference on NFD input, on the same `precompiled_charsmap`.

### Reproduction
```python
import unicodedata, sentencepiece as spm

mp   = AutoTokenizer.from_pretrained("xlm-roberta-base", use_fast=False).vocab_file
sp   = spm.SentencePieceProcessor(model_file=mp)            # reference
fast = AutoTokenizer.from_pretrained("xlm-roberta-base")    # tokenizers

def sp_norm_chars(s):
    return sp.normalize(s).removeprefix("▁")   # drop the dummy-prefix marker to compare the same stage

# (1) >=6-byte NFD grapheme: NFD Hangul 각 (U+1100 U+1161 U+11A8, 9 bytes)
h = unicodedata.normalize("NFD", "각")
print(sp_norm_chars(h), "|", fast.backend_tokenizer.normalizer.normalize_str(h))
# 각 | 각   -- same glyph, but the right side is still decomposed jamo:
print(sp.encode_as_pieces(h), "|", fast.tokenize(h))
# ['▁각'] | ['▁', 'ᄀ', 'ᅡ', 'ᆨ']

# (2) <6-byte NFD grapheme: NFD Vietnamese ắ (U+0061 U+0306 U+0301, 5 bytes)
v = unicodedata.normalize("NFD", "ắ")
print(sp_norm_chars(v), "|", fast.backend_tokenizer.normalizer.normalize_str(v))
# ắ | ă   -- combining acute dropped
print(sp.encode_as_pieces(v), "|", fast.tokenize(v))
# ['▁', 'ắ'] | ['▁', 'ă']
```
Case (1) is the 6-byte grapheme cutoff: the whole-grapheme rule is skipped, so the syllable stays decomposed. Case (2) is *under* the cutoff and still fails, purely because `transform` returns the shortest prefix (`ă`) instead of the longest (`ắ`).

### Root cause
`tokenizers/src/normalizers/precompiled.rs`, `Normalizer::normalize`: a grapheme loop with `if grapheme.len() < 6`, then a per-codepoint fallback, both calling `self.transform`. `spm_precompiled::transform` returns `results[0]` from `common_prefix_search`, i.e. the shortest match. SentencePiece `normalizer.cc` `NormalizePrefix` selects the longest trie result and has no grapheme split or length cutoff. `SpmConverter` copies the charsmap verbatim, so the two should normalize identically, but the grapheme cutoff and the shortest-prefix match diverge. The in-source comment points reviewers at XNLI for parity, but XNLI-style parity checks can miss this because they don't include explicit decomposed NFD cases.

### Impact
Fast tokenizers whose normalizer includes `Precompiled`: XLM-R, mBART-50, multilingual-e5, NLLB, BAAI/bge-m3, bge-reranker-v2-m3. (Verified the `Precompiled` normalizer and the divergence on `xlm-roberta-base` and `intfloat/multilingual-e5-small`. `microsoft/mdeberta-v3-base` is **not** affected — its normalizer is `Replace + NFC + Strip`, and NFD Hangul composes there.) NFD input is not exotic: it can arise from legacy HFS+ filenames, copied text, decomposed Unicode pipelines, or cross-platform file/data interchange — the same string then tokenizes differently depending on its normalization form.

### Direction
A byte-position longest-prefix loop mirroring `NormalizePrefix`. That needs a "longest match + consumed bytes" lookup; `spm_precompiled` currently exposes only `transform` (shortest) and `common_prefix_search` (values, no consumed length). The `tokenizers` manifest declares `spm_precompiled = "0.1.3"` and the latest release is `0.1.4`, but neither adds such an API — so a clean fix touches the `spm_precompiled` crate plus the offset handling in `tokenizers`. I have a prototype byte-position implementation that restores NFD↔NFC parity on the XLM-R charsmap with NFC unchanged. I can open a PR once the approach and the `spm_precompiled` dependency are agreed.

SIDDARTHAREDDY8:
Fixes #2116.

## Problem

`PrecompiledNormalizer` (the normalizer the pipeline actually runs for `Precompiled` configs) normalized through `spm_precompiled`'s `transform()`, which does **shortest**-prefix matching, wrapped in a grapheme walk with two more defects:

1. When a rule key is a proper prefix of a longer matching rule, the shorter rule won (SentencePiece takes the longest match).
2. The whole grapheme was replaced by that short match, silently dropping the bytes the match did not consume (e.g. NFD `ắ` = `a` + U+0306 + U+0301 normalized to `ă`, losing the combining acute).
3. Graphemes of 6+ bytes skipped whole-grapheme lookup entirely and were shredded per codepoint.

## Fix

`PrecompiledNormalizer` now parses the `precompiled_charsmap` once (the same layout `spm_precompiled` parses internally: `u32` LE trie length, `u32` LE double-array units, NUL-separated replacements) into a small `CharsmapMatcher`, and normalizes with SentencePiece-compatible **leftmost-longest** prefix matching over the byte stream: at each position the longest matching rule wins, exactly its key bytes are consumed, and unmatched positions are copied through unchanged.

Notes:
- `spm_precompiled` keeps its parse private and only exposes the shortest-prefix lookup, so the trie walk (including its exact `ArrayUnit` bit layout) is replicated in `tk-encode` rather than reused.
- The `impl Normalizer for Precompiled` (bare value) intentionally keeps its historical behavior with a NOTE comment: the trie is not visible through `spm_precompiled`'s API, so that impl cannot do longest-prefix matching. The pipeline never calls it anymore.
- No public API changes: `from_charsmap`, `charsmap()`, `parsed()` are untouched.

## Testing

- `cargo test -p tk-encode --features normalizers`: 200 passed, 0 failed — including 4 new regression tests in `normalizers::precompiled::tests` built on a minimal hand-rolled charsmap blob (longest rule wins over a shorter prefix, unmatched tail copied through, 7-byte grapheme rule applies, untouched input stays borrowed).
- `cargo test -p tk-serialize --lib`: 40 passed, 0 failed. (The `fold_parity` integration tests fail in this environment because `data/gpt2.json` is a fetched fixture not present here — unrelated to this change.)
- `cargo clippy -p tk-encode --features normalizers --all-targets`: no warnings; `rustfmt --check`: clean.
- Differential check against the real 237 KB `nmt_nfkc` charsmap (from `spm_precompiled`'s own test suite): on inputs where no proper-prefix rule pair fires (`"hello world"`, `"H\u00e9llo"`, Japanese text, …) the new output is byte-identical to the old; on the issue's NFD `ắ` input the old code produced `ă` (acute dropped) and the new code produces `ắ` (U+1EAF) — the correct longest match.

## AI disclosure

This contribution was written with AI assistance (analysis, implementation, and tests by an AI agent acting for the contributor). The fix was verified by running the repository's test suites and the differential checks described above.


ArthurZucker (commented):
thanks, this is something planned to fix, but we want to first work on normalizer performances. See #2209 that should give you an idea.

PR #2435, opened 2026-09-21
more cleanup

ArthurZucker:


tokenizers/CHANGELOG.md
ArthurZucker, line 1: had not written to it in such a while, does not make sense to keep

PR #2446, opened 2026-09-21
chore(release): bump dev version + trigger the release workflow on release/* branches

SBrandeis:


ArthurZucker (commented):
mmm fine as long as the actualy publish is filtered only on the tags still! that's what you are aiming for right?

SBrandeis:
> https://github.com/huggingface/tokenizers/pull/2450/changes/e6d9124db11c9cd5f23a84af019953735bb79134

Yes!
Only tags trigger the actual publish to registries

